PR #72, opened 2020-09-12
Create a deposit progress component

jcoyne:
This helps illustrate that motion is working to solve the kind of problems we want it to.

Screen Shot 2020-09-14 at 9 08 39 PM

### DIFF
--- a/app/javascript/modules/validate-forms.js
+++ b/app/javascript/modules/validate-forms.js
@@ -1,0 +1,21 @@
+// Example starter JavaScript for disabling form submissions if there are invalid fields
+(function () {
+  'use strict'
+  window.addEventListener('turbolinks:load', function() {
+    // Fetch all the forms we want to apply custom Bootstrap validation styles to
+    var forms = document.querySelectorAll('.needs-validation')
+
+    // Loop over them and prevent submission
+    Array.prototype.slice.call(forms)
+      .forEach(function (form) {
+        form.addEventListener('submit', function (event) {
+          if (!form.checkValidity()) {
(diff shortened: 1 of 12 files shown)


app/javascript/packs/application.js
```diff
@@ -19,4 +19,7 @@ import 'stylesheets/main'
 //
 // const images = require.context('../images', true)
 // const imagePath = (name) => images(name, true)
-import "motion"
\ No newline at end of file
+import "motion"
+
```
mjgiarlo, line 23: double blank line but nbd

app/javascript/modules/validate-forms.js
```diff
@@ -0,0 +1,21 @@
+// Example starter JavaScript for disabling form submissions if there are invalid fields
+(function () {
+  'use strict'
+  window.addEventListener('turbolinks:load', function() {
+    // Fetch all the forms we want to apply custom Bootstrap validation styles to
+    var forms = document.querySelectorAll('.needs-validation')
+
+    // Loop over them and prevent submission
```
mjgiarlo, line 8: can you say more about using JavaScript for client-side validation over Motion?

jcoyne: By default browsers do client side validation.  This removes the ugly browser default and adds the bootstrap way. We can also do validation with Motion.

app/javascript/packs/application.js
jcoyne: ESLint is that you?

mjgiarlo (approved):

PR #74, opened 2020-09-13
Allow work objects to be created

jcoyne:
Screen Shot 2020-09-12 at 9 08 40 PM


mjgiarlo (approved):
Looks great!

jcoyne:
@mjgiarlo this is just here so that we can see if Motion is going to be a workable solution. So far this leads me to believe there are complications ( https://github.com/unabridged/motion/issues/53 ). After we merge this PR, we can evaluate if stimulus-reflex might work or if we need to make this form without using the Rails helpers.

PR #79, opened 2020-09-22
Upload files to activestorage

jcoyne:
## Why was this change made?

This illustrates that the activestorage will be a workable filestore and that we are unable to work with dropzone and motion at the same time as motion overwrites the dynamically added file fields.

Screen Shot 2020-09-22 at 3 24 04 PM
Screen Shot 2020-09-22 at 3 24 21 PM
Screen Shot 2020-09-22 at 3 24 30 PM


mjgiarlo (approved):
Thank you for putting Motion through its paces. Using Dropzone in front of ActiveStorage for file uploads, Stimulus for front-end features, and esting with feature tests sounds good to me. And we can always layer in StimulusReflex later if we feel we absolutely *need* a reactive front-end to meet H2's needs.

PR #81, opened 2020-09-24
Add Sorbet for optional type checking

jcoyne:
See https://sorbet.org/docs/overview

Use 'srb tc' to run static type checks

It also does dynamic type checking:
Screen Shot 2020-09-23 at 9 58 11 PM


mjgiarlo (approved):
Neat. Add to inception deck and readme?

PR #104, opened 2020-10-06
Add continuous integration

### LINKED ISSUE #8: Add CI testing

mjgiarlo:
Fixes #8

## Why was this change made?

To better test the codebase

Includes:

* Update README about automatic sorbet rbi updates
* Add common rubocop buddies and make them happy before adding rubocop-sorbet
  * Add rake tasks to begin prepping for CI
* Add rubocop-sorbet style checks
* Update type-checking definitions after adding rubocop-sorbet
* Add CircleCI config
  * Including: rubocop, sorbet, and rspec

CI

README


.rubocop.yml
```diff
@@ -1,107 +1,23 @@
+require:
+  - rubocop-performance
+  - rubocop-rails
+  - rubocop-rspec
+  - rubocop-sorbet
+
 inherit_from: .rubocop_todo.yml
 
 AllCops:
+  NewCops: enable
```
jcoyne, line 10: If we do this, we probably want to pin tightly to the patch level rubocop.  I think we've avoided this in other projects so we can autoupdate rubocop without breaking the rubocop check.

jcoyne (changes requested):

.rubocop.yml
mjgiarlo: @jcoyne as in, instead of `~> 0.90.0`, `0.90.0`?

app/models/work.rb
```diff
@@ -2,8 +2,8 @@
 # frozen_string_literal: true
 
 class Work < ApplicationRecord
-  has_many :contributors
-  has_many :related_links
-  has_many :related_works
+  has_many :contributors, dependent: :nullify
```
mjgiarlo, line 5: @jcoyne are these the right dependent actions for these three relations? I took a big guess, so pls let me know.

jcoyne (changes requested):

app/models/work.rb
jcoyne: I'm thinking `:destroy` might be better.

Gemfile
```diff
@@ -5,49 +5,33 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
...
-gem 'dor-services-client'
+gem 'turbolinks', '~> 5'
+gem 'view_component', '~> 2.18'
+gem 'webpacker', '~> 5.0'
 
 group :development, :test do
   gem 'byebug'
   gem 'rspec-rails'
-  gem 'rubocop'
+  gem 'rubocop', '~> 0.90.0', require: false
```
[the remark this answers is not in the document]
mjgiarlo: @jcoyne wait, didn't you ask me to pin more tightly in your other comment?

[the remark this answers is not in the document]
jcoyne: It was we either need one or the other, but I prefer we are optimistic and keep a list of cops we opt-in to.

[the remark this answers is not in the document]
mjgiarlo: ok, going with the optimistic approach

app/models/work.rb
mjgiarlo: done

jcoyne (approved):

PR #123, opened 2020-10-09
Add the Collection model

jcoyne:
## Why was this change made?

So that we can establish the Work-Collection relationship.


db/schema.rb
```diff
@@ -1,4 +1,3 @@
-# typed: strict
```
mjgiarlo, line 1: Can you restore this? Or would you prefer to handle otherwise?

mjgiarlo (commented):
Question about db/schema.rb and sorbet.

db/schema.rb
jcoyne: Sure, this was just db:migrate + rubocop -a that made things this way.

mjgiarlo (approved):

PR #135, opened 2020-10-14
Add base testing and test coverage

mjgiarlo:
## Why was this change made?

To increase test coverage.

By... running the tests.

none.


Gemfile
```diff
@@ -22,6 +22,8 @@ gem 'webpacker', '~> 5.0'
 group :development, :test do
   gem 'byebug'
   gem 'factory_bot_rails'
+  gem 'rails-controller-testing'
```
jcoyne, line 25: Can we not use this?

spec/controllers/login_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
+# typed: false
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# NOTE: While controller specs are generally frowned upon, and we have largely
+#       leaned on request specs instead, testing session-related code was not
+#       possible in a request spec, and this is application biz logic worth
+#       testing.
+RSpec.describe LoginController do
```
jcoyne, line 10: Can you convert this to a request spec instead?

jcoyne (changes requested):

spec/controllers/login_controller_spec.rb
mjgiarlo: No. See comment above.

Gemfile
mjgiarlo: Removed.

spec/controllers/login_controller_spec.rb
jcoyne: All the maintainers say that we should not be writing new controller specs: https://github.com/rspec/rspec-rails/issues/1838#issuecomment-313410705  We can test session related code in integration test (e.g. feature specs)

mjgiarlo: @jcoyne since this a huge PR and I'd like to get it in before anything else, can I look into replacing it w/ a feature spec in a subsequent PR?

PR #136, opened 2020-10-14
Correct typo in name of work created field

mjgiarlo:
Draft because I'd like https://github.com/sul-dlss/happy-heron/pull/135 merged first.

## Why was this change made?

`created_etdf` should be `created_edtf` (EDTF == extended date-time format).

CI

none

### DIFF
--- a/db/migrate/20201013170323_simplify_columns_in_notifications.rb
+++ b/db/migrate/20201013170323_simplify_columns_in_notifications.rb
@@ -1,5 +1,3 @@
-# typed: true
-
 class SimplifyColumnsInNotifications < ActiveRecord::Migration[6.0]
   def change
     remove_column :notifications, :recipient_id, :integer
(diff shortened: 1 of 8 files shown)


db/migrate/20201013170323_simplify_columns_in_notifications.rb
```diff
@@ -1,5 +1,3 @@
-# typed: true
-
```
ndushay, line 2: whoa - what is this ... and why change an older migration?

ndushay (approved):
hahha.   etd,  edtf,  no possibility of confusion or muscle memory issues there ...

db/migrate/20201013170323_simplify_columns_in_notifications.rb
mjgiarlo: This is a magic comment that Sorbet (type-checker) uses to determine how tightly to type-check a file. But since this was added, we have instructed Sorbet to IGNORE the entire db directory because the files are largely generated, which means when they are regenerated, we'd need to re-add the type comments manually. I removed this comment for consistency with the other files in db, so as not to suggest that type-checking is being done on this file.

PR #139, opened 2020-10-16
Remove sub-type from the form

jcoyne:
## Why was this change made?
It has been removed from the design of this page and has been moved to the starting dialog


mjgiarlo (commented):
Tests failing and a question: did you run:

```
bundle exec rake rails_rbi:all
bundle exec srb rbi hidden-definitions
bundle exec srb rbi suggest-typed
```

:question:

app/models/user.rb
```diff
@@ -1,4 +1,4 @@
-# typed: strict
```
jcoyne, line 1: @mjgiarlo I think `bundle exec srb rbi suggest-typed` did this.  Do we want this?

mjgiarlo: @jcoyne It probably thinks it's helping us. To test that, you can try flipping that back to `typed: strict` and running `srb tc` to see if anything breaks?

mjgiarlo: Also, it seems weird that sorbet-rails removed so many definitions. but whatever?

mjgiarlo (approved):

PR #144, opened 2020-10-18
Create basic dashboard

### LINKED ISSUE #89: Create basic dashboard page.
When I click the Login link, and successfully log in, I am taken to the dashboard. The dashboard will have a list of collections (show the title) I created and a link to start a deposit in each collection. (Editing the collection from the dashboard is a different issue.)

See mock-up for design: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/426926001 For this issue, focus on the "Your collections" section of the design.

jcoyne:
## Why was this change made?

Fixes #89


Gemfile
```diff
@@ -54,3 +54,5 @@ group :deployment do
   gem 'capistrano-rvm', require: false
   gem 'dlss-capistrano', '~> 3.6', require: false
 end
+
+gem 'action_policy', '~> 0.5.3'
```
mjgiarlo, line 58: I think we've been stuffing our gems that are in every env up above the group-specific blocks. Open to moving this up above `bootsnap` or do you think there is good reason to put this down here?

README.md
```diff
@@ -19,6 +19,12 @@ Ruby dependencies can be installed with `bundle install`, JavaScript dependencie
 
 Start up dependencies with `docker-compose up`, then run tests with `bundle exec rspec`.
 
+To run the server in development mode set `REMOTE_USER` because we aren't running behing Shibboleth.
```
mjgiarlo, line 22: Ah, yes, good change!

app/models/collection.rb
```diff
@@ -4,4 +4,5 @@
 # Models a collection in the database
 class Collection < ApplicationRecord
   has_many :works, dependent: :destroy
+  belongs_to :creator, class_name: 'User'
```
mjgiarlo, line 7: Will Rails automatically validate this?

mjgiarlo (commented):
A few questions

Gemfile
jcoyne: This is what happens when you use the CLI: `bundle add action_policy`

app/models/collection.rb
jcoyne: yes. because we don't have `optional: true`

Gemfile
mjgiarlo: @jcoyne OK, cool. How would you feel about shifting the other gems from the top to the bottom, then, so future CLI appends build on what we already have? It's not a major concern but I also would like to head off any future rehashes of this if possible.

app/models/collection.rb
mjgiarlo: Perfect. Resolving.

mjgiarlo (approved):

PR #146, opened 2020-10-19
Deposit using SDR API

jcoyne:
## Why was this change made?

One of the goals for H2 is to develop an API for doing deposits.  If we make the GUI use that API it validates that the API is useful and complete and means we only have to support a single path for both GUI or API deposits.  This will allow us to record "who" is doing the deposit, because in the future we can use the users credentials (rather than the application credentials).


app/jobs/deposit_job.rb
```diff
@@ -3,16 +3,49 @@
...
+
   queue_as :default
 
+  sig { params(work: Work).void }
   def perform(work)
-    result = api_client.objects.register(params: create_model(work))
-    work.druid = result.externalIdentifier
+    job_id = deposit(create_model(work))
+    result = nil
+    until result
```
mjgiarlo, line 14: Not for this PR, but how would you feel about follow-up work that lets sdr-client manage its async interactions on apps' behalf, like dor-services-client does? https://github.com/sul-dlss/dor-services-client#asynchronous-results That will simplify the DepositJob.

config/settings.yml
```diff
@@ -1,6 +1,7 @@
-dor_services:
-  url: ~
-  token: ~
+sdr_api:
```
mjgiarlo, line 1: And confirming here: we already have this in shared_configs for all three deployed environments.

mjgiarlo (approved):
LGTM. Can you leave a little more detail in the PR description going forward? Link to an issue, some description of the rationale (for new folks joining the WC), answering the questions in the template, etc. Thanks!

app/jobs/deposit_job.rb
jcoyne: Sounds good to me.

config/settings.yml
jcoyne: exactly.

Issue #11, opened 2020-08-17
Add okcomputer for monitoring.


jcoyne:
Also: when okcomputer is added, please drop a comment to @tallenaz on https://github.com/sul-dlss/operations-tasks/issues/2333

PR #152, opened 2020-10-19
Adds okComputer.

justinlittman:
refs #11

## Why was this change made?
Add okComputer.

Locally.

NA


config/settings.yml
```diff
@@ -1,5 +1,5 @@
 sdr_api:
-  url: https://sdr-api-stage.stanford.edu
+  url: http://localhost:8080
```
justinlittman, line 2: Changed so that won't provide false positives in other environments.

mjgiarlo: Good call!

config/initializers/okcomputer.rb
```diff
@@ -0,0 +1,23 @@
...
+OkComputer.mount_at = 'status'
+OkComputer.check_in_parallel = true
+
+# Required
+OkComputer::Registry.register 'ruby_version', OkComputer::RubyVersionCheck.new
+
+OkComputer::Registry.register 'background_jobs', OkComputer::SidekiqLatencyCheck.new('default', 25)
+
+# Optional
+OkComputer::Registry.register 'background_jobs', OkComputer::SidekiqLatencyCheck.new('default', 25)
```
mjgiarlo, line 13: same as line 10, methinks

PR #160, opened 2020-10-19
Add collection request spec and reorganize request specs for consistency

mjgiarlo:
## Why was this change made?

To add some test cases.

By running CI.

None


spec/requests/authenticate_user_spec.rb
```diff
@@ -44,6 +44,12 @@
       get "/works/#{work.id}"
       expect(response).to have_http_status(:success)
     end
+
+    it 'redirects from /collections/new to login URL' do
+      get '/collections/new'
```
jcoyne, line 49: Should these go in the collections spec?

spec/requests/collections_creation_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    it 'creates a new collection' do
+      post '/collections', params: collection_params
+      expect(response).to have_http_status(:found)
+      expect(response).to redirect_to(dashboard_path)
+    end
+  end
+
+  # TODO: Uncomment when Collection validations are added
```
jcoyne, line 33: They are added, do a rebase?

spec/requests/authenticate_user_spec.rb
mjgiarlo: I could go either way, but the current pattern concentrates all authN stuff into this request spec.

spec/requests/collections_creation_spec.rb
mjgiarlo: Done.

mjgiarlo:
@jcoyne have another look?

spec/requests/authenticate_user_spec.rb
jcoyne: I think we should lay these out per request (rather than per-feature).

mjgiarlo: @jcoyne K. I'll re-organize in the morning.

mjgiarlo: @jcoyne Done. This is ready for another look now. The request specs now line up better w/ the controllers.

spec/requests/login_spec.rb
```diff
@@ -0,0 +1,48 @@
+# typed: false
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# NOTE: in deployed environments, this interaction is handled by Apache and
+#       Shibboleth in concert with the application
+RSpec.describe 'Login/out requests' do
```
jcoyne, line 8: Can we delete spec/controllers/ now?

mjgiarlo: No, I still need to try to convert that guy to a feature spec to test the "user return to" logic. (I'll do that next.)

jcoyne (approved):

PR #161, opened 2020-10-19
add model validations for presence

### LINKED ISSUE #148: Add model validations for required fields
We have a number of fields that are set as required in the DB schema but that do not have corresponding model validations, and it'd be good to have them so that `instance.valid?` doesn't lie to us.

If you would prefer to break this up into multiple issues, that's OK. If so, ping myself and @jcoyne and we can help.

Here are the fields that could benefit from presence validations in our model classes:

- [ ] Collection
  - [ ] name
  - [ ] description
  - [ ] contact_email
  - [ ] visibility
  - [ ] managers
- [ ] Contributor
  - [ ] first_name
  - [ ] last_name
- [ ] Notification
  - [ ] text
- [ ] RelatedLink
  - [ ] url
- [ ] RelatedWork
  - [ ] citation
- [ ] RoleTerm
  - [ ] label
- [x] User
  - [x] email (already done)
- [ ] Work
  - [ ] title
  - [ ] work_type
  - [ ] subtype
  - [ ] contact_email
  - [ ] created_edtf
  - [ ] abstract
  - [ ] citation
  - [ ] access
  - [ ] license
  - [ ] state

ndushay:
## Why was this change made?

Fixes #148.

Mike and I discussed whether or not to add tests to these validations.  They are straight ActiveRecord code, but that doesn't always mean our code does what we meant it to.   We decided to see how the coverage looked after this commit.

### DIFF
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -5,4 +5,10 @@
 class Collection < ApplicationRecord
   has_many :works, dependent: :destroy
   belongs_to :creator, class_name: 'User'
+
+  validates :contact_email, presence: true
+  validates :description, presence: true
+  validates :managers, presence: true
+  validates :name, presence: true
+  validates :visibility, presence: true
 end
--- a/app/models/contributor.rb
+++ b/app/models/contributor.rb
@@ -4,4 +4,7 @@
 class Contributor < ApplicationRecord
   belongs_to :work
   belongs_to :role_term
+
+  validates :first_name, presence: true
+  validates :last_name, presence: true
 end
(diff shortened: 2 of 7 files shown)


app/models/collection.rb
```diff
@@ -5,4 +5,10 @@
 class Collection < ApplicationRecord
   has_many :works, dependent: :destroy
   belongs_to :creator, class_name: 'User'
+
+  validates :contact_email, presence: true
```
mjgiarlo, line 9: Did/would you consider making these one-liners?

```ruby
validates :contact_email, :description, :managers, :name, :visibility, presence: true
```

:question: More readable? Less readable? I could go either way, tbh. What do you think?

mjgiarlo (approved):
Question about multi-line vs. single line. Approving because not a big deal.

PR #175, opened 2020-10-19
Additional db check for data loss

tallenaz:
## Why was this change made?

Ops wants to have a check for data loss

  I opted to switch from strict typing here b/c sorbet complained `any?` wasn't a method on a `Class`
  for the sig I tried to write: `sig { params(klass: Class) .returns(String)  }` and wasn't sure how to 
  wire it up properly.

Deployed to stage, verified a failing test with no collections in the db.

DevOpsDocs pending

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -1,13 +1,39 @@
-# typed: strict
+# typed: true
 # frozen_string_literal: true
 
 OkComputer.mount_at = 'status'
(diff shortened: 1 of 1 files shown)


jcoyne:
If you deploy this it will immediately start alerting for prod.  Are you sure?

tallenaz:
Good point. Let's hold off then. We're currently sending all prod alerts to nagios-stage only, but even so, I don't think we want this until we have some collections present.

PR #180, opened 2020-10-20
Add description when depositing

justinlittman:
## Why was this change made?
Add description to cocina model when depositing.

Unit

NA

### DIFF
--- a/app/services/description_generator.rb
+++ b/app/services/description_generator.rb
@@ -1,0 +1,77 @@
+# typed: true
+# frozen_string_literal: true
+
+# This generates a RequestDRO Description for a work
+class DescriptionGenerator
+  extend T::Sig
+
+  sig { params(work: Work).returns(Hash) }
+  def self.generate(work:)
+    new(work: work).generate
+  end
+
+  sig { params(work: Work).void }
+  def initialize(work:)
+    @work = work
+  end
+
+  sig { returns(Hash) }
+  def generate
+    {
+      title: title,
+      # TODO: keywords not yet in model.
+      note: [abstract, citation, contact],
+      event: [created_date]
+    }
+  end
+
+  private
+
+  attr_reader :work
+
+  def title
+    [
+      {
+        "value": work.title
+      }
+    ]
+  end
+
+  def abstract
+    {
+      "value": work.abstract,
+      "type": 'summary'
+    }
+  end
+
+  def citation
+    {
+      "value": work.citation,
+      "type": 'preferred citation'
+    }
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def created_date
+    {
+      "type": 'creation',
+      "date": [
+        {
+          "value": work.created_edtf,
+          "encoding": {
+            "code": 'edtf'
+          }
+        }
+      ]
+    }
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  def contact
+    {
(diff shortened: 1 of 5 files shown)


justinlittman:
@arcadiafalcone Do you mind taking a look at this as well? In particular, https://github.com/sul-dlss/happy-heron/pull/180/files#diff-12360b3f566d35f58f562b5272209905b309d638c7c19aa8c786b037edf38e87

app/services/description_generator.rb
```diff
@@ -0,0 +1,81 @@
...
+    {
+      "value": work.abstract,
+      "type": 'summary'
+    }
+  end
+
+  def citation
+    {
+      "value": 'This is the preferred citation.',
+      "type": work.citation
```
jcoyne, line 54: This should be `"preferred citation"`

app/services/description_generator.rb
```diff
@@ -0,0 +1,81 @@
...
+  def abstract
+    {
+      "value": work.abstract,
+      "type": 'summary'
+    }
+  end
+
+  def citation
+    {
+      "value": 'This is the preferred citation.',
```
jcoyne, line 53: This should be `work.citation`

app/services/description_generator.rb
```diff
@@ -0,0 +1,81 @@
...
+  end
+
+  sig { returns(Hash) }
+  def generate
+    {
+      title: title,
+      # TODO: keywords not yet in model.
+      note: [abstract, citation],
+      event: [created_date],
+      access: {
```
jcoyne, line 25: Are you sure about this?  In current hydrus objects, this is just in the notes as:
```
  <note type="contact" displayLabel="Contact">aydin@stanford.edu</note>
```

jcoyne (changes requested):

app/services/request_generator.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  sig { returns(Hash) }
+  # rubocop:disable Metrics/MethodLength
+  def generate
+    {
+      administrative: {
+        hasAdminPolicy: 'druid:pq757cd0790' # TODO: What should this be? this is the hydrus APO.
+      },
+      identification: {
+        sourceId: "hydrus:#{work.id}" # TODO: what should this be?
```
jcoyne, line 30: @andrewjbtw can you give feedback on this?

app/services/description_generator.rb
justinlittman: What does "sure" mean? This was my best guess. Happy to change.

jcoyne: I don't know where you are getting your idea to do access here. I thought maybe your information was better than mine.  My thought was to copy Hydrus unless Arcadia tells us otherwise.

app/services/description_generator.rb
justinlittman: Duh. Fixed.

jcoyne (approved):

app/services/request_generator.rb
andrewjbtw: For the sourceId we should generate it the way we do for Hydrus, though I guess we shouldn't use the word "Hydrus" anymore. But the auto-generated source we've been using has been working fine.

For APO it's more complicated and we should discuss. There are drawbacks to the current method but changing it could cause unexpected problems.

PR #181, opened 2020-10-20
Make hydrus APO a setting

justinlittman:
## Why was this change made?
To allow the hydrus APO to vary by environment.

Unit.

Updated stage and prod shared_configs.

Note: I created an APO in QA. I'm not sure if it matters, but I'm not positive that I got it completely right.

### DIFF
--- a/app/services/request_generator.rb
+++ b/app/services/request_generator.rb
@@ -24,7 +24,7 @@ class RequestGenerator
   def generate
     {
       administrative: {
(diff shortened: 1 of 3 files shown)


mjgiarlo (approved):
LGTM. ~~Do you think this should/does also include making shared_configs tweaks for the three H2 envs?~~ I now see you already did this. Thanks!

PR #182, opened 2020-10-20
Component file should not live among controllers

mjgiarlo:
## Why was this change made?

I'm guessing this was a git misfire. Confirm/deny, @jcoyne?

CI

none


jcoyne:
yes, an unintentional ctrl-v I assume.

jcoyne (approved):

PR #189, opened 2020-10-20
Replace controller spec with feature spec

mjgiarlo:
## Why was this change made?

Controller specs are deprecated and frowned upon because they test too much implementation behavior. Convert to a feature spec where we still have access to sessions via Warden, which we need to test user_return_to logic in the LoginController. As part of this work remove referrer param logic from LoginController which was copypasta from another app and has not been discussed in H2.

CI

None


jcoyne (changes requested):
Can the controller-testing gem be removed too?

mjgiarlo:
@jcoyne :speech_balloon: 
> Can the controller-testing gem be removed too?

AFAICT, it's already gone. I believe I removed it in a past PR.

jcoyne (approved):

PR #196, opened 2020-10-20
Restrict acceptable values for Work access field

### LINKED ISSUE #169: Validate Work#access is in the set
* world
* stanford

We ought to consider using a Rails enum backed by a postgres enum. We've done this in other projects, AFAIK.

mjgiarlo:
Fixes #169 

## Why was this change made?

Because we want access values to be predictable and selected from a system-provided and -enforced field.

Includes:
* Switch DB schema to SQL
  * This allows us to use nice features like Postgres enums, which reinforce Rails-level validations with database-backed ones, because Rails and Ruby let everyone do everything.
* Restrict acceptable values for Work access field
  * This is implemented using a Rails enum backed by a PG enum. I have tested the migration in both directions both with an empty database and with some rows in it, and the migration works as expected.

CI and with local migrations and rollbacks.

None.


app/components/embargo_component.html.erb
```diff
@@ -3,7 +3,7 @@
   <div class="mb-3 row">
     <%= form.label :access, 'Who can access?', class: 'col-sm-2 col-form-label' %>
     <div class="col-sm-10">
-      <%= form.select :access, ['Stanford'], {}, class: "form-select" %>
+      <%= form.select :access, ['stanford'], {}, class: "form-select" %>
```
ndushay, line 6: oooooooooo - nice catch.

app/models/work.rb
```diff
@@ -9,16 +9,13 @@ class Work < ApplicationRecord
...
-  validates :citation, presence: true
-  validates :contact_email, presence: true
-  validates :created_edtf, presence: true
-  validates :license, presence: true
-  validates :state, presence: true
-  validates :subtype, presence: true
-  validates :title, presence: true
-  validates :work_type, presence: true
+  validates :abstract, :access, :citation, :contact_email, :created_edtf,
+            :license, :state, :subtype, :title, :work_type, presence: true
```
ndushay, line 13: ya know, I wondered if I could do this ...  glad to know I could've.

spec/factories/works.rb
```diff
@@ -10,7 +10,6 @@
     created_edtf { '1900' }
     abstract { 'test abstract' }
     citation { 'test citation' }
-    access { 'stanford' }
```
ndushay, line 13: why does this go away?

ndushay (commented):
LGTM, but I'm not confident enough in db foo to approve.   Still wondering who decided what the default should be (amy?  hannah?)

spec/factories/works.rb
mjgiarlo: I removed this so I could test that the DB (vs. this factory) was the actor responsible for setting the default value.

app/models/work.rb
mjgiarlo: :man_dancing:

app/components/embargo_component.html.erb
mjgiarlo: Tests caught it! :tada:

mjgiarlo:
@ndushay :speech_balloon: 
> LGTM, but I'm not confident enough in db foo to approve. Still wondering who decided what the default should be (amy? hannah?)

I believe Amy gave the OK on Slack.

jcoyne (approved):

PR #197, opened 2020-10-20
Feature flag

### LINKED ISSUE #103: Add feature flag in configuration that enables accessioning or updating items
This will allow us to begin deploying to production early without worrying about content changes in production before the P.O. has said we're ready for that.

When feature flag is turned off, display a banner on the welcome page (and works page) that lets the user know that accessioning/updating items is not available at the moment.

ndushay:
## Why was this change made?

Fixes #103 

This feature flag will allow deployment to production early without risk to existing SDR content.  (Think:  before the PO has signed off)

Flag defaults to NOT ALLOWING content to be updated - this makes for more work in tests, but avoids possibly messing up in prod ...  I can reverse the default if desired.

If you try to create a collection or a work, you will be redirected to the dashboard with the alert message shown:

image

^^ NOTE that the close button for the alert text is on the LEFT.  This is how the access team seems to do it, which is good if the window is wide and/or if the alert text is short.  The last commit is the simple stylesheet change I made for this.

on my laptop and I also wrote some request specs for it.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,4 +27,11 @@ class ApplicationController < ActionController::Base
     # :user is the scope we are authenticating
     store_location_for(:user, request.fullpath)
   end
+
+  def ensure_sdr_updatable
+    return if Settings.allow_sdr_content_changes
+
+    flash[:warning] = 'Creating/Updating SDR content (i.e. collections or works) is not yet available.'
+    redirect_to :root
+  end
 end
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -9,7 +9,8 @@ require("@rails/activestorage").start()
 require("channels")
 import 'controllers'
 
-// import 'bootstrap'
+import 'popper.js'
+import 'bootstrap'
 require.context('../images', true)
 import 'stylesheets/main'
 
--- a/app/javascript/stylesheets/alert.scss
+++ b/app/javascript/stylesheets/alert.scss
@@ -1,0 +1,11 @@
+// Dismissible alerts
+
+// Expand the left padding and account for the close button's positioning.
+.alert-dismissible {
+  padding-left: $alert-dismissible-padding-r;
+
+  // Move alert button close link position to left
+  .btn-close {
+    left: 0;
+  }
+}
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -38,6 +38,8 @@
       </nav>
     </header>
 
+    <%= render 'shared/flashes' %>
+
     <%= yield %>
 
     <footer class="stanford-footer p-3 p-md-5 mt-5 bg-light text-center text-sm-left">
--- a/app/views/shared/_flashes.html.erb
+++ b/app/views/shared/_flashes.html.erb
@@ -1,0 +1,34 @@
+<% if [flash[:success], flash[:notice], flash[:multi_line_notice], flash[:warning], flash[:error]].any? %>
+  <div class="flashes">
+    <% if flash[:success] %>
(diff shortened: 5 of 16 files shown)


app/javascript/packs/application.js
```diff
@@ -9,7 +9,8 @@ require("@rails/activestorage").start()
 require("channels")
 import 'controllers'
 
-// import 'bootstrap'
+import 'popper.js'
+import 'bootstrap'
```
ndushay, line 13: the flash messages are handled readily with bootstrap.  Bootstrap 5 depends on popper.

app/views/layouts/application.html.erb
```diff
@@ -38,6 +38,8 @@
       </nav>
     </header>
 
+    <%= render 'shared/flashes' %>
+
```
ndushay, line 42: this is the pattern I saw in a bunch of access team apps.  it seemed like a great way to go.

package.json
```diff
@@ -8,6 +8,7 @@
     "@rails/webpacker": "5.2.1",
     "bootstrap": "^5.0.0-alpha1",
     "dropzone": "^5.7.2",
+    "popper.js": "^1.16.1",
```
ndushay, line 11: while the bootstrap package was already installed, we needed the popper.js package for bootstrap 5.

app/javascript/stylesheets/alert.scss
```diff
@@ -0,0 +1,11 @@
...
+
+// Expand the left padding and account for the close button's positioning.
+.alert-dismissible {
+  padding-left: $alert-dismissible-padding-r;
+
+  // Move alert button close link position to left
+  .btn-close {
+    left: 0;
+  }
+}
```
ndushay, line 11: by default, bootstrap puts the alert close button on the RIGHT.  This can be far away from the alert text, if the window is wide and/or if the alert text is short.   It looked like the access button class used puts the button on the LEFT, so this is the simple change to use the correct bootstrap button and put it on the LEFT.

app/views/shared/_flashes.html.erb
```diff
@@ -0,0 +1,34 @@
+<% if [flash[:success], flash[:notice], flash[:multi_line_notice], flash[:warning], flash[:error]].any? %>
+  <div class="flashes">
+    <% if flash[:success] %>
+      <div class="alert alert-success alert-dismissible" role="alert">
+        <button type="button" class="btn-close" data-dismiss="alert" aria-label="Close"></button>
```
ndushay, line 5: access team used class "close" for the button.  It seemed less impactful to use the expected btn-close for alert-dismissable buttons and modify the scss slightly.

app/controllers/application_controller.rb
```diff
@@ -27,4 +27,11 @@ def store_user_location!
     # :user is the scope we are authenticating
     store_location_for(:user, request.fullpath)
   end
+
+  def allow_sdr_updates?
```
jcoyne, line 31: Can this be `ensure_update_available`?  The `?` method name idiom commonly indicates the return value is a boolean, but this is not.

jcoyne (changes requested):

app/javascript/stylesheets/alert.scss
mjgiarlo: Sounds good to me.

app/controllers/application_controller.rb
ndushay: good point.  changing to `ensure_sdr_updatable`

app/views/shared/_flashes.html.erb
mjgiarlo: WFM

ndushay:
@jcoyne: renamed method - thanks for spotting that.

jcoyne (approved):

PR #214, opened 2020-10-22
Replace role_terms table with code that has the lists

### LINKED ISSUE #208: Seed RoleTerm data
```
Individual
  Advisor
  Author
  Collector
  Contributing author
  Creator
  Editor
  Primary advisor
  Principal investigator
Organization
  Author
  Contributing author
  Degree granting institution
  Distributor
  Publisher
  Sponsor
Event
  Conference

```

jcoyne:
## Why was this change made?
Fixes #208


mjgiarlo (commented):
All looks good. Question: should we stuff the lists of roles and contributor types in config? Or do you think it makes more sense to have them defined as constants directly in the contributor model?

jcoyne:
@mjgiarlo I think putting them in config only adds complexity and indirection to the code.  What benefits are you thinking that would bring us?

mjgiarlo:
> @mjgiarlo I think putting them in config only adds complexity and indirection to the code. What benefits are you thinking that would bring us?

I didn't have any benefits in mind. Just wanted to get your take on it. Whether it's in the H2 codebase or shared_configs, a change is a mere `cap env deploy` away, so no big diff really. (And we likely don't see needing to vary these lists between envs, so :man_shrugging: )

mjgiarlo (approved):

PR #224, opened 2020-10-23
Add publication and created range dates to the work form 

### LINKED ISSUE #49: Add "Date content was created" field to item edit page
Screen Shot 2020-09-24 at 10 59 20 AM
This should persist into the created_etdf as ETDF (https://www.loc.gov/standards/datetime/)

https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403

jcoyne:
## Why was this change made?

Fixes #49 

Its on staging:

Screen Shot 2020-10-26 at 10 20 38 AM


ndushay:
This seems like a good PR for before and after pix ...

Gemfile
```diff
@@ -51,6 +51,7 @@ gem 'okcomputer'
 gem 'pg'
 gem 'rails', '~> 6.0.3', '>= 6.0.3.2'
 gem 'redis', '~> 4.0'
+gem 'reform-rails', '~> 0.2.0'
```
ndushay, line 54: would love 5 words or less explaining what this gem is giving us

app/controllers/works_controller.rb
```diff
@@ -32,6 +34,11 @@ def show
 
   def work_params
     params.require(:work).permit(:title, :work_type, :subtype, :contact_email,
+                                 'published(1i)', 'published(2i)', 'published(3i)',
```
mjgiarlo, line 37: Can you say more about this `'attr({number}i)'` pattern used in this PR? It's new to me. Is it a reform incantation? Something something virtual attributes? I wonder if this is new/different enough to the team that it's worth adding a comment?

Gemfile
jcoyne: Its a way of having form-models.  We use the reform pattern in argo (but we didn't use reform for obscure reasons): https://github.com/sul-dlss/argo/blob/d317403b9bfd8abbe37ab97947abf2dba03e67f0/app/forms/base_form.rb#L3

app/forms/work_form.rb
```diff
@@ -0,0 +1,95 @@
...
+               item = contributors.find_by(id: fragment['id']) if fragment['id'].present?
+
+               if fragment['_destroy'] == '1'
+                 contributors.delete(item)
+                 return skip!
+               elsif fragment['first_name'].blank?
+                 return skip!
+               end
+               item || contributors.append(Contributor.new)
+             } do
```
mjgiarlo, line 60: What is the fragment here? Even though I've used reform a bit in the past, this is a bit inscrutable to me. Add a code comment?

app/forms/work_form.rb
```diff
@@ -0,0 +1,95 @@
...
+    property :first_name
+    property :last_name
+    property :role_term
+    property :_destroy, virtual: true
+  end
+
+  private
+
+  def deserialize_edtf(name, offset = 0)
+    year = send("#{name}(#{offset + 1}i)")
```
mjgiarlo, line 71: Are virtual properties defined as private methods?

app/forms/work_form.rb
```diff
@@ -0,0 +1,95 @@
...
+    day = send("#{name}(#{offset + 3}i)")
+    deserialize_edtf_date(year, month, day)
+  end
+
+  def deserialize_edtf_range(name)
+    start = deserialize_edtf(name)
+    finish = deserialize_edtf(name, 3)
+    return unless start && finish
+
+    [start, finish].join('/')
```
mjgiarlo, line 82: Worth adding a comment here that EDTF time intervals use `/` as a delimiter? Or perhaps name it as a constant?

app/forms/work_form.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def deserialize_edtf_date(year, month, day)
```
mjgiarlo, line 85: I wonder if we might want the EDTF-specific code in a new service object, but perhaps it's best to wait and see if someone other than WorkForm will need this code?

app/models/work.rb
```diff
@@ -11,11 +11,9 @@ class Work < ApplicationRecord
   has_many :related_works, dependent: :destroy
   has_many_attached :files
 
-  accepts_nested_attributes_for :contributors, allow_destroy: true, reject_if: :last_name_blank
-
   validates :abstract, :access, :citation, :state, :subtype, :title, presence: true
   validates :contact_email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
-  validates :created_edtf, presence: true, edtf: true
+  validates :created_edtf, :published_edtf, edtf: true
```
mjgiarlo, line 16: Does this mean we're no longer validating that `created_edtf` is required? Or did that validation move up into the form?

app/forms/work_form.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  collection :contributors,
+             populator: lambda { |fragment:, **|
+               # find out if incoming Contributor is already added.
+               item = contributors.find_by(id: fragment['id']) if fragment['id'].present?
+
+               if fragment['_destroy'] == '1'
+                 contributors.delete(item)
+                 return skip!
+               elsif fragment['first_name'].blank?
```
mjgiarlo, line 56: We used to reject if last name was blank and now we reject if first name is blank? Why the change?

app/services/description_generator.rb
```diff
@@ -21,7 +21,7 @@ def generate
       title: title,
       # TODO: keywords not yet in model.
       note: [abstract, citation, contact],
-      event: [created_date]
+      event: [created_date].compact
```
mjgiarlo, line 24: I guess this is more or less the same as `Array(created_date)`? I think it's a six-of-one-half-dozen-of-the-other situation, though.

app/validators/edtf_validator.rb
```diff
@@ -3,7 +3,7 @@
 
 class EdtfValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
-    return unless EDTF.parse(value).nil?
+    return if value.nil? || EDTF.parse(value)
```
mjgiarlo, line 6: This change allows `nil` `created_edtf` values? Can you say more about that?

mjgiarlo (commented):
I have a handful of questions.

app/controllers/works_controller.rb
jcoyne: That's how rails sends dates that are in parts: https://apidock.com/rails/ActionView/Helpers/DateHelper/date_select#422-Date-select-with-assert-valid-keys

app/forms/work_form.rb
jcoyne: Fragement is the row data from the html form.

app/forms/work_form.rb
jcoyne: I believe they are public

app/controllers/works_controller.rb
mjgiarlo: @jcoyne :beers: thank you. Worth adding a comment about that? It was new to me, anyway.

app/models/work.rb
jcoyne: The designs changed and created date is no longer required.

app/forms/work_form.rb
jcoyne: I don't really know what the requirements should be.  It was arbitrary the first time and it's arbitrary now.

app/forms/work_form.rb
mjgiarlo: @jcoyne sorry to be a broken record, but... can you add a comment about that? I don't think our collective reform knowledge is up to snuff.

app/services/description_generator.rb
jcoyne: no, this eliminates `[nil]`, which `Array(created_date)` would not.

app/forms/work_form.rb
mjgiarlo: @jcoyne open to switching to `public_send`? when I see `send`, I get the message that we're calling private methods.

app/validators/edtf_validator.rb
jcoyne: No, this means "don't validate if they don't have a date".  If we want it to be not nil, that should be done by the presence validator.

app/models/work.rb
mjgiarlo: @jcoyne Thank you!

app/forms/work_form.rb
mjgiarlo: @jcoyne should we at least retain the arbitrariness for now? I know names are fraught with peril, so either way we are likely to get it wrong. :laughing: I'll add a planning discussion topic about this.

app/validators/edtf_validator.rb
mjgiarlo: No longer needs commenting.

app/forms/work_form.rb
mjgiarlo: https://github.com/sul-dlss/happy-heron/issues/232

app/forms/work_form.rb
jcoyne: sure.

app/forms/work_form.rb
jcoyne: It's added.

app/controllers/works_controller.rb
```diff
@@ -6,19 +6,21 @@ class WorksController < ApplicationController
...
   def create
-    @collection = Collection.find(params[:collection_id])
-    @work = Work.new(work_params.merge(collection: @collection))
-    if @work.save
-      DepositJob.perform_later(@work) if params[:commit] == 'Deposit'
-      redirect_to @work
+    work = Work.new
+    @form = WorkForm.new(work)
+
+    stuff = work_params.merge(collection_id: params[:collection_id])
```
mjgiarlo, line 20: :laughing: I'm not sure whether I care about this variable name or not, but I'm getting a kick out of it.

app/forms/work_form.rb
```diff
@@ -0,0 +1,96 @@
...
+    model.published_edtf = deserialize_edtf(:published)
+
+    super
+  end
+
+  validates :title, presence: true
+
+  collection :contributors,
+             populator: lambda { |fragment:, **|
+               # The fragment represents one row of the contributor data from the HTML form
```
mjgiarlo, line 50: Thank you.

mjgiarlo (approved):

PR #226, opened 2020-10-23
workgroup based authorization groundwork, fleshed out permissions checking, rudimentary authZ failure handling

### LINKED ISSUE #102: Prevent unauthorized users from depositing
Admit the user if:

* in `dlss:hydrus-app-adminstrators` LDAP workgroup
* in `dlss:hydrus-app-collection-creators` LDAP workgroup
* assigned reviewer role to one or more collections in the application (not yet implemented)
* assigned depositor role to one or more collections in the application (not yet implemented)

If the user is admitted, they see the dashboard. (What the dashboard renders depends on what roles are assigned to the user in the database.) 

If the user is not admitted, they get the [contact form modal](https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428680772) loaded over the front page.

To do this:
* Grab groups from Shibboleth/WebAuth
* Denote authorized groups in config (shared_configs)
* Use ActionPolicy to authorize individual actions

See current Argo repo or hydra_etd repo for examples.

Rationale: We plan to use a combination of LDAP groups and application database records to determine what a user can access. We think we only need the one LDAP group to gate what happens when a user signs in.

jmartin-sul:
## Why was this change made?

groundwork for Shibboleth based authorization, with the start of an implementation for #102 (minus some of the suggested UI).

closes #102 (ticketed the part that didn't make it into this PR)

unit tests

base settings now has app user workgroup names.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -31,7 +31,8 @@ class ApplicationController < ActionController::Base
   end
 
   def deny_access
-    render status: :unauthorized, html: 'You are unauthorized to see this'
+    flash[:warning] = 'You are not authorized to perform the requested action'
+    redirect_to :root
   end
 
   # Its important that the location is NOT stored if:
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: false
 # frozen_string_literal: true
 
 class ApplicationController < ActionController::Base
@@ -10,7 +10,29 @@ class ApplicationController < ActionController::Base
   # From: https://github.com/heartcombo/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update
   before_action :store_user_location!, if: :storable_location?
 
+  rescue_from ActionPolicy::Unauthorized, with: :deny_access
+
+  sig { returns(T.nilable(User)) }
+  def current_user
+    super.tap do |cur_user|
+      break unless cur_user
+
+      cur_user.groups = ldap_groups
+    end
+  end
+
   private
+
+  sig { returns(T::Array[String]) }
+  def ldap_groups
+    return request.env['eduPersonEntitlement'].split(';') if request.env['eduPersonEntitlement']
+
+    []
+  end
+
+  def deny_access
+    render status: :unauthorized, html: 'You are unauthorized to see this'
+  end
 
   # Its important that the location is NOT stored if:
   # - The request method is not GET (non idempotent)
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -1,11 +1,13 @@
-# typed: true
+# typed: false
 # frozen_string_literal: true
 
 # Displays the list of collections to the user
 class DashboardsController < ApplicationController
   before_action :authenticate_user!
+  verify_authorized
 
   def show
     @collections = authorized_scope(Collection.all)
+    authorize! :dashboard
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,10 +10,35 @@ class User < ApplicationRecord
 
   has_many :notifications, dependent: :destroy
 
+  sig { params(groups: T.nilable(T::Array[String])).returns(T.nilable(T::Array[String])) }
+  attr_writer :groups
+
   devise :remote_user_authenticatable
+
+  sig { returns(T::Boolean) }
+  def application_user?
+    (groups & application_user_groups).present?
+  end
+
+  sig { returns(T::Boolean) }
+  def collection_creator?
+    groups.include?(Settings.authorization_workgroup_names.collection_creators)
+  end
+
+  sig { returns(T::Array[String]) }
+  def groups
+    @groups || []
+  end
 
   sig { returns(String) }
   def to_s
     email
+  end
+
+  private
+
+  sig { returns(T::Array[String]) }
+  def application_user_groups
+    [Settings.authorization_workgroup_names.administrators, Settings.authorization_workgroup_names.collection_creators]
   end
 end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -19,3 +19,8 @@ h2:
 
 # feature flag
 allow_sdr_content_changes: true
+
+# mappings from authorization role to workgroup name
+authorization_workgroup_names:
+  administrators: 'dlss:hydrus-app-adminstrators'
+  collection_creators: 'dlss:hydrus-app-collection-creators'
--- a/lib/test_shibboleth_headers.rb
+++ b/lib/test_shibboleth_headers.rb
@@ -13,7 +13,7 @@ class TestShibbolethHeaders
 
   def call(env)
     env['REMOTE_USER'] = user
-    # env['eduPersonEntitlement'] = Array(groups).join(';')
+    env['eduPersonEntitlement'] = groups.join(';') if groups.present? # allow omission of header for testing
     @app.call(env)
   end
 end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -4,8 +4,10 @@
 class WorksController < ApplicationController
   before_action :authenticate_user!, except: [:show]
   before_action :ensure_sdr_updatable
+  verify_authorized except: [:show]
 
   def new
+    authorize!
     collection = Collection.find(params[:collection_id])
     raise 'Missing required parameter work_type' unless params[:work_type]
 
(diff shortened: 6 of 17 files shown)


app/controllers/application_controller.rb
```diff
@@ -1,20 +1,46 @@
...
+    end
+  end
 
   private
 
+  sig { returns(T::Array[String]) }
+  def ldap_groups
+    return request.env['eduPersonEntitlement'].split(';') if request.env['eduPersonEntitlement']
+
+    []
```
jmartin-sul, line 31: d'oh, just realized this line's code path isn't tested, happy to revisit if people want.

app/controllers/application_controller.rb
```diff
@@ -1,20 +1,46 @@
-# typed: true
+# typed: false
```
jmartin-sul, line 1: this was because sorbet didn't recognize `rescue_from`.  is there a way to tell it "no really this method exists for this class"?

app/controllers/application_controller.rb
```diff
@@ -1,20 +1,46 @@
...
+  rescue_from ActionPolicy::Unauthorized, with: :deny_access
+
+  sig { returns(T.nilable(User)) }
+  def current_user
+    super.tap do |cur_user|
+      break unless cur_user
+
+      cur_user.groups = ldap_groups
+    end
+  end
```
jmartin-sul, line 23: implementation based on patterns used in hydrus and argo, per ticket suggestion

jmartin-sul: also, per jcoyne's suggestion, i played for a couple hours with having a `UserWithGroups` class, that has an instance of the `User` active record class, but i couldn't find a way to get devise to play nicely with a  setup where the object returned by `current_user` isn't of the class that's used for database persistence of the user record.  and my takeaway from that conversation with @jcoyne was that that was the point of trying that, since having `groups` and `groups=` methods on an AR object could be confusing, since we don't persist that info (we always want it from the request env headers).

app/models/user.rb
```diff
@@ -1,6 +1,10 @@
 # typed: strict
 # frozen_string_literal: true
 
+# TODO: consider adding a UserWithGroups class: would take a User instance and a groups (String)
+# list, and would put the groups list in a field on the UserWithGroups obj.  maybe not worth it but
+# play with it and see, could be useful in illustrating via code itself that #groups isn't part of
+# the persisted User obj.
```
jmartin-sul, line 7: d'oh, stray TODO comment, will remove after i get back from lunch...

app/controllers/dashboards_controller.rb
```diff
@@ -1,11 +1,13 @@
-# typed: true
+# typed: false
```
jmartin-sul, line 1: `verify_authorized` made me do it.  would be nice (as with the exception handling in `ApplicationController` if we could tell sorbet "no really, `verify_authorized` will work".  i think that's the only thing preventing this file from being `typed: strict`.

jmartin-sul (commented):
left some inline explanations

app/controllers/application_controller.rb
jmartin-sul: also, fwiw, i suspect the lack of diff coverage on this line is what's failing the build

config/settings.yml
```diff
@@ -19,3 +19,7 @@ h2:
 
 # feature flag
 allow_sdr_content_changes: true
+
+# mappings from authorization role to workgroup name
+authorization_workgroup_names:
+  application_user: 'dlss:h2-app-users'
```
mjgiarlo, line 25: After mid-sprint planning discussion today, this approach is going to change. See the issue for some updated text. Happy to chat and/or pair tomorrow. cc: @jcoyne

jmartin-sul:
currently amending per comment from (and zoom discussion with) @mjgiarlo.

app/controllers/application_controller.rb
jcoyne: Did you try `typed: true`? I've found that works more often than `strict`

app/controllers/application_controller.rb
jmartin-sul: fixed

app/controllers/application_controller.rb
jmartin-sul: yeah, but no luck with that either.

jmartin-sul:
@jcoyne and @mjgiarlo, i think this is ready for an actual review.

per discussion with @mjgiarlo yesterday
* i've tried to update this PR to use the updated group names from #102's current incarnation.
* i've deferred the more robust authZ failure UI for another ticket, which i'll file in a moment (the "If the user is not admitted, they get the contact form modal loaded over the front page" part).  it seems like this work could go in either way, and it'd be nice to get it merged and build on it, even if i just pick up the UI refinement ticket next.
* i tried to flesh out the policy methods that previously always evaluated to `true`, by only allowing collection creators to create works and collections (but not app admins who are not in the collection creation workgroup).

app/controllers/dashboards_controller.rb
```diff
@@ -1,11 +1,13 @@
...
 # frozen_string_literal: true
 
 # Displays the list of collections to the user
 class DashboardsController < ApplicationController
   before_action :authenticate_user!
+  verify_authorized
 
   def show
     @collections = authorized_scope(Collection.all)
+    authorize! :dashboard
```
mjgiarlo, line 11: I don't think it's a big deal, but I wonder if we need line 11 below line 10 here. (Not familiar with `authorized_scope`.)

lib/test_shibboleth_headers.rb
```diff
@@ -13,7 +13,7 @@ def initialize(app)
 
   def call(env)
     env['REMOTE_USER'] = user
-    # env['eduPersonEntitlement'] = Array(groups).join(';')
+    env['eduPersonEntitlement'] = groups.join(';') if groups.present? # allow omission of headeer for testing
```
mjgiarlo, line 16: Typo: `s/headeer/header/`

app/controllers/dashboards_controller.rb
mjgiarlo, line 11: I don't think it's a big deal, but I wonder if we need line 11 below line 10 here. (Not familiar with authorized_scope.)

mjgiarlo (commented):
Looks good! Have a question and a typo comment. Also, I was expecting to see some authorization stuff in the collections controller but apparently it's not needed? Even though it's covered in the specs? :thinking:

app/controllers/dashboards_controller.rb
jmartin-sul: i'm also unfamiliar with `authorized_scope`, and hadn't really considered the possibility it'd cover the authorization (didn't run the tests before adding the `authorize!` on line 11.

but i just tried out commenting out line 11, and apparently `authorized_scope` doesn't fulfill the `verify_authorized` requirement, because leaving off `authorized! :dashboard` gets me 6 failures along the lines of:

```
  1) Create a new collection and deposit to it when successful deposit deposits and renders work show page
     Failure/Error: @app.call(env)
     
     ActionPolicy::UnauthorizedAction:
       Action 'dashboards#show' hasn't been authorized
     # ./lib/test_shibboleth_headers.rb:17:in `call'
     # ------------------
     # --- Caused by: ---
     # Capybara::ElementNotFound:
     #   Unable to find link "+ Create a new collection"
     #   ./spec/features/create_new_work_spec.rb:20:in `block (3 levels) in <top (required)>'
```

(the other 5 were were in `dashboards_spec.rb`)

lib/test_shibboleth_headers.rb
jmartin-sul: good catch, thanks! (my old laptop keyboard is flaking, and giving me enough double-presses that i'm pretty sure it's not just my own clumsiness)

jmartin-sul (commented):
@mjgiarlo 💬 
> Also, I was expecting to see some authorization stuff in the collections controller but apparently it's not needed? Even though it's covered in the specs?

'twas already there (`authorize! collection` in `CollectionsController#new` and `#create`), just had to flesh out the policy that was already being invoked.

responded to the question about the possibly superfluous authZ check (seemed necessary), and will fix the typo now...

jmartin-sul:
@mjgiarlo, ready for another look, i think, fixed the typo and responded to the other questions

lib/test_shibboleth_headers.rb
mjgiarlo: ha!

app/controllers/dashboards_controller.rb
mjgiarlo: Ah, I should have been more specific: I was wondering if these two lines could switch places.

mjgiarlo (approved):

app/controllers/dashboards_controller.rb
jmartin-sul: ah, gotcha, fixed in #277

PR #230, opened 2020-10-25
Add auto-citation

### LINKED ISSUE #46: Preferred citation should default to a generated citation
See mockup in wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403 and 
Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.3nd0v0kp7034

At mid-sprint check-in we discussed that this would autopopulate as information is filled in. This would include a placeholder link for the PURL (something like "_Link to PURL will be included here_.") Once the item is deposited and the PURL is available, the item view page will update so that the citation includes the citation as it will look on the PURL page. This should ideally also be updated in the citation box in the deposit form so that the user doesn't get confused about that text. 

If a depositor switches to the toggle to be able to edit the citation, whatever is in the box should stay there, but the content will no longer autoupdate. If they leave the placeholder text, then the link should be inserted, but if they delete it, then it won't be. (I'm sure we discussed that last bit, but I think it was implied and makes the most sense to me.)

Note that this means the PURL info in the citation will not be preserved.

jcoyne:
## Why was this change made?

Fixes #46 

Screen Shot 2020-12-01 at 11 11 16 AM

Screen Shot 2020-12-01 at 11 16 21 AM

### DIFF
--- a/app/components/works/contributor_row_component.html.erb
+++ b/app/components/works/contributor_row_component.html.erb
@@ -17,12 +17,20 @@
   </div>
   <div class="col-md-4" data-target="contributors.personName">
     <%= form.label :first_name, class: 'form-label' %>
-    <%= form.text_field :first_name, class: 'form-control', data: { action: 'change->contributors#inputChanged', target: 'contributors.personNameInput' }, required: true %>
+    <%= form.text_field :first_name, class: 'form-control',
+        data: {
+          action: 'change->contributors#inputChanged change->auto-citation#updateDisplay',
+          target: 'contributors.personNameInput auto-citation.contributorFirst'
+        }, required: true %>
     <div class="invalid-feedback">You must provide a first name</div>
   </div>
   <div class="col-md-4" data-target="contributors.personName">
     <%= form.label :last_name, class: 'form-label' %>
-    <%= form.text_field :last_name, class: 'form-control', data: { action: 'change->contributors#inputChanged', target: 'contributors.personNameInput' }, required: true %>
+    <%= form.text_field :last_name, class: 'form-control',
+        data: {
+          action: 'change->contributors#inputChanged change->auto-citation#updateDisplay',
+          target: 'contributors.personNameInput auto-citation.contributorLast'
+        }, required: true %>
     <div class="invalid-feedback">You must provide a last name</div>
   </div>
   <div class="col-md-8" data-target="contributors.organizationName">
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -97,15 +97,15 @@ class WorksController < ObjectsController
                      'created(1i)', 'created(2i)', 'created(3i)',
                      'created_range(1i)', 'created_range(2i)', 'created_range(3i)',
                      'created_range(4i)', 'created_range(5i)', 'created_range(6i)',
-                     :created_edtf, :abstract, :citation, :access, :license,
+                     :abstract, :citation_auto, :citation, :default_citation,
+                     :access, :license,
                      :release, 'embargo_date(1i)', 'embargo_date(2i)', 'embargo_date(3i)',
-                     :agree_to_terms,
-                     subtype: [],
-                     attached_files_attributes: %i[_destroy id label hide file],
-                     contributors_attributes: %i[_destroy id full_name first_name last_name role_term],
-                     keywords_attributes: %i[_destroy id label uri],
-                     related_works_attributes: %i[_destroy id citation],
-                     related_links_attributes: %i[_destroy id link_title url])
+                     :agree_to_terms, subtype: [],
+                                      attached_files_attributes: %i[_destroy id label hide file],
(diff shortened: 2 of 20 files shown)


app/components/works/contributor_row_component.html.erb
```diff
@@ -17,12 +17,20 @@
   </div>
   <div class="col-md-4" data-target="contributors.personName">
     <%= form.label :first_name, class: 'form-label' %>
-    <%= form.text_field :first_name, class: 'form-control', data: { action: 'change->contributors#inputChanged', target: 'contributors.personNameInput' }, required: true %>
+    <%= form.text_field :first_name, class: 'form-control',
+        data: {
+          action: 'change->contributors#inputChanged change->auto-citation#contributorFirstChanged',
```
mjgiarlo, line 22: I am really, really digging Stimulus.

app/controllers/works_controller.rb
```diff
@@ -97,15 +97,15 @@ def work_params
                      'created(1i)', 'created(2i)', 'created(3i)',
                      'created_range(1i)', 'created_range(2i)', 'created_range(3i)',
                      'created_range(4i)', 'created_range(5i)', 'created_range(6i)',
-                     :created_edtf, :abstract, :citation, :access, :license,
+                     :abstract, :citation_auto, :citation, :default_citation,
```
mjgiarlo, line 100: Why did `created_edtf` drop out of the list? Handled elsewhere now?

app/javascript/controllers/auto_citation_controller.js
```diff
@@ -0,0 +1,118 @@
...
+    this.title = this.titleFieldTarget.value
+    this.year = this.yearTarget.value
+    this.month = this.monthTarget.value
+    this.day = this.monthTarget.value
+    this.purl = this.data.get("purl") || ":link:" // Use a real purl on a persisted item or a placeholder
+
+    this.updateDisplay()
+
+    this.displayDefault(this.switchTarget.checked)
+    // this.checkField(this.contributorRoleTarget)
```
mjgiarlo, line 17: Cruft? Or a TODO?

mjgiarlo (commented):
Can you include an animated gif or some screenshots to show this in action? Thanks!

app/controllers/works_controller.rb
jcoyne: It isn't currently used. Seemed like old cruft.

app/javascript/controllers/auto_citation_controller.js
```diff
@@ -0,0 +1,96 @@
+import { Controller } from "stimulus";
+
+export default class extends Controller {
+  static targets = ["titleField", "contributorFirst", "contributorLast", "contributorRole",
+    "year", "month", "day", "manual", "auto", "switch"];
+
+  connect() {
+    this.purl = this.data.get("purl") || ":link:" // Use a real purl on a persisted item or a placeholder
```
mjgiarlo, line 8: Do/should we include any hints in the UI so the user knows to leave `:link:` alone for later? Or not necessary?

jcoyne: That's specified in the tool tip in the h2 design documentation, but we haven't gotten to tooltips yet.

app/controllers/works_controller.rb
mjgiarlo: OK, great.

app/javascript/controllers/auto_citation_controller.js
mjgiarlo: @jcoyne OK, works for me.

app/javascript/controllers/auto_citation_controller.js
jcoyne: Removed.

app/javascript/controllers/auto_citation_controller.js
```diff
@@ -0,0 +1,95 @@
...
+      case 1:
+      case 2:
+        return this.authors.join(' and ')
+      default:
+        return `${this.authors.slice(0,-1).join(', ')}, and ${this.authors.slice(-1)}`
+    }
+  }
+
+  get authors() {
+    return this.contributorRoleTargets.map((roleField, index) => {
```
mjgiarlo, line 34: Does this code currently assume that all contributors are authors?

jcoyne: Yes, and I think that's safe to punt on until we get the new designs for authors implemented: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420954309

mjgiarlo: That's fine by me as long as, say, it won't barf if the user has one or more org contributors. If that's the case, approving.

mjgiarlo (approved):

PR #242, opened 2020-10-26
Make the upload look like the mockups

### LINKED ISSUE #51: Change style of file upload widget on work form
See [mock-up](https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420954309). What we have currently works but it's vanilla Dropzone. We need to get it closer to the style in the H2 design.

- [ ] #119
- [ ] #120

### LINKED ISSUE #119: Add file label to the form
When a file is uploaded, a new label field should be displayed:

Screen Shot 2020-10-09 at 8 22 36 AM

This will be stored in a new database field.

### LINKED ISSUE #120: Add hide file button to the form
I can choose not to publish any individual files uploaded, and confirm they are not displayed in PURL

You can choose this option with the checkbox on the right:

Screen Shot 2020-10-09 at 8 22 36 AM

This should be stored in a new database field.

jcoyne:
## Why was this change made?
Fixes #119 
Fixes #120 
Fixes #51

Screen Shot 2020-10-27 at 12 35 16 PM

### DIFF
--- a/app/javascript/controllers/dropzone_controller.js
+++ b/app/javascript/controllers/dropzone_controller.js
@@ -10,10 +10,10 @@ import {
 } from "helpers";
 
 export default class extends Controller {
-  static targets = ["input"];
+  static targets = ["input", "previews", "template"];
 
   connect() {
-    this.dropZone = createDropZone(this);
+    this.dropZone = createDropZone(this, this.templateTarget.innerHTML);
     this.hideFileInput();
     this.bindEvents();
     Dropzone.autoDiscover = false; // necessary quirk for Dropzone error in console
@@ -72,12 +72,15 @@ export default class extends Controller {
   }
 
   get addRemoveLinks() {
-    return this.data.get("addRemoveLinks") || true;
+    return this.data.get("addRemoveLinks") || false;
   }
 }
 
 class DirectUploadController {
+  // source is the input element
+  // file is the File object from dropzone
   constructor(source, file) {
+    this.count = Math.floor(Math.random() * 1_000_000_000)
     this.directUpload = createDirectUpload(file, source.url, this);
     this.source = source;
     this.file = file;
@@ -86,6 +89,8 @@ class DirectUploadController {
   start() {
     this.file.controller = this;
     this.hiddenInput = this.createHiddenInput();
+    this.addDescription();
+    this.addHideCheckbox();
     this.directUpload.create((error, attributes) => {
       if (error) {
         removeElement(this.hiddenInput);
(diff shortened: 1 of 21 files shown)


app/javascript/controllers/dropzone_controller.js
```diff
@@ -72,12 +72,15 @@ export default class extends Controller {
...
-    return this.data.get("addRemoveLinks") || true;
+    return this.data.get("addRemoveLinks") || false;
   }
 }
 
 class DirectUploadController {
+  // source is the input element
+  // file is the File object from dropzone
   constructor(source, file) {
+    this.count = Math.floor(Math.random() * 1_000_000_000)
```
mjgiarlo, line 83: What is this count for?

app/models/work.rb
```diff
@@ -9,7 +9,7 @@ class Work < ApplicationRecord
   has_many :contributors, dependent: :destroy
   has_many :related_links, dependent: :destroy
   has_many :related_works, dependent: :destroy
-  has_many_attached :files
+  has_many :attached_files, dependent: :destroy
```
mjgiarlo, line 12: Can you say more about this change?

mjgiarlo: Nevermind, I now see it's so you can hang the label and hide attrs off of attached files.

spec/models/attached_file_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+RSpec.describe AttachedFile, type: :model do
+  subject(:attached_file) { build(:attached_file, :with_file) }
+
+  it 'has a label' do
+    expect(attached_file.label).to be_present
+  end
+
+  it 'has a hide bit' do
+    expect(attached_file).not_to be_hide
```
mjgiarlo, line 14: `expect(armor).not_to be_hide`

I wonder if `be_hidden` would automagically work here?

mjgiarlo (approved):
Some questions, but looking good.

app/javascript/controllers/dropzone_controller.js
jcoyne: Need a unique identifier when you submit rails nested attributes.  It's the same sort of thing that's going on here (but timestamp doesn't have enough entropy in this case): https://github.com/sul-dlss/happy-heron/blob/main/app/javascript/controllers/nested_form_controller.js#L8

spec/models/attached_file_spec.rb
jcoyne: Nope:
```
 Failure/Error: expect(attached_file).not_to be_hidden
       expected #<AttachedFile id: nil, label: "MyString", hide: false, work_id: nil, created_at: nil, updated_at: nil> to respond to `hidden?`
```

mjgiarlo: :rage:

Issue #235, opened 2020-10-26
Manage depositors for a collection


mjgiarlo:
The creator or manager of a collection or an administrator should be able to add and remove depositors for a given collection. 

See [collection participants section of collection edit mock-up](https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930437) for the design.

PR #252, opened 2020-10-27
Add depositors to a collection

jcoyne:
## Why was this change made?

Ref #235


spec/requests/collections_spec.rb
```diff
@@ -42,7 +42,8 @@
             description: 'This is a very good collection.',
             contact_email: user.email,
             access: 'world',
-            managers: user.email
+            managers: user.email,
+            depositor_sunets: 'maya.aguirre,jcairns, cchavez, premad, giancarlo, zhengyi'
```
ndushay, line 46: fwiw, the spacing after the first comma should be a single space, not no space, eh?

ndushay (approved):
read code;  seems straightforward.  Probably best to get another review tho.  One space missing (see line comment)

spec/requests/collections_spec.rb
jcoyne: I was testing that it can handle with or without a space.

.rubocop.yml
```diff
@@ -148,6 +148,8 @@ Rails/AfterCommitOverride: # (new in 2.8)
   Enabled: true
 Rails/FindById: # (new in 2.7)
   Enabled: true
+Rails/HasAndBelongsToMany:
```
mjgiarlo, line 151: Can you say more about the rationale to flip this one off?

jcoyne: Because we don't need any additional attributes on the association: https://rails.rubystyle.guide/#has-many-through   rubocop-rails is being especially paternalistic with this one.

PR #253, opened 2020-10-28
Allow collections to be edited

### LINKED ISSUE #235: Manage depositors for a collection

jcoyne:
## Why was this change made?
Fixes #235

### DIFF
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -12,15 +12,25 @@ class CollectionForm < Reform::Form
   property :access, default: 'world'
   property :creator, writable: false
   property :depositor_sunets, virtual: true, prepopulator: lambda { |_options|
-    self.depositor_sunets = model.depositors.map { |user| user.email.delete_suffix('@stanford.edu') }
+    self.depositor_sunets = depositor_sunets_from_model.join(', ')
   }
 
   def sync(*)
     sunetids = depositor_sunets.split(/\s*,\s*/)
     emails = sunetids.map { |sunet| "#{sunet}@stanford.edu" }
-    model.depositors = emails.map { |email| User.create_or_find_by(email: email) }
+    model.depositors = emails.map do |email|
+      # It's odd that we need to do both, but this is how it's written.
+      # See: https://github.com/rails/rails/issues/36027
+      User.find_by(email: email) || User.create_or_find_by(email: email)
+    end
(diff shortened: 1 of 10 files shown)


app/forms/collection_form.rb
```diff
@@ -11,15 +11,25 @@ class CollectionForm < Reform::Form
...
   }
 
   def sync(*)
     sunetids = depositor_sunets.split(/\s*,\s*/)
     emails = sunetids.map { |sunet| "#{sunet}@stanford.edu" }
-    model.depositors = emails.map { |email| User.create_or_find_by(email: email) }
+    model.depositors = emails.map do |email|
+      # It's odd that we need to do both, but this is how it's written.
+      # See: https://github.com/rails/rails/issues/36027
+      User.find_by(email: email) || User.create_or_find_by(email: email)
```
mjgiarlo, line 23: I don't grok why we need a find as part of the second operation. Why not:

```ruby
User.find_by(email: email) || User.create(email: email)
```

mjgiarlo (approved):

app/forms/collection_form.rb
jcoyne: Your suggestion leaves the possibility of a race condition where the find fails, but another thread has created the user before our code gets to create.

mjgiarlo: OK, good catch. mind dropping a comment about this incantation being used to avoid said race condition?

app/policies/collection_policy.rb
```diff
@@ -7,11 +7,19 @@ class CollectionPolicy < ApplicationPolicy
...
   # Those who are members of the LDAP collection group may create collections
   sig { returns(T::Boolean) }
   def create?
     administrator? || collection_creator?
   end
 
+  # TODO: don't allow everyone to update collections
+  sig { returns(T::Boolean) }
+  def update?
+    true
```
mjgiarlo, line 21: Is this still right? Or should we do the same checks as in `create?`

jcoyne: That's a separate ticket, #254

PR #274, opened 2020-10-29
Document what all the classes do

jcoyne:
## Why was this change made?
Documentation is easy if you do it from the beginning.


app/helpers/application_helper.rb
```diff
@@ -1,5 +0,0 @@
-# typed: strict
-# frozen_string_literal: true
-
-module ApplicationHelper
```
mjgiarlo, line 4: Good riddance!

mjgiarlo (approved):

PR #275, opened 2020-10-29
Use a configurable header for group membership

### LINKED ISSUE #268: Unable to go to the dashboard on stage
Stage has the latest code deployed, when you click "Start here" you will be redirected back and get a message:  "You are not authorized to perform the requested action".   I've added logging and determined that the application_controller never gets any value from `request.env['eduPersonEntitlement']`  This might be something that needs configuring in puppet.

Maybe we need to set `ShibUseHeaders On`?

jcoyne:
## Why was this change made?
Fixes #268

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -26,9 +26,10 @@ class ApplicationController < ActionController::Base
 
   sig { returns(T::Array[String]) }
   def ldap_groups
-    return request.env['eduPersonEntitlement'].split(';') if request.env['eduPersonEntitlement']
-
-    []
+    raw_header = request.env[Settings.authorization_group_header]
+    raw_header = ENV['ROLES'] if Rails.env.development?
+    logger.debug("Roles are #{raw_header}")
+    raw_header&.split(';') || []
   end
 
   def deny_access
--- a/app/policies/dashboard_policy.rb
+++ b/app/policies/dashboard_policy.rb
@@ -5,6 +5,8 @@
 class DashboardPolicy < ApplicationPolicy
   sig { returns(T::Boolean) }
   def show?
-    user.application_user?
+    administrator? || collection_creator? # TODO: || has created any deposits
   end
+
+  delegate :administrator?, :collection_creator?, to: :user
 end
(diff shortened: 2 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -26,9 +26,10 @@ def current_user
 
   sig { returns(T::Array[String]) }
   def ldap_groups
-    return request.env['eduPersonEntitlement'].split(';') if request.env['eduPersonEntitlement']
-
-    []
+    raw_header = request.env[Settings.authorization_group_header]
+    raw_header = ENV['ROLES'] if Rails.env.development?
+    logger.info("Roles are #{raw_header}")
```
mjgiarlo, line 31: Maybe `logger.debug` but nbd?

app/policies/dashboard_policy.rb
```diff
@@ -5,6 +5,8 @@
 class DashboardPolicy < ApplicationPolicy
   sig { returns(T::Boolean) }
   def show?
-    user.application_user?
+    administrator? || collection_creator? # TODO: || has created any deposits
   end
+
+  delegate :administrator?, :collection_creator?, to: :user
```
mjgiarlo, line 11: If you're delegating here, why not also in the collection policy?

config/settings.yml
```diff
@@ -20,7 +20,8 @@ h2:
 # feature flag
 allow_sdr_content_changes: true
 
+authorization_group_header: HTTP_X_GROUPS
 # mappings from authorization role to workgroup name
 authorization_workgroup_names:
-  administrators: 'dlss:hydrus-app-adminstrators'
+  administrators: 'dlss:hydrus-app-administrators'
```
mjgiarlo, line 26: :1st_place_medal:

app/policies/dashboard_policy.rb
jcoyne: changed to be consistent.

mjgiarlo (approved):

config/settings.yml
jmartin-sul: whoops!

PR #284, opened 2020-10-31
Store file uploads in the configured storage location

### LINKED ISSUE #149: Request dedicated filesystem mount for deposited files
H2 will "hang on" to user-deposited files locally, just as current Hydrus does. We should store these on a dedicated, backed-up mount.

- [x] Initiate a sprint planning discussion about how much space to request in each env for this mount.
- [X] Write up a [new ops-tasks issue](https://github.com/sul-dlss/operations-tasks/issues/new/choose) requesting a mount similar to Hydrus's current mount in `/data/hydrus-files` (call it `/data/h2-files`?) for each of our deployed environments (qa, stage, prod). 
  - [X] Specify that this mount should be backed up regularly just as Hydrus's is.
- [x] If needed, specify the path to this mount in shared_configs (if it varies across environments; else stuff the value in the application repository in `config/settings.yml`)
- [x] Update [DevOpsDocs](https://github.com/sul-dlss/DevOpsDocs/blob/master/projects/h2/operations-concerns.md) with something like the following (based on Hydrus's docs) in the dependencies section:
```markdown
#### Mounted Filesystem

There is a shared mount at /data/h2-files. When a user creates a new self-deposit object, the files are put in a Druid tree on the shared mount. This mount must have available disk space. Files must remain in the mount for looking at the object through the self-deposit app in H2.
```

mjgiarlo:
Fixes #149

## Why was this change made?

This commit contains a number of pieces of getting user-uploaded files stored in the expected location:

* Pull in DruidTools gem to allow arranging file uploads attached to deposited objects in a "druid tree" directory hierarchy, which is consistent with current Hydrus practice and general DOR practice. Principle of least surprise.
* Fix bug in WorkForm that currently fails to associate AttachedFile instances with ActiveStorage blobs
* Use two storage locations:
    1. A subdirectory of a configured mount for draft objects, which is organized by ActiveStorage based on the system-assigned ActiveStorage key.
    2. A subdirectory of a configured mount for deposited objects, which is organized in a "druid tree" (see above).
* When object is successfully deposited, use StorageMigrator service to move blobs from the draft object storage location to the deposited object storage location.
* Delegate from AttachedFile to associated blob to display uploaded files' names, types, and sizes.
* Display more information about files and their attachments on the work show page
* Add and configure a custom ActiveStorage disk service for storing files in druid trees.
* Add a spec helper allowing factories to use fixture_file_upload method for convenience.

In browser and in CI.

Added configuration locally and changed shared_configs.


app/forms/work_form.rb
```diff
@@ -85,7 +85,11 @@ def sync(*)
                  attached_files.delete(item)
                  return skip!
                end
-               item || attached_files.append(AttachedFile.new)
+               return item if item
+
+               new_attached_file = AttachedFile.new
```
mjgiarlo, line 90: This changes makes a connection between AttachedFile instances and AS::Blobs that wasn't happening before.

app/forms/work_form.rb
```diff
@@ -85,7 +85,11 @@ def sync(*)
...
+               new_attached_file.file.attach(ActiveStorage::Blob.find_signed(fragment['file']))
+               attached_files.append(new_attached_file)
```
mjgiarlo, line 92: This changes makes a connection between AttachedFile instances and AS::Blobs that wasn't happening before.

app/jobs/deposit_status_job.rb
```diff
@@ -14,6 +14,7 @@ def perform(work:, job_id:)
     if result.success?
       work.druid = result.value!
       work.deposit!
+      StorageMigrator.migrate(druid: work.druid)
```
mjgiarlo, line 17: Once deposited, migrate uploaded files from draft subdir to deposited subdir. See PR description for more about this.

app/models/attached_file.rb
```diff
@@ -1,8 +1,14 @@
-# typed: strict
+# typed: false
 # frozen_string_literal: true
 
 # Models a File that is attached to a Work
 class AttachedFile < ApplicationRecord
   belongs_to :work
   has_one_attached :file
+
+  def blob
```
mjgiarlo, line 9: We use the AF instance as a proxy for the blob, and the changes to this file make it easier to display richer information about each AF.

app/models/work.rb
```diff
@@ -13,6 +13,8 @@ class Work < ApplicationRecord
   has_many :attached_files, dependent: :destroy
   has_many :keywords, dependent: :destroy
 
+  accepts_nested_attributes_for :attached_files, :contributors, :keywords
```
mjgiarlo, line 16: Helped with unpermitted parameters resulting from `attached_files_attributes` submitted to the WorkForm.

app/services/storage_migrator.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def migrate
+    ActiveStorage::Blob.service = source
+
+    attached_files = Work.find_by(druid: druid).attached_files
+    Rails.logger.info "Migrating #{attached_files.count} blobs to druid disk location for #{druid}..."
+    attached_files.each do |af|
+      Rails.logger.info "\t#{af.filename}"
+      af.blob.open do |blob|
+        target.upload(af.blob.key, blob, checksum: af.blob.checksum)
```
mjgiarlo, line 41: Note that a checksum check is baked into this file move operation. Yay for ActiveStorage.

config/settings.yml
```diff
@@ -28,3 +28,5 @@ authorization_group_header: HTTP_X_GROUPS
 authorization_workgroup_names:
   administrators: 'dlss:hydrus-app-administrators'
   collection_creators: 'dlss:hydrus-app-collection-creators'
+
+file_uploads_root: tmp/uploads
```
mjgiarlo, line 32: This is `/data/h2-files` in our deployed environments, which is already mounted.

config/storage.yml
```diff
@@ -4,7 +4,11 @@ test:
 
 local:
   service: Disk
-  root: <%= Rails.root.join("storage") %>
+  root: <%= Rails.root.join(Settings.file_uploads_root, 'draft') %>
```
mjgiarlo, line 7: Yes, this does the right thing when given relative (local/dev envs) and absolute (deployed/prod envs) paths.

spec/forms/work_form_spec.rb
```diff
@@ -6,15 +6,28 @@
 RSpec.describe WorkForm do
   subject(:form) { described_class.new(work) }
 
-  let(:work) { build(:work) }
+  let(:work) { create(:work) }
 
   describe 'populator on files' do
+    let!(:blob) do
+      ActiveStorage::Blob.create_after_upload!(
```
mjgiarlo, line 13: I spent an hour trying to find a better way to deal with AS-based models in FactoryBot but gave up.

mjgiarlo:
I'll improve test coverage on Monday.

app/jobs/deposit_status_job.rb
jcoyne: Won't depositing the files via SDR-API take care of this?

app/models/work.rb
jcoyne: It's my understanding this isn't necessary with reform. We don't use it for other types of attached things.

config/storage.yml
jcoyne: Should we just add `/draft` into the `file_uploads_root` variable?

spec/forms/work_form_spec.rb
```diff
@@ -6,15 +6,28 @@
 RSpec.describe WorkForm do
   subject(:form) { described_class.new(work) }
 
-  let(:work) { build(:work) }
+  let(:work) { create(:work) }
```
jcoyne, line 9: I'm pretty sure this is unnecessary.

spec/requests/works_spec.rb
```diff
@@ -127,7 +132,7 @@
         expect(response).to have_http_status(:found)
         work = Work.last
         expect(work.contributors.size).to eq 2
-        expect(work.attached_files.size).to eq 2
+        expect(work.attached_files.size).to eq 1
```
jcoyne, line 135: I'd like to show that we ignore destroyed files and can add more than one.

spec/forms/work_form_spec.rb
mjgiarlo: Yep. That was from a prior bit of flailing. Reverting that change.

spec/requests/works_spec.rb
mjgiarlo: I'll work on that now.

mjgiarlo: done.

config/storage.yml
mjgiarlo: No, because the draft subdir isn't the root. Maybe I'm not grokking, though.

app/jobs/deposit_status_job.rb
mjgiarlo: Not on the H2 filesystem, I don't think so.

app/models/work.rb
mjgiarlo: @jcoyne I'll try taking this line out.

mjgiarlo: done.

mjgiarlo:
@jcoyne This is ready for another review and should be :green_heart: momentarily.

app/lib/active_storage/service/druid_disk_service.rb
```diff
@@ -0,0 +1,21 @@
...
+
+# Use a Druid-based disk service
+module ActiveStorage
+  class Service
+    # Provides a druid-based tree path
+    class DruidDiskService < DiskService
+      private
+
+      def folder_for(key)
+        work = ActiveStorage::Blob.find_by(key: key).attachments.first.record.work
```
jcoyne, line 14: Should we handle the case where this happens to be called on an unattached blob?

app/models/attached_file.rb
jcoyne: What if there is no attachment here?  Should we handle that case?

app/models/work.rb
```diff
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: false
```
jcoyne, line 1: Why?

.rubocop_todo.yml
```diff
@@ -1,11 +1,21 @@
...
+# on 2020-11-02 23:35:33 UTC using RuboCop version 0.93.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 2
+# Configuration parameters: IgnoredMethods.
+Metrics/AbcSize:
+  Max: 27
```
jcoyne, line 12: I would really like to avoid setting these thresholds automatically.  Instead by wrapping each method with an exception, we avoid the case where all our methods approach the worst case.

config/storage.yml
jcoyne: Ah, I see.

jcoyne (changes requested):

jcoyne (approved):

app/models/attached_file.rb
mjgiarlo: @jcoyne that seems like an error condition, in which case we might want to raise an exception. no?

app/lib/active_storage/service/druid_disk_service.rb
mjgiarlo: Should/n't that be an error condition?

.rubocop_todo.yml
mjgiarlo: will change to an exclude if that's ok

app/models/work.rb
mjgiarlo: srb did it for me to be helpful, apparently, but I manually reverted this and srb is still happy. thanks!

app/components/works/file_row_component.html.erb
```diff
@@ -18,11 +18,12 @@
   </div>
 
   <div class="dz-details col-md-2">
+    <%= form.hidden_field :_destroy %>
     <%= form.check_box :hide, class: 'form-check-input' %>
     <%= form.label :hide, 'Hide file', class: 'form-check-label col-form-label ' %>
   </div>
 
   <div class="col-md-1">
-    <%= button_tag '🗑', class: "dz-remove pull-right btn", data: { 'dz-remove': true } %>
```
mjgiarlo, line 26: If we use dz-remove, the form elements are removed and never get submitted to the work form, which handles removing AS blobs.

mjgiarlo:
@jcoyne Ready for another look.

.rubocop_todo.yml
```diff
@@ -1,11 +1,27 @@
...
+# `rubocop --auto-gen-config --auto-gen-only-exclude`
+# on 2020-11-03 23:10:27 UTC using RuboCop version 0.93.1.
 # The point is for the user to remove these configuration records
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
+# Offense count: 2
+# Configuration parameters: IgnoredMethods, Max.
+Metrics/AbcSize:
```
jcoyne, line 11: I really like these inline, but if you prefer it here. 🤷

jcoyne: I feel that the TODO file is fine for a legacy project where you use rubocop, but a greenfield project shouldn't have one.

jcoyne (approved):
Please consider comments about rubocop_todo, but feel free to merge.

.rubocop_todo.yml
mjgiarlo: I have a slight preference for concentrating these into the todo file rather than sprinkled throughout the codebase.

jcoyne: It's been my experience, that we never do the things in the TODO.  It's also not clear which areas you are currently working in are flagged for being ripe for refactoring.

mjgiarlo: Do you think this might make for a good dev planning discussion?

PR #286, opened 2020-11-02
Deconflate groups and users

jcoyne:
No functional changes.

## Why was this change made?
A user is a persistence object, which has no concept of groups.

Test suite


jmartin-sul (approved):
nice, i like this.  i was not using quite the right devise incantations when i tried to do something like this.

PR #311, opened 2020-11-03
Cache groups in the session

### LINKED ISSUE #281: LDAP groups are not available, but "Logout" link shows.
I believe this is because we don't have a valid shibboleth cookie, but do have a valid application session cookie.  This means that when you click "Start here" you see the "you don't have permissions" message.

jcoyne:
## Why was this change made?

So that groups are available even if the shibboleth session expires.

Fixes #281

Tested on staging.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,11 +25,15 @@ class ApplicationController < ActionController::Base
   private
 
   sig { returns(T::Array[String]) }
+  # This looks first in the session for groups, and then to the headers.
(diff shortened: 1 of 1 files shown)


peetucket (approved):
I guess there is no security issues if the groups are cached in the session and the user logs out?

mjgiarlo (commented):
Do we also need code elsewhere that knows to look in the app session for groups or does/will this happen transparently?

jcoyne:
It should happen transparently.

PR #321, opened 2020-11-03
use fontawesome wastebasket

### LINKED ISSUE #280: Use fontawesome wastebasket rather than emoji

ndushay:
## Why was this change made?

Fixes #280

### Before

before trashicon change

### After

Note:  not 100% certain i got the desired trash icon -- there are a lot and I didn't understand the design doc notation as there is no `fa-trash-o`.  Also not certain if it's the right size.

after trashicon change

running locally and ensuring unit tests still pass.

### DIFF
--- a/app/components/works/contributor_row_component.html.erb
+++ b/app/components/works/contributor_row_component.html.erb
@@ -1,7 +1,9 @@
 <div class="contributor row inner-container">
   <div class="col-md-12">
-    <%= button_tag "🗑", class: 'btn btn-sm float-right', aria: { label: 'Remove' },
-        data: { action: "click->nested-form#remove_association" } %>
+    <%= button_tag class: 'btn btn-sm float-right', aria: { label: 'Remove' },
+        data: { action: "click->nested-form#remove_association" } do %>
+      <span class="far fa-trash-alt"></span>
+    <% end %>
   </div>
   <div class="col-md-4">
     <%= form.hidden_field :_destroy %>
--- a/app/components/works/file_row_component.html.erb
+++ b/app/components/works/file_row_component.html.erb
@@ -24,6 +24,8 @@
   </div>
 
   <div class="col-md-1">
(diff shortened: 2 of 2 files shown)


jcoyne (changes requested):
These seem a little large compared to what was asked for in the mockups.  Have you considered dropping the `fa-2x` style?

Screen Shot 2020-11-03 at 7 41 38 PM

ndushay:
@jcoyne tweaked to diff fontawesome trash can:

image

app/components/works/contributor_row_component.html.erb
```diff
@@ -1,7 +1,9 @@
 <div class="contributor row inner-container">
   <div class="col-md-12">
-    <%= button_tag "🗑", class: 'btn btn-sm float-right', aria: { label: 'Remove' },
-        data: { action: "click->nested-form#remove_association" } %>
+    <%= button_tag class: 'btn btn-sm float-right', aria: { label: 'Remove' },
+        data: { action: "click->nested-form#remove_association" } do %>
+      <span class="far fa-trash-alt"></span>
```
mjgiarlo, line 5: Does the `far` class mean we are using FontAwesome pro? https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use seems to suggest it. (I didn't know we had pro!) Although looking at this icon, it says it's in the regular set but it's also free? https://fontawesome.com/icons/trash-alt?style=regular I'm confused. (But this PR is A-OK.)

ndushay: I think the s in "fas" may stand for "solid" and the r in "far" may stand for regular.  I got the info from here:

image

ndushay: image

mjgiarlo: I think you're right about what r and s mean. I didn't get the impression that the regular icons were deprecated. But the FA documentation on whether or not the regular icons are free or pro or both twisted my wig real good. Not important, though. What's important: the icon matches what's in the design doc!

jcoyne (approved):

PR #325, opened 2020-11-04
Renames Multimedia with mixed materials in Work Types

### LINKED ISSUE #271: Rename multimedia to mixed materials.

jermnelson:
## Why was this change made?

fixes #271 

unit tests

n/a

### DIFF
--- a/app/models/work_type.rb
+++ b/app/models/work_type.rb
@@ -135,7 +135,7 @@ class WorkType
     [
       new(id: 'data', label: 'Data', icon: 'chart-bar', subtypes: DATA_TYPES),
       new(id: 'image', label: 'Image', icon: 'images', subtypes: IMAGE_TYPES),
-      new(id: 'mixed materials', label: 'Mixed Materials', icon: 'play', subtypes: MIXED_TYPES),
+      new(id: 'mixed material', label: 'Mixed Materials', icon: 'play', subtypes: MIXED_TYPES),
       new(id: 'other', label: 'Other', icon: 'archive', subtypes: []),
       new(id: 'software, multimedia', label: 'Software or Code', icon: 'mouse', subtypes: SOFTWARE_TYPES),
       new(id: 'sound', label: 'Sound', icon: 'microphone-alt', subtypes: SOUND_TYPES),
--- a/app/models/work_type.rb
+++ b/app/models/work_type.rb
@@ -135,9 +135,9 @@ class WorkType
     [
       new(id: 'data', label: 'Data', icon: 'chart-bar', subtypes: DATA_TYPES),
       new(id: 'image', label: 'Image', icon: 'images', subtypes: IMAGE_TYPES),
-      new(id: 'mixed material', label: 'Multimedia', icon: 'play', subtypes: MIXED_TYPES),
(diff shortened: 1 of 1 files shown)


app/models/work_type.rb
```diff
@@ -134,9 +134,9 @@ def self.all
     [
       new(id: 'data', label: 'Data', icon: 'chart-bar', subtypes: DATA_TYPES),
       new(id: 'image', label: 'Image', icon: 'images', subtypes: IMAGE_TYPES),
-      new(id: 'mixed material', label: 'Multimedia', icon: 'play', subtypes: MIXED_TYPES),
+      new(id: 'mixed materials', label: 'Mixed Materials', icon: 'play', subtypes: MIXED_TYPES),
       new(id: 'other', label: 'Other', icon: 'archive', subtypes: []),
-      new(id: 'software, multimedia', label: 'Software or Code', icon: 'mouse', subtypes: SOFTWARE_TYPES),
+      new(id: 'software, mixed materials', label: 'Software or Code', icon: 'mouse', subtypes: SOFTWARE_TYPES),
```
mjgiarlo, line 139: @jcoyne reading the associated ticket and looking at https://docs.google.com/document/d/1tEdGP3tk-IG9ICsaN2ilMhn0UWu1pM7-Y6QFqMtVkoM/edit, it's not clear to me whether this change (to software) is included in what we want. I don't see any guidance related to the IDs in this list in the documentation, and I don't see any information in the doc that would suggest we need anything other than, say, `software` here. Pls advise?

jcoyne: This is a mods resource type, and it shouldn't change: https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/h2_cocina_mappings/h2_to_cocina_form.txt#L162

jcoyne (changes requested):

app/models/work_type.rb
jermnelson: Restored `multimedia` for **Software or Code** in latest commit

mjgiarlo (approved):

app/models/work_type.rb
```diff
@@ -135,7 +135,7 @@ def self.all
     [
       new(id: 'data', label: 'Data', icon: 'chart-bar', subtypes: DATA_TYPES),
       new(id: 'image', label: 'Image', icon: 'images', subtypes: IMAGE_TYPES),
-      new(id: 'mixed material', label: 'Multimedia', icon: 'play', subtypes: MIXED_TYPES),
+      new(id: 'mixed materials', label: 'Mixed Materials', icon: 'play', subtypes: MIXED_TYPES),
```
jcoyne, line 138: Please don't change the id, that's a MODS term: https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/8d8851203a026dec1eff8021cc14ec43805cdfa0/mods_cocina_mappings/mods_to_cocina_typeOfResource.txt#L67

jcoyne (changes requested):

app/models/work_type.rb
jermnelson: Reverted in latest commit.

jcoyne (approved):

PR #326, opened 2020-11-04
Validate that at least one keyword is provided.

### LINKED ISSUE #293: Validate that at least one keyword is provided

jcoyne:
## Why was this change made?

Fixes #293

### DIFF
--- a/app/components/works/work_form_component.html.erb
+++ b/app/components/works/work_form_component.html.erb
@@ -1,8 +1,9 @@
 <div id="editor">
   <h1 class="row">Deposit your work</h1>
-
-  <%= form_with model: work_form, url: [work_form.model.collection, work_form], data: { controller: 'progress' },
-      html: { class: 'needs-validation row', novalidate: true, multipart: true } do |f| %>
+  <%= form_with model: work_form, url: [work_form.model.collection, work_form],
+      local: true,
+      data: { controller: 'progress' },
+      html: { class: "needs-validation row #{'was-validated' if work_form.errors.present?}", novalidate: true, multipart: true } do |f| %>
     <%= f.hidden_field :work_type %>
     <%= render Works::DepositProgressComponent.new %>
     <%= render Works::AddFilesComponent.new(form: f) %>
--- a/app/javascript/stylesheets/keywords.scss
+++ b/app/javascript/stylesheets/keywords.scss
@@ -1,5 +1,7 @@
 .keywords {
   .keywords-container {
+    @extend .form-control;
+
     box-sizing: border-box;
     margin: 0;
     position: relative;
@@ -11,8 +13,8 @@
   }
 
   .keywords-container-open .selection {
-    border-color: $input-focus-border-color;
-    box-shadow: 0 0 0 $input-focus-width rgba($primary, $input-btn-focus-color-opacity);
+    // border-color: $input-focus-border-color;
+    // box-shadow: 0 0 0 $input-focus-width rgba($primary, $input-btn-focus-color-opacity);
   }
 
   .keywords-dropdown {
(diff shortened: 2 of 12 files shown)


app/components/works/work_form_component.html.erb
```diff
@@ -1,8 +1,9 @@
 <div id="editor">
   <h1 class="row">Deposit your work</h1>
-
-  <%= form_with model: work_form, url: [work_form.model.collection, work_form], data: { controller: 'progress' },
-      html: { class: 'needs-validation row', novalidate: true, multipart: true } do |f| %>
+  <%= form_with model: work_form, url: [work_form.model.collection, work_form],
+      local: true,
```
mjgiarlo, line 4: Can you say more about converting this to a local form?

app/javascript/stylesheets/keywords.scss
```diff
@@ -11,8 +13,8 @@
   }
 
   .keywords-container-open .selection {
-    border-color: $input-focus-border-color;
-    box-shadow: 0 0 0 $input-focus-width rgba($primary, $input-btn-focus-color-opacity);
+    // border-color: $input-focus-border-color;
```
mjgiarlo, line 16: Keep these hanging around?

config/initializers/customize_error.rb
```diff
@@ -0,0 +1,12 @@
+# typed: false
+# frozen_string_literal: true
+
+ActionView::Base.field_error_proc = proc do |html_tag, _instance|
```
mjgiarlo, line 4: Can you drop a comment here-ish about what this is used for/why we need it?

mjgiarlo (commented):
Some questions

app/javascript/stylesheets/keywords.scss
jcoyne: Yes, until we get typeahead working.

app/components/works/work_form_component.html.erb
jcoyne: The default is to do a remote form, however, our controller was not set up to respond with XHR requests.  By default in rails if you generate scaffold, it makes a local form like this.

mjgiarlo: And it's your preference to convert this to a local form instead of tweaking the controller to respond to XHR? I'm wondering if folks will be confused to have a mix of local and remote forms.

jcoyne: Do we have a mix?  Currently we just have a form that does not support validations. So if you submit and there is a validation error, you get no feedback.

config/initializers/customize_error.rb
jcoyne: added

app/components/works/work_form_component.html.erb
mjgiarlo: @jcoyne the collecton form is a remote form. I guess I've enjoyed the remote forms in my work on EDTF validation. How about we stick with remote forms for now to avoid the page refresh? Are you open to that?

config/initializers/customize_error.rb
mjgiarlo: thank you

app/javascript/stylesheets/keywords.scss
mjgiarlo: ok

app/components/works/work_form_component.html.erb
jcoyne: Are you confusing the client side validation with a remote for?  Those are two different things.

mjgiarlo: @jcoyne yes, I am.

mjgiarlo (approved):

PR #327, opened 2020-11-04
Copy publication_edtf to the cocina model

### LINKED ISSUE #222: DescriptionGenerator should handle published_edtf
We should be able to see this value in the MODS.

jcoyne:
## Why was this change made?

Fixes #222 

Unit test

### DIFF
--- a/spec/factories/works.rb
+++ b/spec/factories/works.rb
@@ -13,6 +13,14 @@ FactoryBot.define do
     collection
   end
 
+  trait :published do
+    published_edtf { '2020-02-14' }
+  end
+
+  trait :with_creation_dates do
+    created_edtf { '2020-03-04/2020-10-31' }
+  end
+
   trait :with_contributors do
     transient do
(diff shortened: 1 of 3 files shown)


spec/factories/works.rb
```diff
@@ -13,6 +13,14 @@
     collection
   end
 
+  trait :published do
+    published_edtf { '2020-02-14' }
+  end
+
+  trait :with_creation_dates do
+    created_edtf { '2020-03-04/2020-10-31' }
```
mjgiarlo, line 21: This is A-OK. I'm mostly curious why you chose to do this as a trait instead of defaulting this above. To facilitate testing works w/o creation dates?

mjgiarlo (approved):
LGTM. Have a non-blocking question.

spec/factories/works.rb
mjgiarlo: Sounds good

PR #339, opened 2020-11-04
Validate work types and subtypes

### LINKED ISSUE #114: Add Work Subtype Validator
Ensure they are members of a set that is dependent on the work type.

https://docs.google.com/document/d/1tEdGP3tk-IG9ICsaN2ilMhn0UWu1pM7-Y6QFqMtVkoM/edit

mjgiarlo:
Fixes #114

## Why was this change made?

As part of this work, we need two different sorts of validations: one that validates a model instance, and one that can be called from the controller context. We need the latter validation because the new work form is supplied these values via its params, so there is no model instance to validate at that time.

CI

None


ndushay (approved):
LGTM

.rubocop_todo.yml
```diff
@@ -1,19 +1,20 @@
...
 # one by one as the offenses are removed from the code base.
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 2
+# Offense count: 3
 # Configuration parameters: IgnoredMethods, Max.
 Metrics/AbcSize:
   Exclude:
+    - 'app/controllers/works_controller.rb'
```
jcoyne, line 13: I really wish we knew which method had this problem. Putting inline would help.

mjgiarlo: I reckon that if we are shooting to have classes that are relatively small and single-purpose, it will be relatively easy to tell what the offending method is. I'm happy to try a refactoring to remove this exclusion altogether. I'd rather do that than do an inline exclusion.

spec/factories/works.rb
```diff
@@ -5,7 +5,7 @@
   factory :work do
     title { 'Test title' }
     work_type { 'text' }
-    subtype { ['3D model', 'GIS'] }
+    subtype { %w[Article Essay] }
```
jcoyne, line 8: The idea here was to test something with whitespace. How about `Presentation slides`?

jcoyne (changes requested):

spec/factories/works.rb
mjgiarlo: Will do.

mjgiarlo: @jcoyne done. take another look?

jcoyne (approved):

PR #342, opened 2020-11-04
Raise custom exception when deposit status not available

### LINKED ISSUE #233: Raise custom exception when retrying deposit
In `DepositStatusJob`, raise a custom exception (`TryAgainLater`) when there is no result yet:

https://github.com/sul-dlss/happy-heron/blob/21359943aa15f095b6aba77ba48e19031db484f8/app/jobs/deposit_status_job.rb#L12

This exception is raised to force sidekiq to retry with exponential backoff. It is not meant to be logged to Honeybadger. As part of this work, add a new HB exception ignore for `TryAgainLater` (to shared_configs for all envs).

## Backtrace

    line 12 of [PROJECT_ROOT]/app/jobs/deposit_status_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/69184146)

mjgiarlo:
Fixes #233

## Why was this change made?

This exception is used to trigger Sidekiq retries and is not useful for reporting to Honeybadger. By using a custom exception class, we can ignore it in HB config. 

CI

Corresponding shared_configs work has already been done.

### DIFF
--- a/app/exceptions/try_again_later.rb
+++ b/app/exceptions/try_again_later.rb
@@ -1,0 +1,9 @@
+# typed: true
+# frozen_string_literal: true
+
+# Custom exception class to special-case exceptions raised in the
+# DepositStatusJob related to deposits not yet being done. We use these
+# exceptions to signal to Sidekiq to restart the job after incremental backoff.
+# The special-casing is helpful because it allows us to ignore these exceptions
+# in Honeybadger.
+class TryAgainLater < RuntimeError; end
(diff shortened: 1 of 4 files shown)


app/exceptions/try_again_later.rb
```diff
@@ -0,0 +1,4 @@
+# typed: true
+# frozen_string_literal: true
+
+class TryAgainLater < RuntimeError; end
```
jcoyne, line 4: Can you add documentation here?

jcoyne (changes requested):

app/exceptions/try_again_later.rb
mjgiarlo: @jcoyne yep!

mjgiarlo: done

ndushay (approved):

jcoyne (approved):

PR #346, opened 2020-11-05
Organization roles only have a full name

### LINKED ISSUE #294: Combine name entries into a single name field for non-personal names
When the role or the name is not a personal name, switch to a single name field, not a "first name" and "last name". This will accommodate the organizations/events.   

In either case, a name entry is required.  So for personal names, the first and last name are both required.  For non personal names, the single name field is required.

jcoyne:
## Why was this change made?
Fixes #294

contributors

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -70,8 +70,18 @@ class WorkForm < Reform::Form
                if fragment['_destroy'] == '1'
                  contributors.delete(item)
                  return skip!
-               elsif fragment['first_name'].blank?
+               elsif fragment['first_name'].blank? && fragment['full_name'].blank?
                  return skip!
+               end
+
+               # Clear out names that we don't want to store (e.g. first & last name for an organization)
+               # These can get submitted to the server if the user enters a person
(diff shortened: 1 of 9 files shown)


app/forms/work_form.rb
```diff
@@ -70,15 +70,22 @@ def sync(*)
                if fragment['_destroy'] == '1'
                  contributors.delete(item)
                  return skip!
-               elsif fragment['first_name'].blank?
+               elsif fragment['first_name'].blank? && fragment['full_name'].blank?
                  return skip!
                end
+               if fragment['role_term'].start_with?('person')
```
mjgiarlo, line 76: Can you toss a comment here about what's going on? I grok what's going on, but I reckon we'll lose sight of this over time.

mjgiarlo (approved):
Toss a comment in the work form and then feel free to merge.

app/forms/work_form.rb
jcoyne: Good call. I've added it.

mjgiarlo (approved):
merge when :green_circle:

PR #356, opened 2020-11-05
Convert keywords component validation to remote form

mjgiarlo:
Peek 2020-11-05 11-27

## Why was this change made?

This is a spike for the team to be able to evaluate the two approaches: local forms vs remote forms. This PR switches from a local form to a remote form:

* local form: this is the standard rails form. when submitted, the request goes to the server and the server will send back to the client a rendered form (if errors) or redirect to a show page, typically
* remote form: this is the new default in rails. when submitted, turbolinks submits the form via XHR, the server does validation, and if validator errors are present, either a partial is rendered and sent or a JSON representation of errors are sent, and you use JS (Stimulus for us) to parse the errors and show them on the page.

Long story short? new-style rails convention prevents unnecessary page loads (more reactive/quicker user experience), providing a more modern user experience, at the cost of needing some JavaScript glue code. It's also arguably more API-oriented. 

In browser and CI

None.

### DIFF
--- a/app/components/works/keywords_component.html.erb
+++ b/app/components/works/keywords_component.html.erb
@@ -1,5 +1,5 @@
-<div class="keywords" data-controller="keywords" data-action="focusout->keywords#close">
-  <div class="keywords-container <%= 'is-invalid' if errors.present? %>" data-target="keywords.container" data-action="click->keywords#open">
+<div class="keywords" data-controller="keywords" data-action="focusout->keywords#close" data-target="progress.keywordsField">
+  <div class="keywords-container data-target="keywords.container" data-action="click->keywords#open">
     <span class="selection" role="combobox" aria-haspopup="true" aria-expanded="true" tabindex="-1"
       aria-disabled="false" aria-owns="keyword-results">
       <ul class="rendered" id="keyword-container" data-target="keywords.addItem">
@@ -25,9 +25,7 @@
       </span>
     </span>
   </div>
-  <% if errors.present? %>
-    <div class="invalid-feedback"><%= errors %></div>
-  <% end %>
+  <div class="invalid-feedback"></div>
 <%# -- TODO for autocomplete
   <div class="keywords-container keywords-container-open" data-target="keywords.container">
     <span class="keywords-dropdown select2-dropdown--above">
--- a/app/components/works/keywords_component.html.erb
+++ b/app/components/works/keywords_component.html.erb
@@ -1,5 +1,5 @@
 <div class="keywords" data-controller="keywords" data-action="focusout->keywords#close" data-target="progress.keywordsField">
-  <div class="keywords-container data-target="keywords.container" data-action="click->keywords#open">
+  <div class="keywords-container" data-target="keywords.container" data-action="click->keywords#open">
     <span class="selection" role="combobox" aria-haspopup="true" aria-expanded="true" tabindex="-1"
       aria-disabled="false" aria-owns="keyword-results">
       <ul class="rendered" id="keyword-container" data-target="keywords.addItem">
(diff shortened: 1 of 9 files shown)


app/components/works/keywords_component.html.erb
```diff
@@ -1,5 +1,5 @@
-<div class="keywords" data-controller="keywords" data-action="focusout->keywords#close">
-  <div class="keywords-container <%= 'is-invalid' if errors.present? %>" data-target="keywords.container" data-action="click->keywords#open">
+<div class="keywords" data-controller="keywords" data-action="focusout->keywords#close" data-target="progress.keywordsField">
```
mjgiarlo, line 1: Allow progress controller to target the keywords containing div. Progress controller is scoped to the entire work form, so we can use this pattern and the progress controller for more fields in the future if we wish.

app/components/works/keywords_component.html.erb
```diff
@@ -1,5 +1,5 @@
-<div class="keywords" data-controller="keywords" data-action="focusout->keywords#close">
-  <div class="keywords-container <%= 'is-invalid' if errors.present? %>" data-target="keywords.container" data-action="click->keywords#open">
```
mjgiarlo, line 2: Remove class because errors are no longer checked and rendered on the server in this partial.

app/components/works/keywords_component.html.erb
```diff
@@ -25,9 +25,7 @@
       </span>
     </span>
   </div>
-  <% if errors.present? %>
-    <div class="invalid-feedback"><%= errors %></div>
-  <% end %>
+  <div class="invalid-feedback"></div>
```
mjgiarlo, line 28: Render an empty div to allow JavaScript to populate as needed.

app/components/works/work_form_component.html.erb
```diff
@@ -1,8 +1,10 @@
 <div id="editor">
   <h1 class="row">Deposit your work</h1>
   <%= form_with model: work_form, url: [work_form.model.collection, work_form],
-      local: true,
-      data: { controller: 'progress' },
+      data: {
+        controller: 'progress',
+        action: 'ajax:error->progress#displayErrors'
```
mjgiarlo, line 6: `ajax:error` and `ajax:success` are the hooks used by Rails (Turbolinks? Rails-UJS?). Use the error hook to display server-generated errors in the form without a page reload.

app/javascript/controllers/progress_controller.js
```diff
@@ -21,7 +21,18 @@ export default class extends Controller {
     if (stepName === 'file') {
       isComplete = document.querySelectorAll('[type=hidden][name="work[files][]"]').length > 0
     }
-
     step.classList.toggle('active', isComplete)
   }
+
+  displayErrors(event) {
```
mjgiarlo, line 27: Here's the JS glue needed to render the server-generated errors on the client side.

app/components/works/work_form_component.html.erb
```diff
@@ -1,8 +1,10 @@
 <div id="editor">
   <h1 class="row">Deposit your work</h1>
   <%= form_with model: work_form, url: [work_form.model.collection, work_form],
-      local: true,
```
mjgiarlo, line 4: Removing `local: true` converts this to a "remote form" (https://guides.rubyonrails.org/working_with_javascript_in_rails.html).

jcoyne (commented):
👍  going this route.  This seems slightly less jarring than the existing validation on keywords.

jcoyne (approved):

app/javascript/controllers/keywords_controller.js
```diff
@@ -19,6 +19,8 @@ export default class extends Controller {
 
     this.searchTarget.value = ''
     this.addItemTarget.insertAdjacentHTML('beforeend', content)
+    // Remove error indications when keyword(s) added
+    this.containerTarget.classList.remove('is-invalid')
```
mjgiarlo, line 23: I added this after I created the animated gif. Now the keywords field no longer looks like an error when the user enters a new keyword.

PR #358, opened 2020-11-05
Allow a draft deposit to be edited

### LINKED ISSUE #347: Create a draft and resume a deposit
As a depositor I should be able to create a draft.
Later, I should be able to visit the site, find the draft and resume editing.
The form should have all the values I saved before.

jcoyne:
## Why was this change made?

Fixes #347

### DIFF
--- a/app/components/works/embargo_component.html.erb
+++ b/app/components/works/embargo_component.html.erb
@@ -39,7 +39,7 @@
   <div class="mb-3 row">
     <%= form.label :access, 'Who can access?', class: 'col-sm-2 col-form-label' %>
     <div class="col-sm-10">
-      <%= form.select :access, [['Stanford Community', 'stanford'], ['Everyone', 'world']], { selected: "", disabled: "", prompt: "Select..." }, class: "form-select", required: true %>
+      <%= form.select :access, [['Stanford Community', 'stanford'], ['Everyone', 'world']], { disabled: "", prompt: "Select..." }, class: "form-select", required: true %>
     </div>
   </div>
 </section>
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -8,6 +8,9 @@ class DashboardsController < ApplicationController
 
   def show
     authorize! :dashboard
-    @collections = authorized_scope(Collection.all)
+    @presenter = DashboardPresenter.new(
+      collections: authorized_scope(Collection.all),
+      drafts: Work.where(state: 'first_draft')
+    )
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,6 +10,10 @@ class User < ApplicationRecord
             uniqueness: { case_sensitive: false }
 
   has_many :notifications, dependent: :destroy
+  has_many :deposits, class_name: 'Work',
+                      foreign_key: 'depositor_id',
+                      inverse_of: :depositor,
+                      dependent: :destroy
 
   devise :remote_user_authenticatable
 
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -3,12 +3,19 @@
 
 # Authorization policy for Work objects
 class WorkPolicy < ApplicationPolicy
-  sig { returns(T::Boolean) }
+  alias_rule :edit?, to: :update?
 
   # Only depositors in a specific collection are able to create new collection members
+  sig { returns(T::Boolean) }
   def create?
     collection = record.collection
     collection.depositor_ids.include?(user.id) ||
       collection.managers.include?(user.email.delete_suffix('@stanford.edu'))
+  end
+
+  # Only the depositor may edit/update a work
+  sig { returns(T::Boolean) }
+  def update?
+    record.depositor == user
   end
 end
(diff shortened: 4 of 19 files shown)


app/components/works/embargo_component.html.erb
```diff
@@ -39,7 +39,7 @@
   <div class="mb-3 row">
     <%= form.label :access, 'Who can access?', class: 'col-sm-2 col-form-label' %>
     <div class="col-sm-10">
-      <%= form.select :access, [['Stanford Community', 'stanford'], ['Everyone', 'world']], { selected: "", disabled: "", prompt: "Select..." }, class: "form-select", required: true %>
+      <%= form.select :access, [['Stanford Community', 'stanford'], ['Everyone', 'world']], { disabled: "", prompt: "Select..." }, class: "form-select", required: true %>
```
mjgiarlo, line 42: Can you say more about this change?

jcoyne: This was forcing it to reset to the prompt option.  We want it to use the value the user selected originally.

app/controllers/dashboards_controller.rb
```diff
@@ -8,6 +8,9 @@ class DashboardsController < ApplicationController
 
   def show
     authorize! :dashboard
-    @collections = authorized_scope(Collection.all)
+    @presenter = DashboardPresenter.new(
+      collections: authorized_scope(Collection.all),
+      drafts: Work.where(state: 'first_draft')
```
mjgiarlo, line 13: You can also use `Work.with_state(:first_draft)` here, I believe. our state machine library generates these scopes for us. Not a big deal either way.

app/models/user.rb
```diff
@@ -10,6 +10,10 @@ class User < ApplicationRecord
             uniqueness: { case_sensitive: false }
 
   has_many :notifications, dependent: :destroy
+  has_many :deposits, class_name: 'Work',
+                      foreign_key: 'depositor_id',
+                      inverse_of: :depositor,
+                      dependent: :destroy
```
mjgiarlo, line 16: Does this mean if we destroy a user instance, we lose all their works?

app/policies/work_policy.rb
```diff
@@ -3,12 +3,19 @@
...
   def create?
     collection = record.collection
     collection.depositor_ids.include?(user.id) ||
       collection.managers.include?(user.email.delete_suffix('@stanford.edu'))
   end
+
+  # Only the depositor may edit/update a work
+  sig { returns(T::Boolean) }
+  def update?
+    record.depositor == user
```
mjgiarlo, line 19: Do we have a use case for collections managers to be able to edit/update works deposited by other users in their collection? @amyehodge (Not saying we need to have that in this PR, but it would be easy to add?)

app/presenters/dashboard_presenter.rb
```diff
@@ -0,0 +1,19 @@
+# typed: strict
+# frozen_string_literal: true
+
+# The page model for the dashboard
+class DashboardPresenter
+  extend T::Sig
+
+  sig { params(drafts: ActiveRecord::Relation, collections: ActiveRecord::Relation).void }
```
mjgiarlo, line 8: Can you say more about the distinction between a presenter and a view component, conceptually?

spec/factories/works.rb
```diff
@@ -27,7 +28,7 @@
       keywords_count { 3 }
     end
 
-    keywords { Array.new(keywords_count) { association(:keyword) } }
+    keywords { build_list(:keyword, keywords_count) }
```
mjgiarlo, line 31: Nice.

app/components/works/embargo_component.html.erb
mjgiarlo: :beers:

app/models/user.rb
jcoyne: Yes.

app/policies/work_policy.rb
amyehodge: Definitely. This is a must-have.

jcoyne: I would say that is a question for a new ticket.

app/presenters/dashboard_presenter.rb
jcoyne: A presenter is a model of the data and a view component is a model of the view. This comes from Sandi Metz's rules of thumb:
> Controllers can instantiate only one object. Therefore, views can only know about one instance variable and views should only send messages to that object (@object.collaborator.value is not allowed).

spec/factories/works.rb
jcoyne: The other way fails as it tries to persist the keyword before persisting the work.

app/models/user.rb
mjgiarlo: We'll have db backups if we need them, I reckon.

app/policies/work_policy.rb
jcoyne: https://github.com/sul-dlss/happy-heron/issues/363

app/presenters/dashboard_presenter.rb
mjgiarlo: So, for instance, we wouldn't see a component instantiated from within a controller. But we would a presenter. OK.

mjgiarlo (approved):

PR #359, opened 2020-11-05
License grouping

### LINKED ISSUE #239: Add hierarchy to license dropdown in work edit form
See hierarchy in [H2 design doc](https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.1l4doxwpm8vw) (radio button selections > dropdown values):

As part of this work, update any license labels in the app that do not match the labels above.

To do this, see related `grouped_options` work in `RoleTerm` PR: https://github.com/sul-dlss/happy-heron/commit/19eac835e75a20467b2e88b886ddd9f18892eabf

ndushay:
Fixes #239

### Before

image

### After

image

bonus:  you can still create a work!

- wrote a couple of unit tests as I was getting license methods set up
- ran it locally on my machine (see pix above)


app/models/license.rb
```diff
@@ -1,13 +1,16 @@
...
 class License
   extend T::Sig
 
+  # identifier from https://spdx.org/licenses/
   sig { returns(String) }
   attr_reader :id
 
+  # displayed to user in work form
   sig { returns(String) }
   attr_reader :label
```
ndushay, line 15: 1. does it make sense to keep `:label` as an attribute as we can simply derive it from the hash?
2. I assumed the work would keep the "id" (license code), and I did the validation a bit differently from before, without building all the diff License objects.  Is there any problem with this?

Basically - I wasn't sure what was supposed to happen next and if this approach will fooch that somehow.

app/models/license.rb
```diff
@@ -1,13 +1,16 @@
...
 
-# Represents the list of valid work types
+# Class contains valid license types, their identifiers from https://spdx.org/licenses/
+#  and other structures used in the work form and its validation
 class License
   extend T::Sig
 
+  # identifier from https://spdx.org/licenses/
   sig { returns(String) }
   attr_reader :id
```
jcoyne, line 11: Since you removed all the `new()` calls, I think you can remove the attributes too.

jcoyne (changes requested):

app/models/license.rb
jcoyne: I think you should remove both methods as they are no longer in use.

ndushay: do we even want the constructor?

jcoyne: We don't need it since you did away with the calls to it.

jcoyne (approved):

PR #368, opened 2020-11-06
make email validation more strict client/server side

### LINKED ISSUE #332: Validation of contact email on collection form needs to be improved.
(Sorry for the kinda lame title. I found it hard to summarize this).

As I type in an email address, the validation switches back and forth from doesn't validate to validates several times. It should not validate until the last screenshot -- if then, as this is still not technically valid until "edu" is completed.

1. 

2. 

3. 

4.

peetucket:
## Why was this change made?

Fixes #332 

* create an email regex to require a pattern like this XXX@XXX.XX  (some characters ending with exactly one @ followed by some characters (e.g. a subdomain) followed by exactly one . followed by at least two characters (e.g. a top level domain)
* put this email regex in a single config location (just took over the Devise config)
* use this configured regex in all current locations where we validate email addresses (client or server side)
* add this validation to the Reform models as well (where they were not before) <-- not sure if this is needed?

Localhost browser

### DIFF
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -26,7 +26,8 @@ class CollectionForm < Reform::Form
     super
   end
 
-  validates :name, :description, :contact_email, :managers, :access, presence: true
+  validates :name, :description, :managers, :access, presence: true
+  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
 
   private
 
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -60,6 +60,7 @@ class WorkForm < Reform::Form
             allow_nil: true
   validates 'release', presence: true, inclusion: { in: %w[immediate embargo] }
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
+  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
 
   collection :contributors,
              populator: lambda { |fragment:, **|
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -184,7 +184,7 @@ Devise.setup do |config|
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
   # to give user feedback and not to assert the e-mail validity.
-  config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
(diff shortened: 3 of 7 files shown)


app/forms/collection_form.rb
```diff
@@ -26,7 +26,8 @@ def sync(*)
     super
   end
 
-  validates :name, :description, :contact_email, :managers, :access, presence: true
+  validates :name, :description, :managers, :access, presence: true
+  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
```
mjgiarlo, line 30: Can/should we remove the corresponding model validation now that we have the form validation? Cc: @jcoyne

app/forms/work_form.rb
```diff
@@ -60,6 +60,7 @@ def sync(*)
             allow_nil: true
   validates 'release', presence: true, inclusion: { in: %w[immediate embargo] }
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
+  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
```
mjgiarlo, line 63: Can/should we remove the corresponding model validation now that we have the form validation? Cc: @jcoyne

config/initializers/devise.rb
```diff
@@ -184,7 +184,7 @@
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
   # to give user feedback and not to assert the e-mail validity.
-  config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
+  config.email_regexp = /[a-z0-9._%+-]+@[a-z0-9-]+\.[a-z]{2,63}/
```
mjgiarlo, line 187: Can you add a comment here saying why we have customized this?

mjgiarlo (commented):
Just a few small questions.

config/initializers/devise.rb
peetucket: comment added

mjgiarlo: thanks!

app/forms/collection_form.rb
jcoyne: I don't care. This seems reasonable to have in both places.

app/forms/work_form.rb
jcoyne: I don't care. This seems reasonable to have in both places.

jcoyne (approved):

PR #390, opened 2020-11-09
Display the drafts for the currently logged in user

jcoyne:
## Why was this change made?

Drafts should only show for the user who created them.

### DIFF
--- a/spec/requests/dashboards_spec.rb
+++ b/spec/requests/dashboards_spec.rb
@@ -8,7 +8,7 @@ RSpec.describe 'Dashboard requests' do
 
   before { sign_out }
 
-  context 'when user has no workgroup' do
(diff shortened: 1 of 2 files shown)


spec/requests/dashboards_spec.rb
```diff
@@ -8,7 +8,7 @@
 
   before { sign_out }
 
-  context 'when user has no workgroup' do
+  context 'when user has no deposits' do
```
mjgiarlo, line 11: Was the prior test bogus?

jcoyne: Yes, it comes from when we thought there was a group that all users had to be part of.

mjgiarlo (approved):

PR #393, opened 2020-11-09
map contributors from work form into cocina model for deposit

### LINKED ISSUE #240: Map contributors into cocina model
When done, you should see any authors and contributors added for a work mapped into descMD in Argo, with the proper role term specified.

See [mapping](https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/mods_cocina_mappings/mods_to_cocina_name.txt#L168-L199)

May require upstream mapping work, not sure. Test in QA.

ndushay:
Part of #240 -- waiting for Arcadia to provide more complete mappings for roles.

## Why was this change made?

The contributor data from the work form needs to be saved as descriptive metadata in a deposit.

testing locally, running it on stage and seeing that the contributor mappings are being passed into sdr-api, added unit test and tweaked unit test.

### DIFF
--- a/app/services/description_generator.rb
+++ b/app/services/description_generator.rb
@@ -2,6 +2,7 @@
 # frozen_string_literal: true
 
 # This generates a RequestDRO Description for a work
+# rubocop:disable Metrics/ClassLength
 class DescriptionGenerator
   extend T::Sig
 
@@ -19,6 +20,7 @@ class DescriptionGenerator
   def generate
     {
       title: title,
+      contributor: contributors,
       subject: keywords,
       note: [abstract, citation, contact],
       event: [created_date, published_date].compact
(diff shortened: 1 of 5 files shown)


app/services/description_generator.rb
```diff
@@ -107,4 +109,55 @@ def contact
...
+    # TODO: we may know status primary
+    if work_form_contributor.person?
+      {
+        "name": [
+          {
+            "value": "#{work_form_contributor.last_name}, #{work_form_contributor.first_name}"
+          }
+        ],
+        "type": 'person',
+        # TODO: we will know code, uri, source code and source uri
```
ndushay, line 139: expecting this from Arcadia;  will do another PR when I have this info.

app/services/description_generator.rb
```diff
@@ -107,4 +109,55 @@ def contact
...
+    result
+  end
+
+  # in cocina model terms, returns a DescriptiveValue
+  sig do
+    params(work_form_contributor: Contributor)
+      .returns({ name: [{ value: T.untyped }], type: String, role: [{ value: T.untyped }] })
+  end
+  def contributor(work_form_contributor)
+    # TODO: we may know status primary
```
ndushay, line 130: Arcadia isn't yet sure if this will be addressed or not.

jcoyne (approved):

PR #396, opened 2020-11-09
Make the file property a virtual property

### LINKED ISSUE #388: Editing a previously saved draft cannot be saved again
1. Create a new object
2. Save it as a draft
3. Go back to edit the draft
4. Try and save the draft again - it does not appear to save.
5. Look at the javascript errors in the console

jcoyne:
## Why was this change made?
We don't want to try to sync this to an existing model.

Fixes #388

### DIFF
--- a/spec/requests/works_spec.rb
+++ b/spec/requests/works_spec.rb
@@ -311,6 +311,32 @@ RSpec.describe 'Works requests' do
       end
     end
 
+    describe 'update work with an attachment' do
+      let(:work) { create(:work, :with_attached_file) }
+      let(:user) { work.depositor }
+      let(:work_params) do
+        {
+          title: 'New title',
+          work_type: 'text',
+          contact_email: 'io@io.io',
+          abstract: 'test abstract',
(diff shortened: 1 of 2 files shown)


spec/requests/works_spec.rb
```diff
@@ -311,6 +311,32 @@
...
+          keywords_attributes: {
+            '0' => { '_destroy' => 'false', 'label' => 'Feminism', 'uri' => 'http://id.worldcat.org/fast/922671' }
+          },
+          license: 'CC0-1.0',
+          release: 'immediate'
+        }
+      end
+
+      it 'redirects to the work page' do
+        patch "/works/#{work.id}", params: { work: work_params }
```
mjgiarlo, line 335: It's not clear to me how this test tests the change above. What would the behavior of this test be w/o the change to a virtual attr?

jcoyne: It raises the error outlined in #388

mjgiarlo (approved):

PR #397, opened 2020-11-09
Adds work view.

### LINKED ISSUE #33: View a work
Update the view page to look like this as much as possible:

https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930436

justinlittman:
closes #33

## Why was this change made?
To allow viewing the details for a work.

Unit, locally.

settings.yml

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -35,4 +35,11 @@ class Work < ApplicationRecord
       transition deposited: :version_draft
     end
   end
+
+  sig { returns(T.nilable(String)) }
+  def purl
+    return nil unless druid
+
+    File.join(Settings.purl_url, T.must(druid).delete_prefix('druid:'))
+  end
 end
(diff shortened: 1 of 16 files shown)


justinlittman:
image

app/models/work.rb
```diff
@@ -35,4 +35,11 @@ class Work < ApplicationRecord
       transition deposited: :version_draft
     end
   end
+
+  sig { returns(T.nilable(String)) }
+  def purl
+    return nil unless druid
+
+    File.join(Settings.purl_url, T.must(druid).delete_prefix('druid:'))
```
jcoyne, line 43: Can you add a test for this?

jcoyne (changes requested):

app/models/work.rb
mjgiarlo: I'm curious about using `File.join` to do a string/URL join.

justinlittman: File.join is smart about the trailing slash. (We do the same in Argo: https://github.com/sul-dlss/argo/blob/8c31437630dcce1fdb8f9a9c6fb5f8ff5a8a9a32/app/components/external_links_component.rb#L10)

justinlittman: If we were running on Windows, this would be problematic but I'm thinking that is unlikely.

justinlittman: Test added.

jcoyne (approved):

app/models/work.rb
mjgiarlo: ok, thanks!

PR #398, opened 2020-11-09
Simplify the process of attaching a file 

jcoyne:
## Why was this change made?
The code is simpler

### DIFF
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -113,13 +113,14 @@ class WorkForm < Reform::Form
                end
                return item if item
 
-               new_attached_file = AttachedFile.new
-               new_attached_file.file.attach(ActiveStorage::Blob.find_signed(fragment['file']))
-               attached_files.append(new_attached_file)
+               attached_files.append(AttachedFile.new(file: fragment['file']))
              } do
     property :id
     property :label
     property :hide
+    # The file property is virtual so it doesn't get set on update, which causes:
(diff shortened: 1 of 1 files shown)


app/forms/work_form.rb
```diff
@@ -113,9 +113,7 @@ def sync(*)
                end
                return item if item
 
-               new_attached_file = AttachedFile.new
-               new_attached_file.file.attach(ActiveStorage::Blob.find_signed(fragment['file']))
-               attached_files.append(new_attached_file)
+               attached_files.append(AttachedFile.new)
```
mjgiarlo, line 116: This code was needed when I added it. If it's no longer needed, can you point to the change that obviated the need? i.e., how/where do we make the connection between the AF instance and the AS::Blob?

jcoyne: @mjgiarlo reform does this automatically when it syncs the `file` attribute from the form to the `file` attribute on AttacheFile.  The `file` attribute on attached file is generated via the has_one_attached method.  See https://edgeguides.rubyonrails.org/active_storage_overview.html#has-one-attached.

mjgiarlo: @jcoyne This doesn't seem to work for me. Pair?

mjgiarlo (approved):

mjgiarlo:
@jcoyne is planning to rebase this PR. do not merge.

PR #400, opened 2020-11-09
Tweak size of text field divs in collections form

mjgiarlo:
Can you comment on this change, @astridu?

I have a related question, as well. It looks like the space between the right edge of the text fields in the collection form is noticeably larger than it is in the work form. Compare e.g. [collection name](https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930437) and [title of deposit](https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420954309). Is this variance intentional? Thanks!

## Why was this change made?

This is done for greater fidelity to the mock-ups.

### BEFORE

before1

### AFTER

after

None

### DIFF
--- a/app/components/collections/form_component.html.erb
+++ b/app/components/collections/form_component.html.erb
@@ -5,14 +5,14 @@
     <header>Details</header>
     <div class="mb-3 row">
       <%= form.label :name, 'Collection name', class: 'col-sm-2 col-form-label' %>
-      <div class="col-sm-10">
+      <div class="col-sm-10 col-xl-8">
         <%= form.text_field :name, class: "form-control", required: true %>
         <div class="invalid-feedback">You must provide a name</div>
       </div>
     </div>
     <div class="mb-3 row">
       <%= form.label :description, 'Description', class: 'col-sm-2 col-form-label' %>
-      <div class="col-sm-10">
+      <div class="col-sm-10 col-xl-8">
         <%= form.text_field :description, class: "form-control", required: true %>
         <div class="invalid-feedback">You must provide a description</div>
       </div>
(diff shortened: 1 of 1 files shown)


astridu:
Much better. Thank you!!

app/components/collections/form_component.html.erb
```diff
@@ -5,22 +5,22 @@
     <header>Details</header>
     <div class="mb-3 row">
       <%= form.label :name, 'Collection name', class: 'col-sm-2 col-form-label' %>
-      <div class="col-sm-10">
+      <div class="col-sm-8">
```
jcoyne, line 8: Are you sure about `sm`?  That begins at 540px.  How about leaving `col-sm-10` and adding `col-xl-8`?

jcoyne (changes requested):

jcoyne:
The difference between the right and left is due to this style which pads 2rem all around and 4rem on the left side:

Screen Shot 2020-11-09 at 4 06 58 PM

app/components/collections/form_component.html.erb
mjgiarlo: @jcoyne I have no idea. I'm just tweaking what's already there. Re: what column sizing we already have in the app: there are currently 0 `col-xl-*`, 4 `col-lg-*`, and 55 'col-sm-*`. I'm happy to make changes but feel a bit like I'm flying blind.

mjgiarlo: @jcoyne happy to pair on this tomorrow.

mjgiarlo: @jcoyne done. take another look?

jcoyne (approved):

PR #408, opened 2020-11-10
DescriptionGenerator: map work#related_links to Cocina description

### LINKED ISSUE #341: Map related links to descriptive cocina

jmartin-sul:
## Why was this change made?

closes #341 

unit tests, rails console

n/a, hopefully specs cover it

### DIFF
--- a/app/services/description_generator.rb
+++ b/app/services/description_generator.rb
@@ -23,7 +23,8 @@ class DescriptionGenerator
       contributor: contributors,
       subject: keywords,
       note: [abstract, citation, contact],
-      event: [created_date, published_date].compact
+      event: [created_date, published_date].compact,
+      relatedResource: related_resource
     }
   end
 
@@ -157,6 +158,25 @@ class DescriptionGenerator
           }
         ]
       }
+    end
+  end
+
+  sig do
+    returns(T::Array[
+      T.any(
+        { type: String, access: { url: T::Array[{ value: String }] } },
+        { type: String, access: { url: T::Array[{ value: String }] }, title: T::Array[{ value: String }] }
+      )
+    ])
+  end
+  def related_resource
+    work.related_links.map do |rel_link|
+      {
+        type: 'related to',
+        access: { url: [{ value: rel_link.url }] }
+      }.tap do |h|
+        h[:title] = [{ value: rel_link.link_title }] if rel_link.link_title.present?
+      end
     end
   end
 end
--- a/spec/services/description_generator_spec.rb
+++ b/spec/services/description_generator_spec.rb
@@ -7,7 +7,8 @@ RSpec.describe DescriptionGenerator do
   subject(:model) { described_class.generate(work: work) }
 
   let(:work) do
-    build(:work, :with_creation_dates, :published, :with_keywords, :with_contributors)
+    build(:work, :with_creation_dates, :published, :with_keywords,
+          :with_contributors, :with_some_untitled_related_links)
   end
   let(:contrib1_name) { "#{work.contributors.first.last_name}, #{work.contributors.first.first_name}" }
   let(:contrib2_name) { "#{work.contributors[1].last_name}, #{work.contributors[1].first_name}" }
@@ -34,6 +35,26 @@ RSpec.describe DescriptionGenerator do
         { name: [{ value: contrib1_name }], role: [{ value: 'Contributing author' }], type: 'person' },
         { name: [{ value: contrib2_name }], role: [{ value: 'Contributing author' }], type: 'person' },
         { name: [{ value: contrib3_name }], role: [{ value: 'Contributing author' }], type: 'person' }
+      ],
+      relatedResource: [
+        {
+          type: 'related to',
+          title: [{ value: 'My Awesome Research' }],
+          access: { url: [{ value: 'http://my.awesome.research.io' }] }
+        },
+        {
+          type: 'related to',
+          title: [{ value: 'My Awesome Research' }],
+          access: { url: [{ value: 'http://my.awesome.research.io' }] }
+        },
+        {
+          type: 'related to',
+          access: { url: [{ value: 'https://your.awesome.research.ai' }] }
+        },
+        {
+          type: 'related to',
+          access: { url: [{ value: 'https://your.awesome.research.ai' }] }
+        }
       ]
     )
   end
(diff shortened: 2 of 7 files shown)


app/services/description_generator.rb
```diff
@@ -146,5 +147,18 @@ def contributor(work_form_contributor)
       ]
     }
   end
+
+  sig { returns(T::Array[T::Hash[String, T::Array[T::Hash[String, String]]]]) }
+  def related_resource
+    work.related_links.map do |rel_link|
+      uri_hash = {
+        "uri": rel_link.url
```
mjgiarlo, line 155: I think `"uri":` here should be `uri:` in line with the changes in https://github.com/sul-dlss/happy-heron/pull/407/files#diff-f21f5ecf9be9f56bd3c574ea7cb992626d29a9a41cc2926f9f643eadf3a15c55 Because these keys are not actually strings; they're symbols. Not sure why Sorbet is tricked by this!

jmartin-sul: ah, great!  yeah, i thought the `"string":` construction for hash keys was a bit odd, but i figured i'd stay consistent.   mean to ask but forgot, happy to change.

mjgiarlo: I wonder if some of this could move into a new `RelatedLink#to_h` method, but it's not a big deal. Also, we don't (yet?) do this in other models. This way, each model could take a little more responsibility for its own representation, and the description generator can handle the formatting external to the model? I could go either way on it. Would like to hear what you think, though.

spec/services/description_generator_spec.rb
```diff
@@ -34,6 +34,10 @@
         { name: [{ value: contrib1_name }], role: [{ value: 'Contributing author' }], type: 'person' },
         { name: [{ value: contrib2_name }], role: [{ value: 'Contributing author' }], type: 'person' },
         { name: [{ value: contrib3_name }], role: [{ value: 'Contributing author' }], type: 'person' }
+      ],
+      relatedResource: [
+        { identifier: [{ uri: 'http://my.awesome.research.io', displayLabel: 'My Awesome Research' }] },
```
mjgiarlo, line 39: It might be nice to have an example that shows a RelatedLink with a missing title. Worth making factory & spec tweaks for this?

app/services/description_generator.rb
```diff
@@ -146,5 +147,18 @@ def contributor(work_form_contributor)
       ]
     }
   end
+
+  sig { returns(T::Array[T::Hash[String, T::Array[T::Hash[String, String]]]]) }
+  def related_resource
+    work.related_links.map do |rel_link|
```
mjgiarlo, line 153: I wonder if some of this could move into a new RelatedLink#to_h method, but it's not a big deal. Also, we don't (yet?) do this in other models. This way, each model could take a little more responsibility for its own representation, and the description generator can handle the formatting external to the model? I could go either way on it. Would like to hear what you think, though.

spec/services/description_generator_spec.rb
jmartin-sul: good call, will do...

app/services/description_generator.rb
jmartin-sul: i like this idea, though i'd also worry slightly about the `#to_h` representation being very cocina specific... but maybe that's ok?  it does seem like the main use of a model's hash representation would be for translation to cocina representation.  

looking at this example:  https://github.com/sul-dlss-labs/cocina-descriptive-metadata/pull/115/files

i'm not sure that i'd expect such a cocina specific represenation from `#to_h`.

pref for starting that here in this PR, vs doing this one case in a follow on PR, vs updating all of the hash building in this class in one fell swoop in a separate PR?

could also have a `#to_cocina` method instead of using the `#to_h` method.

jmartin-sul:
turned to draft, will fix the output to match the example here after i get back from lunch: https://github.com/sul-dlss-labs/cocina-descriptive-metadata/pull/115/files

app/services/description_generator.rb
mjgiarlo: `#to_cocina` WFM too. Opinions, co-TL, @jcoyne ?

jmartin-sul:
ok, i:
* fixed the weird not-string keys in `#related_resource`
* updated the factory and spec to test for missing link title
* fixed the output to match the structure described in this example:  https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/h2_cocina_mappings/h2_to_cocina_related_resource.txt

i'll start a branch based on this branch to try out the `#to_h`/`#to_cocina` refactoring (happy to roll that work into this PR if i get that done before this is merged).

rebased on latest `main`, ready for another look, i think.

jmartin-sul:
d'oh, `sig` is now incorrect for the newly added method, turning to draft and fixing...

app/services/description_generator.rb
```diff
@@ -159,5 +160,17 @@ def contributor(work_form_contributor)
       }
     end
   end
+
+  sig { returns(T::Array[T::Hash[String, T::Array[T::Hash[String, String]]]]) }
```
ndushay, line 164: ew.  but we probably have this in other spots.   this is where maybe we could use the references to cocina types?

app/services/description_generator.rb
ndushay: i'd prefer `#to_cocina` or even `#to_cocina_hash` or whatever, but wondering if this refactoring should get priority over other work in ready column ...

ndushay (approved):

app/services/description_generator.rb
mjgiarlo: Like I said, not a big deal, so not a high priority.

mjgiarlo (approved):
LGTM, merge when undrafted.

app/services/description_generator.rb
jmartin-sul: yeah, this looks yucky.  i'm about to push something that's... a little longer, but much clearer about the expected structure (imho), so hopefully that'll be less yuck?

jmartin-sul:
@ndushay and @mjgiarlo here's the `sig` i ended up with, which i find clearer, though it is longer:
```ruby
sig do
  returns(T::Array[
    T.any(
      { type: String, access: { url: T::Array[{ value: String }] } },
      { type: String, access: { url: T::Array[{ value: String }] }, title: T::Array[{ value: String }] }
    )
  ])
end
```

PR #410, opened 2020-11-10
Send deposits in collection with reviewers to mediation

### LINKED ISSUE #349: Deposit for review
When I deposit an item in a collection that is configured for review (See #348), It should not immediately deposit to SDR and should be put into a "needs review" state.

jcoyne:
## Why was this change made?
Fixes #349

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -1,4 +1,4 @@
-# typed: false
+# typed: true
 # frozen_string_literal: true
 
 # The endpoint for CRUD about a Work
@@ -25,8 +25,7 @@ class WorksController < ApplicationController
 
     @form = WorkForm.new(work)
     if @form.validate(work_params) && @form.save
-      DepositJob.perform_later(work) if params[:commit] == 'Deposit'
-      redirect_to work
+      after_save(work)
     else
       # Send form errors to client in JSON format to be parsed and rendered there
       render json: @form.errors, status: :bad_request
@@ -47,8 +46,7 @@ class WorksController < ApplicationController
 
     @form = WorkForm.new(work)
     if @form.validate(work_params) && @form.save
-      DepositJob.perform_later(work) if params[:commit] == 'Deposit'
-      redirect_to work
+      after_save(work)
     else
       # Send form errors to client in JSON format to be parsed and rendered there
       render json: @form.errors, status: :bad_request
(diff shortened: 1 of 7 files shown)


app/controllers/works_controller.rb
```diff
@@ -61,22 +59,35 @@ def show
 
   private
 
+  sig { params(work: Work).void }
+  def after_save(work)
+    if params[:commit] == 'Deposit'
+      if work.collection.review_enabled?
+        work.submit_for_review!
```
ndushay, line 66: this is the new functionality?

app/forms/collection_form.rb
```diff
@@ -14,7 +14,7 @@ class CollectionForm < Reform::Form
   property :depositor_sunets, virtual: true, prepopulator: lambda { |_options|
     self.depositor_sunets = depositor_sunets_from_model.join(', ')
   }
-  property :review_enabled, virtual: true, default: -> { model.reviewers.present? ? 'true' : 'false' }
+  property :review_enabled, virtual: true, default: -> { model.review_enabled? ? 'true' : 'false' }
```
ndushay, line 17: and this change is also part of new functionality?

app/models/collection.rb
```diff
@@ -13,4 +13,9 @@ class Collection < ApplicationRecord
   validates :managers, presence: true
   validates :name, presence: true
   validates :access, presence: true
+
+  sig { returns(T::Boolean) }
+  def review_enabled?
+    reviewers.present?
+  end
```
ndushay, line 20: and this?

spec/requests/works_spec.rb
```diff
@@ -150,214 +150,258 @@
...
-          },
-          keywords_attributes: {
-            '0' => { '_destroy' => 'false', 'label' => 'Feminism', 'uri' => 'http://id.worldcat.org/fast/922671' }
-          },
-          license: 'CC0-1.0',
-          release: 'immediate'
-        }
-      end
+      context 'with a moderated collection' do
+        let(:collection) { create(:collection, :with_reviewers, depositors: [user]) }
```
ndushay, line 324: is it the "with_reviewers" that makes this test relevant to the code changes?

ndushay (commented):
this looks ok to me, but it's a bit hard to separate the functional code changes from sorbet only changes from refactoring changes.  Maybe in future those three things could be in separate commits, or you could call out in comments "this is part of the actual code change" for those sections?

It looks like we now check if review is enabled, and if so, submit the work for review?  and maybe you added a trait to the collection for a test so it is now magically testing this?

app/controllers/works_controller.rb
jcoyne: Yes.

app/forms/collection_form.rb
jcoyne: I created `#review_enabled?` to be used by new functionality, so I'm also using it here to avoid duplicate code.

app/models/collection.rb
jcoyne: Yep, this is used by the WorksController.

spec/requests/works_spec.rb
jcoyne: Yes.

mjgiarlo (approved):

mjgiarlo:
@jcoyne can you respond to @ndushay's question:

> It looks like we now check if review is enabled, and if so, submit the work for review? and maybe you added a trait to the collection for a test so it is now magically testing this?

I think that's the last thing to resolve before we can merge.

PR #412, opened 2020-11-10
Use the structs we have rather than defining complex interfaces

jcoyne:
Unfortunately sorbet doesn't understand dry-struct signatures, so I had to manually define a partial interface to Cocina::Models.  We might want to look at moving cocina-models from dry-struct to https://sorbet.org/docs/tstruct

## Why was this change made?

Discussion  in infrastructure meeting today.

### DIFF
--- a/app/services/description_generator.rb
+++ b/app/services/description_generator.rb
@@ -6,7 +6,7 @@
 class DescriptionGenerator
   extend T::Sig
 
-  sig { params(work: Work).returns(T::Hash[Symbol, T::Array[T::Hash[String, T.untyped]]]) }
+  sig { params(work: Work).returns(Cocina::Models::Description) }
   def self.generate(work:)
     new(work: work).generate
   end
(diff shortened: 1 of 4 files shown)


mjgiarlo:
@jcoyne :speech_balloon: 
> ... We might want to look at moving cocina-models from dry-struct to [sorbet.org/docs/tstruct](https://sorbet.org/docs/tstruct)

:sweat_smile:

mjgiarlo:
# Note to reviewers

You should manually expand `sorbet/rbi/custom/cocina-models.rbi` in the files tab, which will collapse by default. 

If we are going to go forward with this approach, maybe we should add an ignore for `sorbet/rbi/custom/**/* linguist-generated=false` to https://github.com/sul-dlss/happy-heron/blob/main/.gitattributes?

mjgiarlo:
@jcoyne I  notice in your recent review-related PR, you added custom definitions to `sorbet/custom/` and this PR, you added them to `sorbet/rbi/custom/`. Is that an intentional difference? Wondering when to use one over the other. thanks!

ndushay (commented):
I like this direction very much.  

Any idea if tstruct will be more compatible with ruby 3 type checking?

whee!   For every thing, churn, churn, churn ...

mjgiarlo:
@ndushay :speech_balloon: 
> 
> Any idea if tstruct will be more compatible with ruby 3 type checking?

I don't think we know. https://sorbet.org/docs/faq#when-ruby-3-gets-types-what-will-the-migration-plan-look-like is the official answer. Apparently, also, from [this blog post](https://developer.squareup.com/blog/the-state-of-ruby-3-typing/):

> To achieve interoperability, RBS gem ships with a translator from RBI to RBS, and a translator from RBS to RBI is being developed.
 
So it seems like we'll have one or more migration paths, but it's hard to say right now 

> whee! For every thing, churn, churn, churn ...

<iframe src="https://giphy.com/embed/xp4VN7OJ5hR1UadiaM" width="480" height="270" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/kaceymusgraves-biscuits-kacey-musgraves-xp4VN7OJ5hR1UadiaM">via GIPHY</a></p>

mjgiarlo (changes requested):
Can you move the custom cocina-models RBI from sorbet/rbi/custom/ to sorbet/custom/ pls? Thank you for doing this!

mjgiarlo (approved):

PR #413, opened 2020-11-10
Revert "Defer to contributor to have knowledge of its name and type"

mjgiarlo:
Reverts sul-dlss/happy-heron#407


ndushay:
this PR somehow broke the contributor mappings in descriptive_generator so reverting for now.

PR #414, opened 2020-11-11
Allow saving of draft works and collections without validations

### LINKED ISSUE #314: Saving item draft does not require the Required fields
Currently, the deposit form will not let you save a draft if the required fields are not filled out. 

The desired behavior is that the form validations do not affect the ability to save a draft. The required fields are only required for depositing.

peetucket:
## Why was this change made?

Fixes #314 and #437 

Skip validations (client and server side) when saving a work as a draft.

- tweaks to javascript to skip client side validation if the user clicks the save draft button
- move the `WorkForm` to a new `DraftWorkForm`, but remove (most) validations (and ditto for the `CollectionForm`)
- change the `WorkForm` to subclass from the new `DraftWorkForm`, with the rest of the validations (and ditto for the `CollectionForm`)
- update the `Work` and `Collection` models to remove most validations (to allow the models to be saved when in draft)
- update the `Work` and `Collection` controllers to use the draft form classes when the save as draft button is pressed
- some refactoring of the collection spec to make the params style match what is in the work spec

Note: it looks like there are a lot of changes to the `WorkForm` and the `CollectionForm` - there are not.  It is just moving all that code to the new `DraftXXX` classes, except for the validations.  We then subclass from the new `DraftXXX` classes form to get the deposit classes.  Note that some validation that we need in both cases remains in the regular form classes.

Updated and new unit tests, in localhost browser


app/controllers/works_controller.rb
```diff
@@ -96,4 +111,12 @@ def validate_work_types!
     flash[:error] = errors.join("\n")
     redirect_to dashboard_path
   end
+
+  def submit_type
```
jcoyne, line 115: consider changing this to a boolean. e.g.: `deposit?`

app/javascript/modules/validate-forms.js
```diff
@@ -9,6 +9,7 @@
     Array.prototype.slice.call(forms)
       .forEach(function (form) {
         form.addEventListener('submit', function (event) {
+          if (event.submitter.id === 'save-draft-button') return // do not run client side validation for saving drafts
```
jcoyne, line 12: Can this be moved to `app/javascript/controllers/edit_deposit_controller.js`?  This current file applies to all forms, but this functionality is only for a particular form.

spec/requests/works_spec.rb
```diff
@@ -265,7 +265,7 @@
       end
 
       it 'displays the work' do
-        post "/collections/#{collection.id}/works", params: { work: work_params }
+        post "/collections/#{collection.id}/works", params: { work: work_params, commit: 'Deposit' }
```
jcoyne, line 268: making all of these have "Deposit", means we're no longer testing the draft branch.

app/controllers/works_controller.rb
peetucket: yup, makes sense

app/javascript/modules/validate-forms.js
peetucket: i was concerned about doing this globally too, which is why we try to do it this way (it works as expected except for buttons with an id of "save-draft-button").  not sure how we would extract just this scenario without removing the global event listener?

spec/requests/works_spec.rb
peetucket: we definitely will need a new test or two to verify the ability to save a draft without validation (which currently doesn't work in this draft PR by the way, as we have validation at the model level as well, preventing the save)...

were some of these tests actually exercising the draft button before?

app/models/work.rb
```diff
@@ -14,12 +14,8 @@ class Work < ApplicationRecord
   has_many :attached_files, dependent: :destroy
   has_many :keywords, dependent: :destroy
 
-  validates :abstract, :access, :state, :title, presence: true
-  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
+  validates :state, presence: true
```
jcoyne, line 17: I think we still want the work_type and work_subtype validators here.  I think we also want all the format validators here (just not the presence validations)

peetucket: ok

app/javascript/modules/validate-forms.js
peetucket: Note: this is now used as well for the collection form too to bypass client side validation for collections saved as drafts.

spec/requests/collections_spec.rb
```diff
@@ -235,19 +229,38 @@
...
+          {
+            name: '',
+            description: '',
+            contact_email: '',
+            access: 'world',
+            depositor_sunets: ''
+          }
+        end
+
+        xit 'creates a new draft collection' do
```
peetucket, line 255: this test should work based on it working in the browser, but doesn't...not sure why

spec/requests/works_spec.rb
peetucket: Added a new test specifically to show that you can save a draft without the normal validations.  Switched all other tests to explicitly submit for both works and collections.

mjgiarlo (commented):
I wonder if you would consider renaming the forms. "WorkFormDraft" suggest to me that the form is a draft, but the form is intended to be used for draft works. Would you consider `{STATE}{TYPE}Form`? Like, `DraftWorkForm` or `DepositCollectionForm`?

peetucket:
> I wonder if you would consider renaming the forms. "WorkFormDraft" suggest to me that the form is a draft, but the form is intended to be used for draft works. Would you consider `{STATE}{TYPE}Form`? Like, `DraftWorkForm` or `DepositCollectionForm`?

Sounds good.

peetucket:
Note: just did some refactoring of the collection and work controllers to select the form (draft vs regular) in a single method (to avoid code duplication)

app/controllers/application_controller.rb
```diff
@@ -22,6 +22,11 @@ def user_with_groups
 
   private
 
+  sig { returns(T::Boolean) }
+  def deposit?
+    params[:commit] == 'Deposit'
```
jcoyne, line 27: I don't think this belongs in the Application controller.  Please move to a mixin or a subclass.

app/models/collection.rb
```diff
@@ -8,11 +8,7 @@ class Collection < ApplicationRecord
   has_and_belongs_to_many :depositors, class_name: 'User', join_table: 'depositors'
   has_and_belongs_to_many :reviewers, class_name: 'User', join_table: 'reviewers'
 
-  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
-  validates :description, presence: true
-  validates :managers, presence: true
-  validates :name, presence: true
-  validates :access, presence: true
+  validates :contact_email, format: { with: Devise.email_regexp }, unless: -> { contact_email.blank? }
```
jcoyne, line 11: Is the unless necessary?

app/models/work.rb
```diff
@@ -15,10 +15,9 @@ class Work < ApplicationRecord
   has_many :keywords, dependent: :destroy
   has_many :events, dependent: :destroy
 
-  validates :abstract, :access, :state, :title, presence: true
-  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
+  validates :contact_email, format: { with: Devise.email_regexp }, unless: -> { contact_email.blank? }
```
jcoyne, line 18: Is unless required?

jcoyne (changes requested):

app/models/collection.rb
peetucket: This is needed if we want to allow a draft collection or work to be saved with a blank email, but also want to verify that the format of the email is valid if is provided.

jcoyne: doesn't the format only run if it's not nil?

peetucket: Not sure about nil, but it is an empty string when submitted on the form (and left blank by the user), which seems to run the format check.

jcoyne: Looks like you're right.  Can you use `allow_blank: true` here instead? That seems to be more clear.

peetucket: good idea, that also lets us get back to the strict type check (#437)

app/controllers/application_controller.rb
peetucket: Moved to a new controller that I subclass Collections and Works from to keep it separate from other controllers.  If we use a module, we'd have to pass in the params which is not as clean reading in the code.

app/models/work.rb
```diff
@@ -15,10 +15,9 @@ class Work < ApplicationRecord
   has_many :keywords, dependent: :destroy
   has_many :events, dependent: :destroy
 
-  validates :abstract, :access, :state, :title, presence: true
-  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
+  validates :contact_email, format: { with: Devise.email_regexp }, allow_blank: true
+  validates :state, presence: true
   validates :created_edtf, :published_edtf, edtf: true
-  validates :license, presence: true, inclusion: { in: License.license_list }
```
jcoyne, line 21: We still ought to validate that license is in this set.

jcoyne (changes requested):

app/models/work.rb
peetucket: ok, put back on the model and added some tests to the model and the form for the license

app/views/dashboards/show.html.erb
```diff
@@ -49,7 +49,7 @@
           <% @presenter.collections.each do |collection| %>
             <li>
               <div>
-                <header><%= collection.name %>
+                <header><%= collection.name.presence || 'No Title' %>
```
peetucket, line 52: Now that we allow draft collections, they might not have a title (similar to the works, which do something like this too for display purposes on the dashboard).

jcoyne (approved):

PR #418, opened 2020-11-11
Type alias antics

jmartin-sul:
leaving this as a draft based on the unmerged branch it refactors, for discussion.

## Why was this change made?

As an experiment with refactoring and type aliases, to see whether people like any of the changes here.  Could be a useful pattern for dealing with complex return types?

I started on the refactoring suggested by @mjgiarlo in a review comment on another PR (https://github.com/sul-dlss/happy-heron/pull/408#discussion_r520778696), because it seemed a quick enough thing to try while that part of the code was already in my working memory.

Pushing the cocina hash building back into the relevant model in the first commit made me want to define reusable types to cut down on repetition.

Then in the second commit I sort of leaned on type aliases to build up somewhat more exacting return signatures that (I found) readable.

Off the top of my head, the pros and cons of the second commit seem to be...
* pros
  * less of a mess of nested array and hash types everywhere
  * reusable/composable parts let you define canonical types in one place and put them together elsewhere
  * lets you do validation of complex shapes built up by helper methods, without so much repetitive type declaration
  * lets you easily see when two methods which have complex return shapes actually return the same thing
* cons
  * more indirection/constant definition
  * may have to page or tab around or look at multiple sigs to figure out the full shape of a type composed from type aliases
  * i have no idea whether this type aliasing construct will need re-working once ruby supports type info natively

unit tests

arguably makes the code more self-documenting?

### DIFF
--- a/app/models/related_link.rb
+++ b/app/models/related_link.rb
@@ -3,7 +3,25 @@
 
 # Models a URI that is related to a work
 class RelatedLink < ApplicationRecord
+  extend T::Sig
+
+  COCINA_HASH_TYPE_NO_TITLE = T.type_alias { { type: String, access: { url: T::Array[{ value: String }] } } }
+  COCINA_HASH_TYPE_WITH_TITLE = T.type_alias do
+    { type: String, access: { url: T::Array[{ value: String }] }, title: T::Array[{ value: String }] }
+  end
+  COCINA_HASH_TYPE = T.type_alias { T.any(COCINA_HASH_TYPE_NO_TITLE, COCINA_HASH_TYPE_WITH_TITLE) }
+
   belongs_to :work
 
   validates :url, presence: true
+
+  sig { returns(COCINA_HASH_TYPE) }
+  def to_cocina_hash
+    {
+      type: 'related to',
+      access: { url: [{ value: url }] }
+    }.tap do |h|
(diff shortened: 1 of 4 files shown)


jmartin-sul:
ha, i hadn't seen #412 when i did this.  that approach would probably obviate the need for this PR, and at first glance i like that one more than this one, but will leave this open till #412 is discussed by the team.

app/models/related_link.rb
```diff
@@ -3,7 +3,25 @@
 
 # Models a URI that is related to a work
 class RelatedLink < ApplicationRecord
+  extend T::Sig
+
+  COCINA_HASH_TYPE_NO_TITLE = T.type_alias { { type: String, access: { url: T::Array[{ value: String }] } } }
```
jcoyne, line 8: How about naming these after the Cocina types that they represent?  e.g DescriptiveValue, Title, Event, etc.

ndushay (commented):
I much prefer JCoyne's approach in PR #412.   I would be more inclined to use vaguer types than declare the ones here ... but I suspect we'll head more towards 412.

jmartin-sul:
closing in favor of #412 based on discussion from this morning

mjgiarlo:
@jmartin-sul thank you for spiking on this and helping us learn about sorbet type aliases!

PR #421, opened 2020-11-11
Unauthenticated users should not able to view works

jcoyne:
## Why was this change made?

We don't have a requirement to let in unauthenticated users.


mjgiarlo (changes requested):
Hold. Checking with the P.O.

mjgiarlo (approved):

PR #447, opened 2020-11-13
contributor mappings improvements

ndushay:
part of #240 (map contributors into cocina model)

## Why was this change made?

This PR is mostly focused on getting the h2 contributor details mapped to the cocina model.  My two reference docs from Arcadia:
- https://github.com/sul-dlss-labs/cocina-descriptive-metadata/blob/master/h2_cocina_mappings/h2_to_cocina_contributor.txt#L717
- https://docs.google.com/spreadsheets/d/1CvEd_NODprNhM2D9VfvJBFs1jfAMEUr0kDxXHe2HkL4

This is maybe 85% of the work;  I thought it best to get this up in a PR while I work on the rest, which may be tricky.

The beginnings of these mappings were in the description_generator;  i move those to the new contributors_generator class.

I wrote unit tests, mostly corresponding to Arcadia's mappings, but in 3 cases I haven't yet implemented her specs. I also added a test or two of my own, including one for the full descriptive metadata.

my questions peppered Arcadia as I worked;  the two documents above are excellent.

### DIFF
--- a/app/services/contributors_generator.rb
+++ b/app/services/contributors_generator.rb
@@ -1,0 +1,243 @@
+# typed: true
+# frozen_string_literal: true
+
+# generates Cocina::Models::Contributors to be used by DescriptionGenerator (ultimately in a Cocina::Models::RequestDRO)
+# rubocop:disable Metrics/ClassLength
+class ContributorsGenerator
+  extend T::Sig
+
+  sig { params(work: Work).returns(T::Array[Cocina::Models::Contributor]) }
+  def self.generate(work:)
+    new(work: work).generate
+  end
+
+  sig { params(work: Work).returns(T::Array[Cocina::Models::DescriptiveValue]) }
+  def self.form_from_contributors(work:)
+    new(work: work).form_from_contributors
+  end
+
+  sig { params(work: Work).void }
+  def initialize(work:)
+    @work = work
+  end
+
+  sig { returns(T::Array[Cocina::Models::Contributor]) }
+  def generate
+    result = []
+
+    work.contributors.each do |work_form_contributor|
+      result << contributor(work_form_contributor)
+    end
+
+    result
+  end
+
+  FORM_EVENT = Cocina::Models::DescriptiveValue.new(
+    {
+      value: 'Event',
+      type: 'resource types',
+      source: { value: 'DataCite resource types' }
+    }
+  )
+
+  sig { returns(T::Array[Cocina::Models::DescriptiveValue]) }
+  def form_from_contributors
+    work.contributors.each do |contributor|
+      next unless contributor.role == 'Conference' || contributor.role == 'Event'
+
+      return [FORM_EVENT]
+    end
+
+    []
+  end
+
+  private
+
+  sig { returns(Work) }
+  attr_reader :work
+
+  sig { params(contributor: Contributor).returns(T.nilable(Cocina::Models::Contributor)) }
+  def contributor(contributor)
+    # FIXME: TODO: mappings for status (primary) and/or order
+    Cocina::Models::Contributor.new(
+      "name": name_descriptive_value(contributor),
+      "type": contributor_type(contributor),
+      "role": cocina_roles(contributor.role)
+    )
+  end
+
+  sig { params(contributor: Contributor).returns(T::Array[Cocina::Models::DescriptiveValue]) }
+  def name_descriptive_value(contributor)
+    return [Cocina::Models::DescriptiveValue.new(value: full_name(contributor))] unless contributor.person?
+
+    [Cocina::Models::DescriptiveValue.new(value: full_name(contributor), type: 'inverted full name')]
+  end
+
+  sig { params(contributor: Contributor).returns(T.nilable(String)) }
+  def full_name(contributor)
+    return "#{contributor.last_name}, #{contributor.first_name}" if contributor.person?
+
+    contributor.full_name
+  end
+
+  sig { params(contributor: Contributor).returns(String) }
+  def contributor_type(contributor)
+    return 'conference' if contributor.role == 'Conference'
+    return 'event' if contributor.role == 'Event'
+
+    contributor.contributor_type
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  sig { params(role: String).returns(T::Array[Cocina::Models::DescriptiveValue]) }
+  def cocina_roles(role)
+    result = [
+      Cocina::Models::DescriptiveValue.new(
+        value: role,
+        source: {
+          value: 'Stanford self-deposit contributor types'
+        }
+      )
+    ]
+    result << T.must(marcrelator_role(role)) if marcrelator_role(role)
+    result << T.must(datacite_role(role)) if datacite_role(role)
+    result
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  # rubocop:disable Metrics/MethodLength
+  sig { params(role: String).returns(T.nilable(Cocina::Models::DescriptiveValue)) }
+  def marcrelator_role(role)
+    mr_code = ROLE_TO_MARC_RELATOR_CODE[role]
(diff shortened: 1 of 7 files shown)


app/services/description_generator.rb
```diff
@@ -20,11 +20,12 @@ def initialize(work:)
...
     Cocina::Models::Description.new({
                                       title: title,
-                                      contributor: contributors,
+                                      contributor: ContributorsGenerator.generate(work: work),
                                       subject: keywords,
                                       note: [abstract, citation, contact],
                                       event: [created_date, published_date].compact,
-                                      relatedResource: related_links + related_works
+                                      relatedResource: related_links + related_works,
+                                      form: ContributorsGenerator.form_from_contributors(work: work)
```
ndushay, line 28: yes, the contributor information from h2 can trigger `form` elements in the cocina model.  yay!

app/services/description_generator.rb
```diff
@@ -108,50 +109,6 @@ def contact
...
-        # TODO: we will know code, uri, source code and source uri
-        role: [
-          {
-            value: work_form_contributor.role
-          }
-        ]
-      )
-    end
-  end
-
```
ndushay, line 154: all this moved to the new contributors_generator class

sorbet/custom/cocina-models.rbi
```diff
@@ -16,8 +16,15 @@ module Cocina::Models
...
-    def initialize(value:, type: nil, displayLabel: nil); end
+    sig do
+      params(value: T.nilable(String),
+             code: T.nilable(String),
+             uri: T.nilable(String),
+             type: T.nilable(String),
+             source: T.nilable(Object),
+             displayLabel: String).void
+    end
+    def initialize(value: nil, code: nil, uri: nil, type: nil, source: nil, displayLabel: nil); end
```
ndushay, line 27: this work needed more out of our DescriptiveValue type

spec/services/description_generator_spec.rb
```diff
@@ -6,13 +6,12 @@
 RSpec.describe DescriptionGenerator do
   subject(:model) { described_class.generate(work: work).to_h }
 
+  let(:contributor) { build(:contributor, :with_org_contributor) }
```
ndushay, line 9: this let me simplify this spec;  the complex contributor mapping tests are now in the contributors_generator_spec class

spec/factories/works.rb
```diff
@@ -43,10 +43,6 @@
     contributors { Array.new(contributors_count) { association(:contributor) } }
   end
 
-  trait :with_mixed_contributors do
-    contributors { [association(:contributor), association(:contributor, :with_org_contributor)] }
-  end
-
```
ndushay, line 49: this is no longer used (or needed)

app/services/contributors_generator.rb
```diff
@@ -0,0 +1,243 @@
...
+  sig { returns(T::Array[Cocina::Models::DescriptiveValue]) }
+  def form_from_contributors
+    work.contributors.each do |contributor|
+      next unless contributor.role == 'Conference' || contributor.role == 'Event'
+
+      return [FORM_EVENT]
+    end
+
+    []
+  end
```
mjgiarlo, line 52: Let me see if I'm reading this right: if *any* contributor has a role of conference or event, we return a single descriptive value? That's true if there's one conf/event contrib or three or seven, and there's no information from the contributor that is returned? If that's true, I wonder if the `#each` here could be jettisoned in favor of something like:

```ruby
return [FORM_EVENT] if work.contributors.pluck(:role).any? { |role| ['Event', 'Conference'].include?(role) }

[]
```

app/services/contributors_generator.rb
```diff
@@ -0,0 +1,243 @@
...
+
+    contributor.full_name
+  end
+
+  sig { params(contributor: Contributor).returns(String) }
+  def contributor_type(contributor)
+    return 'conference' if contributor.role == 'Conference'
+    return 'event' if contributor.role == 'Event'
+
+    contributor.contributor_type
```
mjgiarlo, line 88: So this method returns one of four values:

1. person
2. organization
3. conference
4. event

:question:

app/services/contributors_generator.rb
```diff
@@ -0,0 +1,243 @@
...
+    result << T.must(datacite_role(role)) if datacite_role(role)
+    result
+  end
+  # rubocop:enable Metrics/MethodLength
+
+  # rubocop:disable Metrics/MethodLength
+  sig { params(role: String).returns(T.nilable(Cocina::Models::DescriptiveValue)) }
+  def marcrelator_role(role)
+    mr_code = ROLE_TO_MARC_RELATOR_CODE[role]
+    mr_value = MARC_RELATOR_CODE_TO_VALUE[mr_code]
```
mjgiarlo, line 112: I was a little surprised to find these constants down below. I prefer having constants defined together near the top of a class, but I'd also be OK with them located just above the method where they are used since I know that tends to be your preference.

app/services/contributors_generator.rb
```diff
@@ -0,0 +1,243 @@
...
+    return unless mr_code || mr_value
```
mjgiarlo, line 113: For some reason, `unless` combined with `||` always twists my brain, :laughing: (no changes needed).

app/services/contributors_generator.rb
```diff
@@ -0,0 +1,243 @@
...
+    datacite_role = ROLE_TO_DATA_CITE_VALUE[role]
+    return if datacite_role.blank?
+
+    case datacite_role
+    when String
+      datacite_source = 'DataCite contributor types'
+    when :creator
+      datacite_role = datacite_role.to_s.capitalize
+      datacite_source = 'DataCite properties'
+    end
```
mjgiarlo, line 139: This could be extracted to another method to make this one shorter, since it seems to have two responsibilities: looking up a role string to derive role and source values, and then injecting those values into a new DV. But I don't think this method is hard to grok, so I'm fine with it staying as it with the inline exclusions. :+1:

app/services/description_generator.rb
mjgiarlo: :1st_place_medal:

app/services/description_generator.rb
mjgiarlo: love it

sorbet/custom/cocina-models.rbi
mjgiarlo: well done!

spec/factories/works.rb
mjgiarlo: :fire:

spec/services/description_generator_spec.rb
mjgiarlo: :man_dancing:

mjgiarlo (commented):
Looking good! Well tested PR. Have a couple questions and some niggles about iterators, nothing too big I don't think.

app/services/contributors_generator.rb
ndushay: so far, yes.

app/services/contributors_generator.rb
ndushay: yes ... but is that actually clearer?  Or just more clever?

and correct - only one such value regardless of how many contributors.

app/services/contributors_generator.rb
ndushay: agree - could do `if !mr_code && !mr_value` but that's no better here.

app/services/contributors_generator.rb
ndushay: yeah, i thought about it but deriving two values in a single method starts to need a data structure and that gets harder to grok.

mjgiarlo (approved):
LGTM. Thanks!

app/services/contributors_generator.rb
ndushay: this works, but the above did not:

```ruby
work.contributors.select { |c| ['Event', 'Conference'].include?(c.role) }.empty?
```

PR #449, opened 2020-11-13
[Refactor] contributor_generator small refactors for better clarity

ndushay:
## Why was this change made?

Some optional refactors were suggested for PR #447;  I experimented and ended up with these.

by the tests I wrote for PR #447, of course!


mjgiarlo (approved):
Thank you!

PR #456, opened 2020-11-16
Update dependencies

sul-devops-team:


.circleci/config.yml
```diff
@@ -105,6 +105,9 @@ jobs:
       - node/install-packages:
           cache-key: yarn.lock
           pkg-manager: yarn
+      - run:
```
jcoyne, line 108: Is this necessary?

jcoyne (changes requested):

.circleci/config.yml
jmartin-sul: nope, it was a bit of flailing to reproduce a local build failure in CI to get another data point.  this is a pretty churny PR as-is, and i was planning to clean it up in a bit, but broke for dinner 🙂  (will put in draft in the meantime)

jmartin-sul: removed

jcoyne (approved):
Thank you.

PR #458, opened 2020-11-16
Properly serialize EDTF dates

### LINKED ISSUE #457: MM/YYYY Dates are not persisting as edtf
Enter 2020 - 11 as a publication date and save. On the show page we should see "2020-11", but currently it adds a day "01":

```
ww = Work.find 1234
ww.update(created_edtf: EDTF.parse('2020-11'))
ww.reload
ww.created_edtf
=> "2020-11-01"
```

jcoyne:
## Why was this change made?

Fixes #457

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -51,4 +51,15 @@ class Work < ApplicationRecord
 
     File.join(Settings.purl_url, T.must(druid).delete_prefix('druid:'))
   end
+
+  sig { params(edtf: T.nilable(T.any(EDTF::Interval, Date))).void }
+  # Ensure that EDTF dates get an EDTF serialization
+  def created_edtf=(edtf)
+    case edtf
+    when nil, EDTF::Interval
+      super
+    when Date
+      super(edtf.to_edtf)
+    end
+  end
 end
--- a/spec/factories/works.rb
+++ b/spec/factories/works.rb
@@ -16,7 +16,7 @@ FactoryBot.define do
   end
 
   trait :published do
-    published_edtf { '2020-02-14' }
+    published_edtf { EDTF.parse('2020-02-14') }
   end
 
   trait :pending_approval do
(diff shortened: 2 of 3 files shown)


spec/models/work_spec.rb
```diff
@@ -26,32 +26,60 @@
     expect(work.attached_files).to be_present
   end
 
-  it 'has a valid contact email' do
-    work.contact_email = 'notavalidemail'
-    expect { work.save! }.to raise_error(ActiveRecord::RecordInvalid, /Contact email is invalid/)
-  end
+  describe 'contact_email' do
```
justinlittman, line 29: Does this belong in this PR?

spec/factories/works.rb
```diff
@@ -16,15 +16,15 @@
   end
 
   trait :published do
-    published_edtf { '2020-02-14' }
+    published_edtf { EDTF.parse('2020-02-14') }
```
mjgiarlo, line 19: Can you say more about why changes to this file are needed? I thought these strings were crafted to be EDTF-compliant.

app/models/work.rb
```diff
@@ -51,4 +51,15 @@ def purl
 
     File.join(Settings.purl_url, T.must(druid).delete_prefix('druid:'))
   end
+
+  sig { params(edtf: T.nilable(T.any(EDTF::Interval, Date))).void }
```
mjgiarlo, line 55: Was this ever working properly? Or did it fall through the cracks when we switched over to form-based validation?

spec/models/work_spec.rb
jcoyne: It was changed to be consistent with how it was being done in this PR.

spec/factories/works.rb
jcoyne: Because published_edtf= should always receive a date/interval object, never a string.

app/models/work.rb
jcoyne: It's because the EDTF gem uses monkeypatching on the Date object rather than creating it's own object types. 😢

mjgiarlo: Oh boy.

spec/factories/works.rb
mjgiarlo: OK

mjgiarlo (approved):

PR #459, opened 2020-11-16
Validate that publication date is in the past

### LINKED ISSUE #329: Add validation to publication date to make sure it is in the past.
While the year validation seems to be working, I can pick a month/day that are in the future.

related to #373

jcoyne:
## Why was this change made?
Fixes #329

### DIFF
--- a/app/validators/created_in_past_validator.rb
+++ b/app/validators/created_in_past_validator.rb
@@ -20,6 +20,6 @@ class CreatedInPastValidator < ActiveModel::EachValidator
   def validate_date(record, attribute, value, prefix = nil)
     prefix &&= "#{prefix} "
     record.errors.add(attribute, "#{prefix}must have a four digit year") if Settings.earliest_year > value.year
-    record.errors.add(attribute, "#{prefix}must be in the past") if Time.zone.today.year < value.year
+    record.errors.add(attribute, "#{prefix}must be in the past") if Time.zone.today < value
   end
 end
--- a/spec/validators/created_in_past_validator_spec.rb
+++ b/spec/validators/created_in_past_validator_spec.rb
@@ -29,13 +29,41 @@ RSpec.describe CreatedInPastValidator do
       end
     end
 
-    context 'with a date after the current year' do
+    describe 'validate date is in the past' do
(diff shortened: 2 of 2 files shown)


spec/validators/created_in_past_validator_spec.rb
```diff
@@ -29,13 +29,41 @@
...
+      let(:today) { Time.zone.today }
 
-      it 'has errors' do
-        expect(record.errors.full_messages).to eq ['Created edtf must be in the past']
+      context 'with a date after the current year' do
+        let(:year) { today.year + 1 }
+        let(:value) { EDTF.parse(year.to_s) }
+
+        it 'has errors' do
+          expect(record.errors.full_messages).to eq ['Created edtf must be in the past']
```
justinlittman, line 41: Is this a useful error message?

app/validators/created_in_past_validator.rb
```diff
@@ -20,6 +20,6 @@ def validate_each(record, attribute, value)
   def validate_date(record, attribute, value, prefix = nil)
     prefix &&= "#{prefix} "
     record.errors.add(attribute, "#{prefix}must have a four digit year") if Settings.earliest_year > value.year
-    record.errors.add(attribute, "#{prefix}must be in the past") if Time.zone.today.year < value.year
+    record.errors.add(attribute, "#{prefix}must be in the past") if Time.zone.today < value
```
mjgiarlo, line 23: What is `value` here? Is it a string, a date, an edtf, etc.?

spec/validators/created_in_past_validator_spec.rb
mjgiarlo: For a developer, sure. For an end-user, not so much.

jcoyne: The end user will never look at `full_messages`, so no problem.

app/validators/created_in_past_validator.rb
jcoyne: It is a Date

mjgiarlo (approved):

Issue #392, opened 2020-11-09
Draft item edit page does not seem show a previously uploaded file


peetucket:
1. Create a new item
2. Upload a file
3. Save it as a draft
4. Return to edit the draft
5. Previously uploaded file does not appear to show correctly.

![Screen Shot 2020-11-09 at 11.23.46 AM.png](https://images.zenhubusercontent.com/59b9a35fb0222d5de477a1fb/f736267e-cabb-4038-8e4f-32a74c5fc41d)

jcoyne:
@astridu I'm not sure we have a good design for what a file is supposed to look like if it was previously uploaded, then saved as a draft and then revisited.  Does it even make sense to have a progress bar?

astridu:
in that scenario, could we display the file name with a green check to show that it was added previously? 
![image](https://user-images.githubusercontent.com/42216982/98599860-87454c00-2291-11eb-8845-86a1ebbcbdc5.png)

jcoyne:
@astridu Yes, we can do that. Thank you.

PR #474, opened 2020-11-16
Display already uploaded files in work form

### LINKED ISSUE #440: Improve positioning of "Choose files" widget in work form
The "Choose files" widget should render above uploaded files. I was able to reproduce this by coming back to the edit view after depositing a work with some files.

Screenshot from 2020-11-12 14-08-18

This is how it ought to look:

Screenshot from 2020-11-12 14-11-55

mjgiarlo:
## Why was this change made?

Fixes #440, #392 

See commit messages.

CI and in browser locally.

### BEFORE

Screenshot from 2020-11-16 14-40-21

### AFTER
Screenshot from 2020-11-16 14-40-43

None


spec/components/works/add_files_component_spec.rb
```diff
@@ -3,12 +3,35 @@
...
+# rubocop:disable RSpec/AnyInstance
 RSpec.describe Works::AddFilesComponent do
   let(:form) { ActionView::Helpers::FormBuilder.new(nil, nil, controller.view_context, {}) }
   let(:rendered) { render_inline(described_class.new(form: form)) }
 
-  it 'renders the component' do
-    expect(rendered.to_html).to include('Add your files')
-    expect(rendered.css('button.dz-clickable').to_html).to include('Choose files')
+  before do
+    allow_any_instance_of(ActionView::Helpers::FormBuilder).to receive(:object).and_return(object)
```
jcoyne, line 13: why don't you pass the object on line 9 rather than stubbing?

jcoyne (changes requested):

spec/components/works/add_files_component_spec.rb
mjgiarlo: I tried that and it didn't work. NoMethodError: Cannot call `object` on `nil`.

mjgiarlo: The formbuilder instance that is created on line 9 is different than the one instantiated by the view component, apparently. They have different object IDs, which explains the NoMethodError. Hence the gross any_instance jazz.

@jcoyne :point_up:

spec/components/works/add_files_component_spec.rb
```diff
@@ -3,12 +3,35 @@
...
+
+    it 'renders the component' do
+      expect(rendered.to_html).to include('Add your files')
+      expect(rendered.css('button.dz-clickable').to_html).to include('Choose files')
+    end
+  end
+
+  context 'with a persisted file component' do
+    let(:model) { create(:attached_file, :with_file) }
+    let(:object) { double('anonymous form instance', persisted?: true, model: model) }
```
jcoyne, line 27:
```suggestion
  let(:form) { ActionView::Helpers::FormBuilder.new(nil, work_form, controller.view_context, {}) }
  let(:rendered) { render_inline(described_class.new(form: form)) }
  let(:work_form) { WorkForm.new(object) }

  context 'with an unpersisted file component' do
    let(:object) { build(:work) }

    it 'renders the component' do
      expect(rendered.to_html).to include('Add your files')
      expect(rendered.css('button.dz-clickable').to_html).to include('Choose files')
    end
  end

  context 'with a persisted file component' do
    let(:model) { create(:attached_file, :with_file) }
    let(:object) { build(:work, attached_files: [model]) }
```

jcoyne (changes requested):

spec/components/works/add_files_component_spec.rb
jcoyne: How about this?

mjgiarlo: Looks good. Looks like the trick is injecting a form as the "object" of the form builder? I'll patch it in. Thanks!

jcoyne: yes.

mjgiarlo: @jcoyne pushed. have another look?

jcoyne (approved):

PR #477, opened 2020-11-17
publisher contributors correctly mapped to cocina event, not cocina contributor

ndushay:
## Why was this change made?

Part of #240;   when a contributor in h2 is a "Publisher" it maps to a Cocina "Event" not a "Contributor".  And for fun, you have to combine it with the publication date which is optional.  And for even more fun, Cocina model objects are immutable, so ... 

unit tests from Arcadia's h2 mapping spec for contributors AND a couple of my own added in.

### DIFF
--- a/app/services/contributors_generator.rb
+++ b/app/services/contributors_generator.rb
@@ -16,14 +16,24 @@ class ContributorsGenerator
     new(work: work).form_array_from_contributor_event
   end
 
+  sig do
+    params(work: Work, pub_date: T.nilable(Cocina::Models::Event))
+      .returns(T::Array[Cocina::Models::Event])
+  end
+  def self.events_from_publisher_contributors(work:, pub_date: nil)
+    new(work: work).publication_event_values(pub_date)
+  end
+
   sig { params(work: Work).void }
   def initialize(work:)
     @work = work
   end
 
+  # H2 Publisher becomes a Cocina::Models::Event, not a Contributor.  See events_from_publisher_contributors.
   sig { returns(T::Array[T.nilable(Cocina::Models::Contributor)]) }
   def generate
-    work.contributors.map { |work_form_contributor| contributor(work_form_contributor) }
+    work.contributors.reject { |c| c.role == 'Publisher' }
+        .map { |work_form_contributor| contributor(work_form_contributor) }
   end
 
   ROLES_FOR_FORM = %w[Event Conference].freeze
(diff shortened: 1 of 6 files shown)


app/services/contributors_generator.rb
```diff
@@ -16,14 +16,24 @@ def self.form_array_from_contributor_event(work:)
     new(work: work).form_array_from_contributor_event
   end
 
+  sig do
+    params(work: Work, pub_date: T.nilable(Cocina::Models::Event))
+      .returns(T::Array[Cocina::Models::Event])
+  end
```
ndushay, line 22: personally, i find doing this for "line too long" rubocop exceptions to be annoying and to make it harder to identify the actual method, but I'm playing along with the team.

app/services/contributors_generator.rb
```diff
@@ -42,12 +52,32 @@ def form_array_from_contributor_event
...
+    end
+  end
+  # rubocop:enable Metrics/MethodLength
+
   private
 
   sig { returns(Work) }
   attr_reader :work
 
-  sig { params(contributor: Contributor).returns(T.nilable(Cocina::Models::Contributor)) }
```
ndushay, line 50: fun times (not) with sorbet type checking madness.

sorbet/custom/cocina-models.rbi
```diff
@@ -29,8 +29,12 @@ module Cocina::Models
   end
 
   class Event
-    sig { params(type: String, date: T::Array[T.any(DescriptiveValue, T::Hash[T.untyped, T.untyped])]).void }
-    def initialize(type:, date:); end
+    sig do
+      params(type: String,
+             date: T.nilable(T::Array[T.any(DescriptiveValue, T::Hash[T.untyped, T.untyped])]),
+             contributor: T.nilable(T::Array[T.any(Contributor, T::Hash[T.untyped, T.untyped])])).void
```
ndushay, line 35: turns out events can have contributors.  who knew?

app/services/contributors_generator.rb
```diff
@@ -42,12 +52,32 @@ def form_array_from_contributor_event
     ]
   end
 
+  # rubocop:disable Metrics/MethodLength
+  sig { params(pub_date: T.nilable(Cocina::Models::Event)).returns(T::Array[Cocina::Models::Event]) }
+  def publication_event_values(pub_date)
+    work.contributors.select { |c| c.role == 'Publisher' }.map do |publisher|
+      if pub_date
+        Cocina::Models::Event.new(
```
justinlittman, line 60: I'd recommend something like:
```
event = {
  type: 'publication',
  contributor: [contributor(publisher)]
}
event[:date] = pub_date.date if pub_date

Cocina::Models::Event.new(event)
```

justinlittman (approved):
Approve pending consideration of suggestion.

app/services/contributors_generator.rb
ndushay: that's brilliant!  thank you.

app/services/contributors_generator.rb
mjgiarlo: I kinda prefer this arrangement of breaking up the `params` and the `returns`, but I don't care too much either way.

ndushay: that doesn't bother me -- it's the "do" and "end" that make it look like a method.

PR #482, opened 2020-11-17
add validation for at least one contributor

### LINKED ISSUE #386: Validate that there is at least one contributor (with a role) is added for deposit
At least one contributor and the associated role must be added to validate for deposit

peetucket:
## Why was this change made?

Fixes #386 

- validate that there is at least one contributor at the work form level, as well as client side.  
- Need to disable client side validation for specific fields depending on role selected, which is done via the same JS methods that hide/show input fields.
- Rename `contributor` JS controller to `contributors` controller to be consistent with keywords

Updated tests to add contributors to continue to get them to pass now that contributor is required.
Added some new unit tests

### DIFF
--- a/app/components/works/contributor_row_component.html.erb
+++ b/app/components/works/contributor_row_component.html.erb
@@ -1,4 +1,4 @@
-<div class="contributors row inner-container" data-controller="contributors" data-action="error->contributors#error" data-target="edit-deposit.contributorsField">
+<div class="row inner-container" data-controller="contributors" data-action="error->contributors#error" data-target="edit-deposit.contributorsField">
   <div class="contributors-container col-md-12" data-target="contributors.container">
     <%= button_tag type: 'button', class: 'btn btn-sm float-right', aria: { label: 'Remove' },
         data: { action: "click->nested-form#removeAssociation" } do %>
--- a/app/components/works/contributor_row_component.html.erb
+++ b/app/components/works/contributor_row_component.html.erb
@@ -1,5 +1,5 @@
-<div class="contributor row inner-container" data-controller="contributor" data-target="edit-deposit.contributorsField">
-  <div class="col-md-12">
+<div class="contributors row inner-container" data-controller="contributors" data-action="error->contributors#error" data-target="edit-deposit.contributorsField">
+  <div class="contributors-container col-md-12" data-target="contributors.container">
     <%= button_tag type: 'button', class: 'btn btn-sm float-right', aria: { label: 'Remove' },
         data: { action: "click->nested-form#removeAssociation" } do %>
       <span class="far fa-trash-alt"></span>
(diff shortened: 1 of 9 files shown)


app/components/works/contributor_row_component.html.erb
```diff
@@ -1,5 +1,5 @@
-<div class="contributor row inner-container" data-controller="contributor">
-  <div class="col-md-12">
+<div class="contributors row inner-container" data-controller="contributors" data-action="error->contributors#error" data-target="edit-deposit.contributorsField">
```
mjgiarlo, line 1: Do you know if we're using this `contributors` (or `contributor`) class, or if it's a placeholder?

spec/forms/work_form_spec.rb
```diff
@@ -53,12 +54,37 @@
     end
   end
 
+  describe 'contributors validation' do
+    let(:contributor_error) { { contributors: ['Please add at least one contributor.'] } }
+
+    context 'with no contributors' do
+      it 'does not validate' do
+        form.valid?
```
mjgiarlo, line 62: I wonder if you could rework this a bit, like:

```ruby
expect(form).not_to be_valid
expect(form.errors.messages.to include(contributor_error)
```

spec/forms/work_form_spec.rb
```diff
@@ -53,12 +54,37 @@
...
+        expect(form.errors.messages).to include(contributor_error)
+      end
+    end
+
+    context 'with contributors' do
+      let(:form_with_contributors) { described_class.new(work_with_contributors) }
```
mjgiarlo, line 68: Could you replace this line with:

```ruby
  let(:work) { build(:work, :with_contributors) }
```

and then you can use the form object above which should be lazily evaluated with the "right" work here, rather than creating a parallel structure? if so, then you can also dump the `work_with_contributors` `let` statement above.

spec/forms/work_form_spec.rb
```diff
@@ -53,12 +54,37 @@
...
+
+      it 'validates' do
+        form_with_contributors.valid?
```
mjgiarlo, line 71: same question as above re: `expect(form).to be_valid`

spec/forms/work_form_spec.rb
```diff
@@ -53,12 +54,37 @@
...
+  end
+
   describe 'email validation' do
     it 'does not validate with an invalid contact email' do
       form.validate(contact_email: 'notavalidemail')
       expect(form).not_to be_valid
       expect(form.errors.messages).to include({ contact_email: ['is invalid'] })
     end
+
+    it 'validates with a correct contact email' do
```
mjgiarlo, line 84: good addition!

spec/requests/works_spec.rb
```diff
@@ -313,6 +314,11 @@
           )
         end
 
+        let(:contributors) do
+          { '999' =>
+            { '_destroy' => 'false', 'full_name' => 'Stanford', 'role_term' => 'organization|Host institution' } }
```
mjgiarlo, line 319: Don't all three name attrs come across in both cases? might want to add first name and last name here (set to empty strings) for higher fidelity w/ the deployed app.

spec/requests/works_spec.rb
```diff
@@ -389,6 +396,12 @@
           }
         end
 
+        let(:contributors) do
+          { '999' =>
```
mjgiarlo, line 400: As above, consider adding full_name?

mjgiarlo (commented):
Some comments and questions

spec/requests/works_spec.rb
peetucket: Done.

spec/requests/works_spec.rb
peetucket: Yup. Done.

spec/forms/work_form_spec.rb
peetucket: yes, had to build a more complete work object to actually make it valid in order to do this (else it would not be valid even with contributors), but still looks better

spec/forms/work_form_spec.rb
peetucket: sounds good, done

app/components/works/contributor_row_component.html.erb
peetucket: Does not appear to be used - I tested without the class and see no difference.  Will just remove to get rid of cruft.

app/javascript/controllers/contributors_controller.js
```diff
@@ -0,0 +1,44 @@
+import { Controller } from "stimulus"
+
```
peetucket, line 2: Note that this is not an entirely new controller - it was renamed from `contributor_controller.js` to `contributors_controller.js` and also changed a bit.

mjgiarlo (approved):

app/javascript/controllers/contributors_controller.js
mjgiarlo: soon we will document why this is so

PR #486, opened 2020-11-17
[Full Team Sign Off] Rubocop adjustments

ndushay:
## Why was this change made?

Per discussion in dev planning 2020-11-17, here is a PR to increase the class length max and the method length max for rubocop.  
- the first commit is a refactor to slightly reorganize the rubocop.yml file (for ease of maintenance);  
- the second commit is the rule changes
- the third commit is the inline comments that must be removed with these rule changes
- the fourth commit is regenerating rubocop_todo (by hand, so maxes aren't increased)

Weigh in below, check the box when you've reviewed the PR at least once.  If there's lively discussion below, we'll move this to an infradev meeting
- [x] @aaron-collier 
- [x] @jermnelson 
- [x] @jmartin-sul 
- [x] @justinlittman 
- [x] @jcoyne 
- [x] @mjgiarlo 
- [x] @peetucket

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -6,12 +6,19 @@ require:
 
 inherit_from: .rubocop_todo.yml
 
+#### Local Settings
+
 AllCops:
   Exclude:
     - db/**/*
     - node_modules/**/*
     - bin/**/*
     - vendor/**/*
+
+Lint/MissingSuper: # (new in 0.89)
+  Enabled: true
+  Exclude:
+    - app/components/**/*
 
 Metrics/BlockLength:
   Exclude:
@@ -21,23 +28,25 @@ RSpec/DescribeClass:
   Exclude:
     - spec/features/**/*
     - spec/requests/**/*
+
 RSpec/EmptyExampleGroup:
   Exclude:
     - spec/policies/**/*
+
 RSpec/ExampleLength:
   Enabled: false
+
 RSpec/MultipleMemoizedHelpers:
   Max: 15
+
 RSpec/NestedGroups:
   Max: 5
+
 RSpec/ScatteredSetup:
   Exclude:
     - spec/policies/**/*
 
-Lint/MissingSuper: # (new in 0.89)
-  Enabled: true
-  Exclude:
-    - app/components/**/*
+#### Turning on New Rules
 
 Layout/BeginEndAlignment: # (new in 0.91)
   Enabled: true
@@ -45,6 +54,7 @@ Layout/EmptyLinesAroundAttributeAccessor: # (new in 0.83)
   Enabled: true
 Layout/SpaceAroundMethodCallOperator: # (new in 0.82)
   Enabled: true
+
 Lint/BinaryOperatorWithIdenticalOperands: # (new in 0.89)
   Enabled: true
 Lint/ConstantDefinitionInBlock: # (new in 0.91)
@@ -53,9 +63,13 @@ Lint/DeprecatedOpenSSLConstant: # (new in 0.84)
   Enabled: true
 Lint/DuplicateElsifCondition: # (new in 0.88)
   Enabled: true
+Lint/DuplicateRegexpCharacterClassElement: # (new in 1.1)
+  Enabled: true
 Lint/DuplicateRequire: # (new in 0.90)
   Enabled: true
 Lint/DuplicateRescueException: # (new in 0.89)
+  Enabled: true
+Lint/EmptyBlock: # (new in 1.1)
   Enabled: true
 Lint/EmptyConditionalBody: # (new in 0.89)
   Enabled: true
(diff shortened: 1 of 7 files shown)


.rubocop.yml
```diff
@@ -6,45 +6,61 @@ require:
...
     - db/**/*
     - node_modules/**/*
     - bin/**/*
     - vendor/**/*
 
+Lint/MissingSuper: # (new in 0.89)
+  Enabled: true
+  Exclude:
+    - app/components/**/*
+
```
ndushay, line 22: moved up to alphabetize

.rubocop.yml
```diff
@@ -6,45 +6,61 @@ require:
...
+  Max: 150
+
+Metrics/MethodLength:
+  Max: 20
+
 RSpec/DescribeClass:
   Exclude:
     - spec/features/**/*
     - spec/requests/**/*
+
```
ndushay, line 37: spaced them out for easier skimming

.rubocop.yml
```diff
@@ -6,45 +6,61 @@ require:
...
-    - app/components/**/*
+#### Turning on New Rules
 
 Layout/BeginEndAlignment: # (new in 0.91)
   Enabled: true
 Layout/EmptyLinesAroundAttributeAccessor: # (new in 0.83)
   Enabled: true
 Layout/SpaceAroundMethodCallOperator: # (new in 0.82)
   Enabled: true
+
```
ndushay, line 63: attempt to clearly delineate between different types of cops

.rubocop.yml
```diff
@@ -53,10 +69,14 @@ Lint/DeprecatedOpenSSLConstant: # (new in 0.84)
   Enabled: true
 Lint/DuplicateElsifCondition: # (new in 0.88)
   Enabled: true
+Lint/DuplicateRegexpCharacterClassElement: # (new in 1.1)
+  Enabled: true
```
ndushay, line 73: alphabetizing ...

.rubocop.yml
```diff
@@ -6,45 +6,61 @@ require:
...
+
 Metrics/BlockLength:
   Exclude:
     - spec/**/*
 
+Metrics/ClassLength:
+  Max: 150
+
+Metrics/MethodLength:
+  Max: 20
```
ndushay, line 31: cops actually changed in this PR

peetucket (approved):

jcoyne (approved):

jermnelson (approved):

.rubocop.yml
aaron-collier: For me, 20 pushes the limits of readability. I'm ok with it here, as we've discussed, but I'd prefer we avoid making it our "standard".

aaron-collier (approved):

.rubocop.yml
ndushay: @aaron-collier make me an offer.

aaron-collier: I'd be happy at 15.

ndushay: Just added a commit with method length max at 15.  It restores two inline exceptions for method length.  I can live with it ... unless we get a lot more of those for methods that have no clear refactor.

mjgiarlo: I wouldn't die on this hill but... I'll ask: how would folks feel about 10?

.rubocop_todo.yml
```diff
@@ -6,35 +6,20 @@
 # Note that changes in the inspected code, or installation of new
 # versions of RuboCop, may require this file to be generated again.
 
-# Offense count: 3
+# Offense count: 2
```
mjgiarlo, line 9: Are the hand edits here the same as we'd get from `rubocop --auto-gen-config --auto-gen-only-exclude` ?

mjgiarlo (commented):
Overall, looking good! Have a couple questions.

.rubocop_todo.yml
ndushay: nope - it changes the max in some with that option.  that's why I ended up doing them by hand.

.rubocop.yml
ndushay: 10 is too small for me.  it is the default;  this is me wanting to increase the default.

ndushay: or Mike, are you saying you would prefer to keep it at 10, but you could compromise?  While I am saying I would prefer it to be 20, and I have compromised down to 15 ... but we are not agreeing yet?

mjgiarlo: @ndushay I thought the default was 5, not 10. :laughing: let's stick with 15 for now.

mjgiarlo (approved):

justinlittman (approved):

jmartin-sul (approved):

PR #488, opened 2020-11-17
Add embargo_date validation and display results

### LINKED ISSUE #288: Validate embargo is in the valid range
From the design doc:
> Latest date is 3 years from today’s date or the latest date selected by the collection manager

aaron-collier:
## Why was this change made?

Fixes #288 

Displays an error message if the release date selected is greater than 3 years from today.

Screen Shot 2020-11-18 at 12 19 33 PM

Unit test

N/A

### DIFF
--- a/app/components/works/embargo_component.html.erb
+++ b/app/components/works/embargo_component.html.erb
@@ -11,8 +11,8 @@
      </div>
    </div>
 
-   <div class="embargo_date" data-controller="embargo_date" data-action="error->embargo_date#error" data-target="edit-deposit.embargo_dateField">
-   <div class="mb-5 row" data-target="embargo_date.container">
+   <div class="embargo-date" data-controller="embargo-date" data-action="error->embargo-date#error" data-target="edit-deposit.embargo-dateField">
+   <div class="mb-5 row" data-target="embargo-date.container">
      <div class="col-sm-10 release-option">
           <%= form.radio_button :release, 'embargo', class: 'form-check-input' %>
           <%= form.label :release_embargo, 'On this date' %>
@@ -32,9 +32,9 @@
             <%= select_day nil, { prefix: 'work', field_name: 'embargo_date(3i)', prompt: 'day'}, id: 'work_embargo_day', class: "form-control" %>
           </div>
         </div>
-        <div data-target="embargo_date.error" class="invalid-feedback">blah</div>
       </div>
-   </div>
+      <div data-target="embargo-date.error" class="invalid-feedback"></div>
+      </div>
 
   <p>Select which audience you would like to have access to download your files once they are available.</p>
   <div class="mb-3 row">
--- a/app/components/works/embargo_component.html.erb
+++ b/app/components/works/embargo_component.html.erb
@@ -11,26 +11,29 @@
      </div>
    </div>
 
+   <div class="embargo_date" data-controller="embargo_date" data-action="error->embargo_date#error" data-target="edit-deposit.embargo_dateField">
    <div class="mb-5 row">
      <div class="col-sm-10 release-option">
-       <%= form.radio_button :release, 'embargo', class: 'form-check-input' %>
-       <%= form.label :release_embargo, 'On this date' %>
+          <%= form.radio_button :release, 'embargo', class: 'form-check-input' %>
+          <%= form.label :release_embargo, 'On this date' %>
 
-       <div class="year">
-         <label for="work_embargo_year" class="visually-hidden">Embargo year</label>
-         <%= select_year Time.zone.today, { prefix: 'work', field_name: 'embargo_date(1i)', start_year: Time.zone.today.year, end_year: 3.years.from_now.year }, id: 'work_embargo_year', class: "form-control" %>
-       </div>
-       <div class="month">
-         <label for="work_embargo_month" class="visually-hidden">Embargo month</label>
-         <%= select_month nil, { prefix: 'work', field_name: 'embargo_date(2i)', prompt: 'month'}, id: 'work_embargo_month', class: "form-control" %>
-       </div>
+          <div class="year">
+            <label for="work_embargo_year" class="visually-hidden">Embargo year</label>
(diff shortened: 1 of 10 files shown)


app/components/works/embargo_component.html.erb
```diff
@@ -11,26 +11,30 @@
...
+          </div>
+
+          <div class="day">
+            <label for="work_embargo_day" class="visually-hidden">Embargo day</label>
+            <%= select_day nil, { prefix: 'work', field_name: 'embargo_date(3i)', prompt: 'day'}, id: 'work_embargo_day', class: "form-control" %>
+          </div>
+        </div>
+      </div>
+      <div data-target="embargo-date.error" class="invalid-feedback"></div>
+      </div>
```
jcoyne, line 37: Please fix indentation.

app/javascript/controllers/edit_deposit_controller.js
```diff
@@ -29,7 +30,7 @@ export default class extends Controller {
   displayErrors(event) {
     const [data, _status, _xhr] = event.detail;
     for (const [fieldName, errorList] of Object.entries(data)) {
-      const key = `${fieldName}Field`
+      const key = `${fieldName.replace("_", "-")}Field`
```
jcoyne, line 33: Have you considered putting this transform logic go into `app/views/works/errors.json.jbuilder` instead?  We're already doing some field name transformation there.

app/javascript/controllers/embargo_date_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  // static targets = ["search", "container", "error", "selectedTemplate", "addItem"]
+  static targets = ["container", "error"]
+
+  // Triggered when edit-deposit controller sends an error event
+  error(e) {
+    console.log("HERE?")
```
jcoyne, line 9: cruft?

app/javascript/controllers/embargo_date_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  // static targets = ["search", "container", "error", "selectedTemplate", "addItem"]
```
jcoyne, line 4: cruft?

jcoyne (changes requested):

app/javascript/controllers/edit_deposit_controller.js
mjgiarlo: good call

app/components/works/embargo_component.html.erb
aaron-collier: Should this one be indented? It looks a bit odd as a one liner there, but I think the indentation is correct.

jcoyne: shouldn't it be outdented (note there is a close `</div>` on line 36)?

app/javascript/controllers/edit_deposit_controller.js
```diff
@@ -3,7 +3,8 @@ import { Controller } from "stimulus";
 export default class extends Controller {
   static targets = ["title", "titleField",
                     "file", "fileField",
-                    "keywordsField", "contributorsField"];
+                    "keywordsField", "contributorsField",
+                    "embargo_dateField"];
```
jcoyne, line 7: given that https://github.com/sul-dlss/happy-heron/pull/488/files#diff-b4ce442316cd2ea232e97f158d51b8e9b60077a59e600f7a1b76191fcf33b841R14 is `embargo-dateField`, shouldn't this match?

jcoyne (changes requested):

app/javascript/controllers/edit_deposit_controller.js
jcoyne: And, not particularly important, but `embargoDateField` would be more aesthetically pleasing.

aaron-collier: I can try that, but this was what we discussed flipping between the javascript magic and the rails magic...

app/views/works/errors.json.jbuilder
```diff
@@ -3,4 +3,5 @@
 @form.errors.each do |key, value|
   # Transform the property name in the model to the value used by the javascript:
   json.set! key == :attached_files ? 'file' : key, [value]
+  json.set! key == :embargo_date ? 'embargo-date' : key, [value]
```
jcoyne, line 6: won't this second thing cause a duplicate set of keys in the response?

jcoyne (changes requested):

app/javascript/controllers/edit_deposit_controller.js
jcoyne: There is no magic here.  These names just need to match what you wrote in the HTML.

aaron-collier: Yeah, that works, fixed.

app/views/works/errors.json.jbuilder
aaron-collier: Can you suggest an alternative?

jcoyne: call json.set only one time. adjust the parameters prior to calling it.

aaron-collier: Moved this into a helper method.

app/javascript/controllers/edit_deposit_controller.js
aaron-collier: My first pass at renaming broke this, I'll take another swipe, but wouldn't hold this up over it myself.

app/controllers/application_controller.rb
```diff
@@ -21,6 +21,14 @@ def user_with_groups
   end
   helper_method :user_with_groups
 
+  def normalize_key(key)
```
jcoyne, line 24: Can this be in the works controller rather than the application_controller since it's only used by the one?

jcoyne (changes requested):

app/controllers/application_controller.rb
aaron-collier: done.

jcoyne (approved):

PR #491, opened 2020-11-18
Don't show "not deposited" when work is a version draft

### LINKED ISSUE #487: Show page indicates "draft" after deposit
When you create a draft, then go back and edit it, and press the "deposit" button. It takes you to the show page and says "Draft - Not depoisted", which is confusing, because the deposit button has been pushed.  It stays a "draft" until the background job updates it with a druid, which could take several minutes.  Instead it should say "depositing" or something similar.

Code is:
```
      <% unless @work.deposited? %><span class="font-italic" style="margin-left: 30px;">Draft - Not deposited</span><% end %>
```

jcoyne:
This message is misleading

Fixes #487


app/components/works/detail_component.html.erb
```diff
@@ -0,0 +1,214 @@
+<section class="details">
```
jcoyne, line 1: Moved this in from `app/views/works/show.html.erb` so that we can easily test the new condition.

mjgiarlo (approved):

app/components/works/detail_component.html.erb
mjgiarlo: :ok_hand:

PR #496, opened 2020-11-18
contributor -> cocina model gets status=primary

ndushay:
~~[DRAFT] until I get a final confirmation from Arcadia that the implementation of this is as she desired.~~
~~[DRAFT] until team agrees on ordering and primary status implications from H2 UI and cocina models.~~

## Why was this change made?

Part of #240.

unit tests from Arcadia's h2 mapping spec for contributors, with a bunch of TODO / FIXME comments removed!

### DIFF
--- a/app/services/contributors_generator.rb
+++ b/app/services/contributors_generator.rb
@@ -32,8 +32,14 @@ class ContributorsGenerator
   # H2 Publisher becomes a Cocina::Models::Event, not a Contributor.  See events_from_publisher_contributors.
   sig { returns(T::Array[T.nilable(Cocina::Models::Contributor)]) }
   def generate
+    count = 0
     work.contributors.reject { |c| c.role == 'Publisher' }
-        .map { |work_form_contributor| contributor(work_form_contributor) }
+        .map do |work_form_contributor|
+          count += 1
+          # First entered contributor is always status: "primary" (except for Publisher)
+          primary = count == 1
+          contributor(work_form_contributor, primary)
+        end
   end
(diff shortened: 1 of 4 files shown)


app/services/contributors_generator.rb
```diff
@@ -32,8 +32,14 @@ def initialize(work:)
   # H2 Publisher becomes a Cocina::Models::Event, not a Contributor.  See events_from_publisher_contributors.
   sig { returns(T::Array[T.nilable(Cocina::Models::Contributor)]) }
   def generate
+    count = 0
     work.contributors.reject { |c| c.role == 'Publisher' }
-        .map { |work_form_contributor| contributor(work_form_contributor) }
+        .map do |work_form_contributor|
+          count += 1
+          # First entered contributor is always status: "primary" (except for Publisher)
```
jcoyne, line 39: @astridu @amyehodge  I'm not sure that the form makes it clear that the first author is "primary"

mjgiarlo: @peetucket We may want to set up a brief check-in today about this. Also including @arcadiafalcone @ndushay @jcoyne and myself.

jcoyne: We're setting up a meeting on Friday where we can discuss this.

ndushay: I'll convert this PR to draft, then.  I thought "primary" was a separate animal from "order" but perhaps not.

jcoyne (approved):

PR #498, opened 2020-11-18
Prevent form submission when pressing return in the keyword widget

### LINKED ISSUE #476: Typing a keyword into the keyword box and hitting enter should save the keyword but not save the entire form.
I have replicated this issue on qa. 

When I type in a keyword and then hit enter, I am taking immediately to the item view page. It seems hitting enter here is saving the draft of the item but should just be saving the keyword in this field.

jcoyne:
## Why was this change made?
Fixes #476

### DIFF
--- a/app/javascript/controllers/keywords_controller.js
+++ b/app/javascript/controllers/keywords_controller.js
@@ -29,6 +29,13 @@ export default class extends Controller {
     this.containerTarget.classList.remove('is-invalid')
   }
 
+  checkForEnter(event) {
+    if (event.key === 'Enter') {
+      event.preventDefault() // prevent form submission
+      this.add(event);
+    }
+  }
+
   removeAssociation(event) {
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/keywords_controller.js
```diff
@@ -29,6 +29,13 @@ export default class extends Controller {
     this.containerTarget.classList.remove('is-invalid')
   }
 
+  checkForEnter(event) {
+    if (event.keyCode === 13) {
```
mjgiarlo, line 33: Would you consider to changing this to `event.key === 'Enter'`? I read earlier today that `.keyCode` is deprecated.

jcoyne: Done. Thanks for the pointer.

mjgiarlo (approved):

PR #506, opened 2020-11-18
Add a depositing state

### LINKED ISSUE #490: Add a "depositing" state in the state machine
This would occur between when the users presses "deposit" and when the item is actually deposited.  This can prevent edits until the  deposit has been accepted in SDR.

jcoyne:
## Why was this change made?

So the user can see when their work has been submitted for deposit, but isn't done yet.
Fixes #490

### DIFF
--- a/app/controllers/reviews_controller.rb
+++ b/app/controllers/reviews_controller.rb
@@ -10,6 +10,7 @@ class ReviewsController < ApplicationController
     work = Work.find(params[:work_id])
     authorize! work, to: :review?
     if params[:state] == 'approve'
+      work.begin_deposit!
       DepositJob.perform_later(work)
     else
       Event.create!(work: work, user: current_user, event_type: 'return', description: params[:reason])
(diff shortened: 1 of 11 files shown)


app/controllers/reviews_controller.rb
```diff
@@ -10,6 +10,7 @@ def create
     work = Work.find(params[:work_id])
     authorize! work, to: :review?
     if params[:state] == 'approve'
+      work.begin_deposit!
```
mjgiarlo, line 13: I see we call this above each invocation of `DepositJob.perform_later(work)`, but I assume we *don't* want to move this code to the job because we want the state change to happen immediately. True?

mjgiarlo (approved):

app/controllers/reviews_controller.rb
jcoyne: Yes, it should be set to the new state even while the job is enqueue.

PR #509, opened 2020-11-18
Add a banner that displays when a deposit is sent for review

### LINKED ISSUE #461: Banner when approval needed
If form is filled out properly, and the collection uses the review workflow, then the user returns to deposit view page with a banner message that says 
“Your deposit has been sent for approval. You will receive an email once your deposit has been approved.”

jcoyne:
## Why was this change made?
Fixes #461

### DIFF
--- a/app/components/works/detail_component.rb
+++ b/app/components/works/detail_component.rb
@@ -60,5 +60,10 @@ module Works
     def keywords
       work.keywords.map(&:label).join(', ')
     end
+
+    sig { returns(T::Boolean) }
+    def display_approval?
+      work.pending_approval? && !helpers.allowed_to?(:review?, work)
+    end
   end
 end
--- a/app/components/works/detail_component.html.erb
+++ b/app/components/works/detail_component.html.erb
@@ -1,4 +1,11 @@
 <section class="details">
+  <% if display_approval? %>
+    <div class="alert alert-warning alert-dismissible" role="alert">
+      Your deposit has been sent for approval. You will receive an email once your deposit has been approved.
+      <button type="button" class="btn-close" data-dismiss="alert" aria-label="Close"></button>
+    </div>
+  <% end %>
+
   <header class="title">
     <span class="header-text"><%= title %></span>
     <%= link_to edit_work_path(@work), aria: { label: "Edit #{title}"} do %><span class="fas fa-pencil-alt edit"></span><% end %>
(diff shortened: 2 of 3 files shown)


app/components/works/detail_component.rb
```diff
@@ -64,5 +64,10 @@ def keywords
     def draft?
       work.first_draft?
     end
+
+    sig { returns(T::Boolean) }
+    def display_approval?
+      work.pending_approval? && !helpers.allowed_to?(:review?, work)
```
peetucket, line 70: i'm unfamiliar with `helpers.allowed_to?(XXXX)` ... where is that defined/coming from?

peetucket (approved):

peetucket:
Looks like a rebase is needed now

app/components/works/detail_component.rb
mjgiarlo: `helpers` is a container object that is available to all view components, and `allowed_to?` is a helper method set up by ActionPolicy.

peetucket: ok, so `app/policies/work_policy.rb` is being used (because we are asking about a `work`) and we are hitting the `review?` method defined there

i'm familiar with the traditional rails helpers available to views, so there is an equivalent for components but stored elsewhere?  thanks

mjgiarlo: @peetucket :speech_balloon: 
> ok, so `app/policies/work_policy.rb` is being used (because we are asking about a `work`) and we are hitting the `review?` method defined there

Yep!
 
> i'm familiar with the traditional rails helpers available to views, so there is an equivalent for components but stored elsewhere? thanks

Yeah, they're the same helpers, and the ViewComponent library pulls them in via this `helpers` method which is available to all view components: https://github.com/github/view_component/blob/ec91c69a0552b90e5d6a891ebb5dd642ed44f2f8/lib/view_component/base.rb#L117-L121

peetucket: thanks for the explanations and sorry if missed these earlier

and it looks like actionpolicy (and the state machine used for works/collections) are gems and not core rails

mjgiarlo: Correct!

mjgiarlo:
@jcoyne you wanna fix the conflict or back away from this or what?

jcoyne:
@mjgiarlo I don't know if the changes you made recently to the status satisfy this requirement already.

mjgiarlo:
@jcoyne ok, maybe we should discuss with @amyehodge

amyehodge:
@mjgiarlo and @jcoyne I'm happy to help, but not sure what the question is.

mjgiarlo (approved):

mjgiarlo:
@amyehodge no worries! @jcoyne and I just chatted about this and it's good to go. This adds a banner to the work show page to set users' expectations a bit better w/r/t works waiting on approval.

PR #526, opened 2020-11-19
Update work show page with state changes and new PURLs

### LINKED ISSUE #508: Display changes on work show page when deposit finishes
Show the "depositing" state and change it to "deposited" once done (via ActionCable), and also show the shiny new PURL.

mjgiarlo:
Fixes #508

## Why was this change made?

This commit makes the work show page more interactive by updating the display dynamically, via ActionCable, whenever the work undergoes state changes (moving from draft to depositing to deposited, etc.). It also updates the Purl link when the item gets a druid assigned to it. This makes for a more interactive experience, and prevents page reloads for updates. It also serves as a working example of how to use ActionCable to deliver user-facing features.

Question for @astridu: should we animate/highlight/flash these changes when they occur, or is it preferable to go with the more inobtrusive experience (see screen cap below to see current implementation).

Includes:
* Add an ActionCable channel for streaming work updates to clients
* Rename 'return' to 'reject' for consistent domain language across the app
* Add a Stimulus controller to the work show page to handle ActionCable interaction
* Create a new event service that makes sure every work state change is accompanied by an associated event and a broadcast sent to the WorkUpdatesChannel
* Allow the work model to return user-friendly labels for supported states

CI and local browser.

Peek 2020-11-19 13-08

None

### DIFF
--- a/app/services/event_service.rb
+++ b/app/services/event_service.rb
@@ -1,0 +1,48 @@
+# typed: strict
+# frozen_string_literal: true
+
+# Triggers state changes, logs events, and broadcasts changes
+class EventService
+  extend T::Sig
+
+  class << self
+    include ActionView::Helpers::UrlHelper
+  end
+
+  sig { params(work: Work, user: User, description: String).returns(T.nilable(Integer)) }
+  def self.reject(work:, user:, description:)
+    work.reject!
+    Event.create!(work: work, user: user, event_type: 'reject', description: description)
+    WorkUpdatesChannel.broadcast_to(work, state: work.current_state_display_label)
+  end
+
+  sig { params(work: Work, user: User, description: String).returns(T.nilable(Integer)) }
+  def self.begin_deposit(work:, user:, description: '')
+    work.begin_deposit!
+    Event.create!(work: work, user: user, event_type: 'begin_deposit', description: description)
+    WorkUpdatesChannel.broadcast_to(work, state: work.current_state_display_label)
+  end
+
+  sig { params(work: Work).returns(T.nilable(Integer)) }
+  def self.deposit_complete(work:)
+    work.deposit_complete!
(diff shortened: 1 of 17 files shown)


mjgiarlo:
@astridu See the question I left for you above. Thanks!

app/services/event_service.rb
```diff
@@ -0,0 +1,48 @@
...
+# Triggers state changes, logs events, and broadcasts changes
+class EventService
+  extend T::Sig
+
+  class << self
+    include ActionView::Helpers::UrlHelper
+  end
+
+  sig { params(work: Work, user: User, description: String).returns(T.nilable(Integer)) }
+  def self.reject(work:, user:, description:)
```
mjgiarlo, line 13: Because the methods in this class are so similar, I was tempted to refactor it but I could either way.

jcoyne (approved):

astridu:
It looks ok as is, but it was slightly confusing after you clicked the 'Deposit' button because the status still said 'Deposit in progress' for a couple seconds so I wasn't sure if it had done anything. I'm wondering if that should have a spinner next to it and then have 'Deposited' appear so it's clear it is working on depositing and then deposit is complete. Happy to chat if that doesn't make sense.

mjgiarlo:
@astridu :speech_balloon: 
> It looks ok as is, but it was slightly confusing after you clicked the 'Deposit' button because the status still said 'Deposit in progress' for a couple seconds so I wasn't sure if it had done anything. 

In this case, that's because the deposit does take 15-30 seconds. So it does take a wee bit of time. That said...

> I'm wondering if that should have a spinner next to it and then have 'Deposited' appear so it's clear it is working on depositing and then deposit is complete. Happy to chat if that doesn't make sense.

I really like that. I'll make it so!

PR #528, opened 2020-11-19
Deposit collections

### LINKED ISSUE #92: Collection changes should immediately deposit into SDR
Ref #56

When a collection is created for the first time, it should create the collection in SDR. This can be done with a background job similar to https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/deposit_job.rb

This may require work in SDR API, SDR-client, or elsewhere in SDR to support this in H2. It also may not.

jcoyne:
## Why was this change made?
Fixes #92

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -29,9 +29,7 @@ class CollectionsController < ObjectsController
 
     @form = collection_form(collection)
     if @form.validate(collection_params) && @form.save
-      # TODO: https://github.com/sul-dlss/happy-heron/issues/92
-      # DepositCollectionJob.perform_later(@collection) if deposit_button_pushed?
-      redirect_to dashboard_path
+      after_save(collection)
     else
       render :new
     end
@@ -43,15 +41,20 @@ class CollectionsController < ObjectsController
 
     @form = collection_form(collection)
     if @form.validate(collection_params) && @form.save
-      # TODO: https://github.com/sul-dlss/happy-heron/issues/92
-      # DepositCollectionJob.perform_later(@collection) if deposit_button_pushed?
-      redirect_to dashboard_path
+      after_save(collection)
     else
       render :edit
     end
   end
 
   private
+
+  sig { params(collection: Collection).void }
+  def after_save(collection)
+    deposit_button_pushed? ? collection.begin_deposit! : collection.update_metadata!
+
+    redirect_to dashboard_path
+  end
 
   def collection_form(collection)
     return CollectionForm.new(collection) if deposit_button_pushed?
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -14,4 +14,23 @@ class Collection < ApplicationRecord
   def review_enabled?
     reviewers.present?
   end
+
+  state_machine initial: :first_draft do
+    after_transition on: :begin_deposit do |collection, _transition|
+      DepositCollectionJob.perform_later(collection)
+    end
+
+    event :begin_deposit do
+      transition %i[first_draft version_draft deposited] => :depositing
+    end
+
+    event :deposit_complete do
+      transition depositing: :deposited
+    end
+
+    event :update_metadata do
+      transition deposited: :version_draft
+      transition %i[first_draft version_draft] => same
+    end
+  end
 end
--- a/app/jobs/deposit_collection_job.rb
+++ b/app/jobs/deposit_collection_job.rb
@@ -1,0 +1,51 @@
+# typed: false
+# frozen_string_literal: true
+
+# Deposits a Collection into SDR API.
+class DepositCollectionJob < BaseDepositJob
+  queue_as :default
+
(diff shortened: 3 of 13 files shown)


mjgiarlo (approved):

app/controllers/collections_controller.rb
```diff
@@ -43,16 +41,25 @@ def update
 
     @form = collection_form(collection)
     if @form.validate(collection_params) && @form.save
-      # TODO: https://github.com/sul-dlss/happy-heron/issues/92
-      # DepositCollectionJob.perform_later(@collection) if deposit_button_pushed?
-      redirect_to dashboard_path
+      collection.update!
```
mjgiarlo, line 44: Can you say more about what this line is doing?

app/models/collection.rb
```diff
@@ -14,4 +14,18 @@ class Collection < ApplicationRecord
...
+    event :begin_deposit do
+      transition first_draft: :depositing, version_draft: :depositing
+    end
+
+    event :deposit_complete do
+      transition depositing: :deposited
+    end
+
+    event :update do
+      transition [:deposited, :version_draft] => :version_draft
```
mjgiarlo, line 28: What does this syntax mean? How is it different from the `begin_deposit` transition syntax?

mjgiarlo (approved):

app/controllers/collections_controller.rb
jcoyne: It's a state transition... maybe `update` is a bad name for an event.

app/models/collection.rb
jcoyne: because this means you don't have to check `work.deposited?` you can always have this event, because it's a no-op to go from version_draft back to version_draft.

app/controllers/collections_controller.rb
mjgiarlo: Oh! Yes, would you mind renaming?

app/models/collection.rb
mjgiarlo: I'm not sure I follow why we need this for collections (and not for works). Maybe a state diagram for collections would help.

jcoyne: We could have it for works, but you moved the transition to the EventService and wrapped it with a conditional: https://github.com/sul-dlss/happy-heron/blob/a138efe8fa2ee4975914303cfa662c95e42d89a8/app/controllers/works_controller.rb#L49

app/controllers/collections_controller.rb
jcoyne: renamed to `update_metadata`

mjgiarlo: Thank you, sir.

app/models/collection.rb
mjgiarlo: The conditional was there before the EventService came along: https://github.com/sul-dlss/happy-heron/commit/64b54490ca250aeb3b3d6fb34e4ac0f3057fc853#diff-26bc953ee940475a6382d93330c2315a15a75269bbf9ad0798c926f8c2b5a7c7L49

app/models/collection.rb
```diff
@@ -14,4 +14,23 @@ class Collection < ApplicationRecord
...
     reviewers.present?
   end
+
+  state_machine initial: :first_draft do
+    after_transition on: :begin_deposit do |collection, _transition|
+      DepositCollectionJob.perform_later(collection)
+    end
+
+    event :begin_deposit do
+      transition %i[first_draft version_draft deposited] => :depositing
```
mjgiarlo, line 24: This is an improvement. Thanks!

app/models/collection.rb
```diff
@@ -14,4 +14,23 @@ class Collection < ApplicationRecord
   def review_enabled?
     reviewers.present?
   end
+
+  state_machine initial: :first_draft do
+    after_transition on: :begin_deposit do |collection, _transition|
```
mjgiarlo, line 19: Heads-up to @sul-dlss/infrastructure-team that we're now leaning on our state machine library to handle side effects we want state changes to trigger. Note that these are automatically done in a transaction

app/models/collection.rb
```diff
@@ -14,4 +14,23 @@ class Collection < ApplicationRecord
...
+      transition %i[first_draft version_draft deposited] => :depositing
+    end
+
+    event :deposit_complete do
+      transition depositing: :deposited
+    end
+
+    event :update_metadata do
+      transition deposited: :version_draft
+      transition %i[first_draft version_draft] => same
```
mjgiarlo, line 33: I think I grok this now. This allows us to safely invoke the `update_metadata` event on collections in the `first_draft` and `version_draft` states without changing their states. Like you said, a no-op. :sunglasses:

mjgiarlo (approved):

app/models/collection.rb
jcoyne: yes!

PR #534, opened 2020-11-20
[Refactor] move display logic from the model to a component

jcoyne:
## Why was this change made?

The model is for concerns about persistence and the component is for presentation.


mjgiarlo (approved):
Nice.

PR #535, opened 2020-11-20
Move away from the event service by leveraging transition actions

jcoyne:

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: false
 # frozen_string_literal: true
 
 # Models the deposit of an digital repository object in H2.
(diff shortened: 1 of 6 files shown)


app/models/work.rb
```diff
@@ -44,8 +50,9 @@ class Work < ApplicationRecord
       transition pending_approval: :first_draft
     end
 
-    event :new_version do
+    event :update_metadata do
```
mjgiarlo, line 53: Could this event also include changes to the uploaded files? I'm wondering if `new_version` is more or less descriptive than `update_metadata` here.

jcoyne: I think new_version is not the right event, because the event is "the user presses the save button"

mjgiarlo: This event is fired whenever the user clicks the "Deposit" button after the initial press of the "Deposit" button, right? So it's not just metadata. But to the end user, it's not different from the first press of the "Deposit" button.

mjgiarlo: I'm OK with moving away from the event service, but I think I'd benefit from pairing a bit so we can discuss design.

app/models/work.rb
```diff
@@ -28,6 +28,12 @@ class Work < ApplicationRecord
   }
 
   state_machine initial: :first_draft do
+    after_transition deposited: :version_draft do |work, _transition|
+      Event.create!(work: work, user: work.depositor, event_type: 'new_version')
```
mjgiarlo, line 32: Should `new_version` also be changed to `update_metadata` here?

jcoyne: I don't think so, because new_version is really a name of a transition that happened.  Some update_metadata actions trigger that transition, but not all of them.

mjgiarlo: Let's chat about this today.

mjgiarlo (approved):

PR #549, opened 2020-11-23
Add collection membership to the cocina model

### LINKED ISSUE #515: Ensure that deposited items belong to the correct H2 related collection in Argo

jcoyne:
## Why was this change made?
So that the object has the correct collection in SDR. Fixes #515

### DIFF
--- a/app/services/file_generator.rb
+++ b/app/services/file_generator.rb
@@ -1,0 +1,98 @@
+# typed: true
+# frozen_string_literal: true
+
+# This generates a File for a work
+class FileGenerator
+  extend T::Sig
+
+  sig do
+    params(
+      work: Work,
+      attached_file: AttachedFile
+    ).returns(T.nilable(T.any(Cocina::Models::File, Cocina::Models::RequestFile)))
+  end
+  def self.generate(work:, attached_file:)
+    new(work: work, attached_file: attached_file).generate
+  end
+
+  sig { params(work: Work, attached_file: AttachedFile).void }
+  def initialize(work:, attached_file:)
+    @work = work
+    @attached_file = attached_file
+  end
+
+  attr_reader :work, :attached_file
+
+  sig { returns(T.nilable(T.any(Cocina::Models::File, Cocina::Models::RequestFile))) }
+  def generate
+    return nil unless blob
+
+    if work.druid
+      Cocina::Models::File.new(file_attributes)
+    else
+      Cocina::Models::RequestFile.new(request_file_attributes)
+    end
+  end
+
+  def request_file_attributes
+    {
+      type: 'http://cocina.sul.stanford.edu/models/file.jsonld',
+      version: work.version,
+      label: attached_file.label,
+      filename: filename,
+      access: access,
+      administrative: administrative,
+      hasMimeType: blob.content_type,
+      hasMessageDigests: message_digests,
+      size: blob.byte_size
+    }
+  end
+
+  def file_attributes
+    request_file_attributes.merge(externalIdentifier: external_identifier)
+  end
+
+  def filename
+    blob.filename.to_s # File.basename(filename(blob.key))
+  end
+
+  def external_identifier
+    "#{work.druid}/#{filename}" if work.druid
+  end
+
+  def administrative
+    {
+      sdrPreserve: true,
+      shelve: !attached_file.hide?
+    }
+  end
+
+  def message_digests
+    [
+      { type: 'md5', digest: base64_to_hexdigest(blob.checksum) },
+      { type: 'sha1', digest: Digest::SHA1.file(file_path(blob.key)).hexdigest }
+    ]
+  end
+
+  def blob
+    @blob ||= attached_file.file&.attachment&.blob
+  end
+
+  sig { returns(Hash) }
+  # TODO: This varies based on what the user selected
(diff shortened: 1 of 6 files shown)


app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  sig { params(work: Work, attached_file: AttachedFile).void }
+  def initialize(work:, attached_file:)
+    @work = work
+    @attached_file = attached_file
+  end
+
+  attr_reader :work, :attached_file
+
+  sig { returns(T.nilable(T.any(Cocina::Models::File, Cocina::Models::RequestFile))) }
```
mjgiarlo, line 26: Given that the class method above calls this instance method, I'm a wee bit surprised that the return value here is `nilable` but the return value above is not. I would have expected sorbet to make these signatures match. Seems like the class method definition needs the `nilable` bit too.

app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+      size: blob.byte_size
+    }
+  end
+
+  def file_attributes
+    request_file_attributes.merge(externalIdentifier: external_identifier)
+  end
+
+  def filename
+    blob.filename.to_s # File.basename(filename(blob.key))
```
mjgiarlo, line 56: What's the comment here for?

app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+  end
+
+  def external_identifier
+    "#{work.druid}/#{blob.filename}" if work.druid
```
mjgiarlo, line 60: Could `blob.filename` here be `filename`?

app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  def message_digests
+    [
+      { type: 'md5', digest: base64_to_hexdigest(blob.checksum) },
+      { type: 'sha1', digest: Digest::SHA1.file(file_path(blob.key)).hexdigest }
+    ]
+  end
+
+  def blob
+    @blob ||= attached_file.file&.attachment&.blob
```
mjgiarlo, line 78: Can you say more about the safe navigation here? Shouldn't an attached_file at this point have a blob? If so, I believe this can be shortened to `attached_file.blob`, thanks to https://github.com/sul-dlss/happy-heron/blob/main/app/models/attached_file.rb#L9-L11. But if we need the safe navigation in this context, obviously that won't work.

app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+    }
+  end
+
+  sig { params(key: String).returns(String) }
+  def file_path(key)
+    ActiveStorage::Blob.service.path_for(key)
+  end
+
+  def base64_to_hexdigest(base64)
+    Base64.decode64(base64).unpack1('H*')
```
mjgiarlo, line 96: This returns an MD5?

app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+  def administrative
+    {
+      sdrPreserve: true,
+      shelve: !attached_file.hide?
+    }
+  end
+
+  def message_digests
+    [
+      { type: 'md5', digest: base64_to_hexdigest(blob.checksum) },
```
mjgiarlo, line 72: I guess `blob.checksum` here is a b64-encoded MD5 value?

app/services/file_generator.rb
jcoyne: I don't know. It was in the RequestGenerator prior to this change. I assume it means that those are equivalent?

app/services/file_generator.rb
```diff
@@ -0,0 +1,98 @@
...
+      download: 'stanford'
+    }
+  end
+
+  sig { params(key: String).returns(String) }
+  def file_path(key)
+    ActiveStorage::Blob.service.path_for(key)
+  end
+
+  def base64_to_hexdigest(base64)
```
mjgiarlo, line 95: Can you add a comment here about needing this method bc ActiveStorage stores checksums as b64-encoded MD5s?

app/services/file_generator.rb
mjgiarlo: After spending ten minutes reviewing the changes in this file, I now see that it's a refactoring, so take all review up until now with a grain of salt.

app/services/structural_generator.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  attr_reader :work
+
+  sig { returns(T.any(Cocina::Models::DROStructural, Cocina::Models::RequestDROStructural)) }
+  def generate
+    klass = work.druid ? Cocina::Models::DROStructural : Cocina::Models::RequestDROStructural
+    klass.new(
+      contains: build_filesets,
+      isMemberOf: [work.collection.druid]
```
mjgiarlo, line 25: OK, so *this* is the change that adds the collection membership. Cool.

jcoyne: Yes

mjgiarlo (approved):

app/services/file_generator.rb
jcoyne: This was in the original code, I just moved it to a new class.

PR #553, opened 2020-11-23
Map access and embargo to the cocina model

### LINKED ISSUE #429: Map access and embargo to cocina
When I select "Everyone" can access my work, it shows as "World" in argo.

When I select "Stanford Community" can access my work, it shows "Stanford" access on the Argo side.

When "On this date" is selected, the embargo date should be visible in Argo.

jcoyne:
## Why was this change made?
Fixes #429

### DIFF
--- a/app/services/file_generator.rb
+++ b/app/services/file_generator.rb
@@ -79,12 +79,12 @@ class FileGenerator
   end
 
   sig { returns(Hash) }
-  # TODO: This varies based on what the user selected
   def access
-    {
-      access: 'stanford',
-      download: 'stanford'
-    }
+    if attached_file.hide?
+      { access: 'dark', download: 'none' }
+    else
+      { access: 'world', download: 'world' }
+    end
   end
 
   sig { params(key: String).returns(String) }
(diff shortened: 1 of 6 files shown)


app/services/file_generator.rb
```diff
@@ -79,12 +79,12 @@ def blob
...
-  # TODO: This varies based on what the user selected
   def access
-    {
-      access: 'stanford',
-      download: 'stanford'
-    }
+    if attached_file.hide?
+      { access: 'dark', download: 'none' }
+    else
+      { access: 'world', download: 'world' }
```
mjgiarlo, line 86: Should this vary based on the access specified for the work? Is it OK if the work is set to Stanford-only and its files are world?

jcoyne: @mjgiarlo currently Fedora 3 doesn't have a way to persist this, so it's either "world" (which defers to the object level access) or  "dark" which marks "publish=no".

mjgiarlo: OK.

mjgiarlo (approved):

PR #557, opened 2020-11-23
Do not display the deposit button if the collections state is in draft

### LINKED ISSUE #514: A deposit can not be made into a collection that is in first_draft or depositing states

aaron-collier:
## Why was this change made?

Fixes #514 by removing the deposit link if the collections status is `first_draft` or `depositing`.

Screen Shot 2020-11-23 at 3 56 46 PM

Unit and feature tests

N/A

### DIFF
--- a/app/components/dashboard/collection_component.html.erb
+++ b/app/components/dashboard/collection_component.html.erb
@@ -43,12 +43,14 @@
     <div class="mb-3">See all deposits</div>
   <% end %>
 
-  <%= button_tag '+ Deposit to this collection',
-        data: {
-          destination: new_collection_work_path(collection),
-          toggle: 'modal',
-          target: '#workTypeModal',
-          action: "work-type#set_collection"
-        },
-        class: "btn btn-primary" %>
+  <% if allowed_to?(:deposit?, collection) %>
+    <%= button_tag '+ Deposit to this collection',
+          data: {
+            destination: new_collection_work_path(collection),
+            toggle: 'modal',
+            target: '#workTypeModal',
+            action: "work-type#set_collection"
+          },
+          class: "btn btn-primary" %>
+  <% end %>
 </section>
--- a/spec/factories/collections.rb
+++ b/spec/factories/collections.rb
@@ -14,6 +14,7 @@ FactoryBot.define do
     default_license { 'MyString' }
     email_when_participants_changed { false }
     managers { 'maya.aguirre, jcairns' }
+    state { 'deposited' }
     creator
   end
 
--- a/spec/features/create_new_collection_spec.rb
+++ b/spec/features/create_new_collection_spec.rb
@@ -1,0 +1,31 @@
+# typed: false
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Create a new collection', js: true do
+  let(:user) { create(:user) }
+
+  before do
+    sign_in user, groups: ['dlss:hydrus-app-collection-creators']
+  end
+
+  context 'when successful deposit' do
+    let(:collection_attrs) { attributes_for(:collection) }
+
+    it 'deposits and renders work show page' do
+      visit dashboard_path
+
+      click_link '+ Create a new collection'
+
+      fill_in 'Collection name', with: collection_attrs.fetch(:name)
+      fill_in 'Description', with: collection_attrs.fetch(:description)
+      fill_in 'Contact email', with: collection_attrs.fetch(:contact_email)
+
+      click_button 'Deposit'
+
+      expect(page).to have_content(collection_attrs.fetch(:name))
+      expect(page).not_to have_content('+ Deposit to this collection')
+    end
+  end
+end
--- a/spec/features/create_new_work_spec.rb
+++ b/spec/features/create_new_work_spec.rb
@@ -3,8 +3,9 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Create a new collection and deposit to it', js: true do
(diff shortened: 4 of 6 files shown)


app/components/dashboard/collection_component.html.erb
```diff
@@ -43,6 +43,7 @@
     <div class="mb-3">See all deposits</div>
   <% end %>
 
+  <% unless deposit_blocked? %>
   <%= button_tag '+ Deposit to this collection',
```
mjgiarlo, line 47: Can you indent the lines wrapped by the `unless` check?

app/components/dashboard/collection_component.rb
```diff
@@ -23,5 +25,10 @@ def deposits
       policy = WorkPolicy.new(user: current_user, user_with_groups: user_with_groups)
       policy.authorized_scope(collection.works.limit(5), as: :depositor)
     end
+
+    sig { returns(T::Boolean) }
+    def deposit_blocked?
+      NO_DEPOSIT_STATUS.include? collection.state
```
mjgiarlo, line 31: Two questions:

1. Are you open to switching the logic here such that we could replace the `unless` in the component ERB with an `if`?
1. Would you consider moving this determination into the `Collection` model? So the code here, together with :point_up:, could look something like:

```ruby
def deposit_allowed?
  collection.allows_deposit?
end
```

That way the collection can own the logic governing whether deposits are allowed or not and the component doesn't need knowledge of the collection's internal state.

spec/factories/collections.rb
```diff
@@ -40,4 +41,12 @@
...
     depositors { build_list(:user, depositor_count) }
   end
+
+  trait :depositing_state do
+    state { 'depositing' }
+  end
+
+  trait :first_draft_state do
+    state { 'first_draft' }
+  end
```
mjgiarlo, line 51: Not a big deal, but the Work factory defines related traits without the `_state` suffix on them. They don't need to evolve in lockstep, of course.

spec/features/create_new_collection_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe 'Create a new collection', js: true do
+  let(:user) { create(:user) }
+
+  before do
+    sign_in user, groups: ['dlss:hydrus-app-collection-creators']
+    allow(Settings).to receive(:allow_sdr_content_changes).and_return(true)
```
mjgiarlo, line 11: I suspect this is not needed here as it's the default value and this test isn't testing the `false` case.

spec/features/create_new_collection_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  context 'when successful deposit' do
```
mjgiarlo, line 14: Thanks for extracting this!

spec/features/create_new_work_spec.rb
```diff
@@ -3,8 +3,9 @@
 
 require 'rails_helper'
 
-RSpec.describe 'Create a new collection and deposit to it', js: true do
+RSpec.describe 'Create a new work in a deposited collection', js: true do
```
mjgiarlo, line 6: Good change. :clap:

spec/features/create_new_work_spec.rb
```diff
@@ -17,14 +18,6 @@
...
 
-      click_link '+ Create a new collection'
-
-      fill_in 'Collection name', with: collection_attrs.fetch(:name)
-      fill_in 'Description', with: collection_attrs.fetch(:description)
-      fill_in 'Contact email', with: collection_attrs.fetch(:contact_email)
-
-      click_button 'Deposit'
-
       click_button '+ Deposit to this collection' # , match: :first
```
mjgiarlo, line 21: And we know this is depositing into the collection on line 8 because it's the only collection that exists at this point, I guess?

aaron-collier: Correct.

app/components/dashboard/collection_component.rb
jcoyne: Seems like this should be in the collection policy maybe?

aaron-collier: Good suggestion, on it.

mjgiarlo: I dig that.

mjgiarlo (commented):
Note to reviewers: please squash-merge this when it's approved.

app/components/dashboard/collection_component.rb
jcoyne: On reflection, it looks like the work policy would be the place.  I'd do `allows? :create?, Work.new(collection: collection)`  This is what we already do in WorkController#create

app/components/dashboard/collection_component.rb
```diff
@@ -27,8 +25,8 @@ def deposits
     end
 
     sig { returns(T::Boolean) }
-    def deposit_blocked?
-      NO_DEPOSIT_STATUS.include? collection.state
+    def deposit_allowed?
```
jcoyne, line 28: Rather than doing this, can we just modify the query in WorkPolicy so it excludes these collections from appearing?

jcoyne (changes requested):

app/components/dashboard/collection_component.rb
aaron-collier: Looking. Thanks for the suggestion.

jcoyne (approved):

PR #559, opened 2020-11-23
Only attempt to parse errors when the server gives us errors

jcoyne:
## Why was this change made?

Previously it was trying to parse an unparsable response when the server returned a 500.

### DIFF
--- a/app/javascript/controllers/edit_deposit_controller.js
+++ b/app/javascript/controllers/edit_deposit_controller.js
@@ -28,7 +28,18 @@ export default class extends Controller {
   }
 
   displayErrors(event) {
-    const [data, _status, _xhr] = event.detail;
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/edit_deposit_controller.js
```diff
@@ -37,5 +48,6 @@ export default class extends Controller {
       else
         console.error(`unable to find target for ${key}`)
     }
+    window.scrollTo(0, 80)
```
mjgiarlo, line 51: @jcoyne what's this doing?

jcoyne: Scrolls back to the top when errors are present.

mjgiarlo (approved):

PR #561, opened 2020-11-23
Bump Ruby version to 2.7.2

mjgiarlo:
## Why was this change made?

To keep pace.

CI

README


.ruby-version
```diff
@@ -1 +1 @@
-2.7.1
+2.7.2
```
mjgiarlo, line 1: @jcoyne you added this file in the initial commit. Do we want .ruby-version committed or nah? I can amend the PR if not.

jcoyne: I think rails generates this by default. But I think it'll just give us headaches when it doesn't match what we have on the server. I'd ditch it.

jcoyne (approved):

.ruby-version
mjgiarlo: Removing

PR #583, opened 2020-11-24
When editing, show the previously selected roles on the form

### LINKED ISSUE #571: Organization authors need to display for deposited items
Not all authors are loading into the form properly when editing a deposited item. 

This already deposited item (in a collection with the review workflow on) had an organizational author entered, but when it was reloaded to create a new version it looked like this:

When user selected "Organization - author" the entry showed up in the box.

jcoyne:
## Why was this change made?
Fixes #571


mjgiarlo (approved):

mjgiarlo:
@jcoyne Approved. Rebase?

PR #606, opened 2020-11-25
Move EventService methods into to the state machine

jcoyne:
## Why was this change made?

To complete the merger of EventService into the state machine.


ndushay (approved):
Mike and I reviewed this pairing on zoom.  He says "merge ok" and I am happy to be led;  I saw nothing against.

Issue #576, opened 2020-11-24
Item returned for corrections displays wrong status in table


amyehodge:
This item was submitted by the depositor for review, was reviewed and returned to the depositor for changes.

![Screen Shot 2020-11-23 at 5.57.21 PM.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/07cd82a1-b03c-4421-8fad-d3a144d8fff8)

The Deposit Status in the table should say "Returned."

jcoyne:
@amyehodge do we need to differentiate if this is "returned draft" or "returned version"?

amyehodge:
@jcoyne No. In both cases this should just be "Returned."

mjgiarlo:
Connects to #581

ndushay:
blocked until PR #606 is merged

mjgiarlo:
This is no longer blocked.

PR #607, opened 2020-11-25
Allow version drafts to be submitted for review

### LINKED ISSUE #516: Ensure review workflow (mediated deposits) is compatible with versioning
Ensure that when an object is versioned, that on the next version it continues to go through the review workflow.

mjgiarlo:
Fixes #516

## Why was this change made?

See title.

CI

None


app/models/work.rb
```diff
@@ -34,19 +34,20 @@ class Work < ApplicationRecord
     end
 
     event :begin_deposit do
-      transition first_draft: :depositing, version_draft: :depositing, pending_approval: :depositing
+      transition %i[first_draft version_draft pending_approval] => :depositing
```
mjgiarlo, line 37: This is pure refactoring, using an array to indicate the before states all transition to the after state.

app/models/work.rb
```diff
@@ -34,19 +34,20 @@ class Work < ApplicationRecord
...
     end
 
     event :deposit_complete do
       transition depositing: :deposited
     end
 
     event :submit_for_review do
-      transition first_draft: :pending_approval
+      transition %i[first_draft version_draft] => :pending_approval
```
mjgiarlo, line 45: This is the bug fix.

app/models/work.rb
```diff
@@ -34,19 +34,20 @@ class Work < ApplicationRecord
...
     end
 
     event :reject do
-      transition pending_approval: :first_draft
+      transition pending_approval: :first_draft, if: ->(work) { work.druid.blank? }
```
mjgiarlo, line 49: This is the next bug we'd hit: having a version_draft be submitted for review and then returning to first_draft.

app/services/event_service.rb
```diff
@@ -9,21 +9,21 @@ class << self
     include ActionView::Helpers::UrlHelper
   end
 
-  sig { params(work: Work, user: User, description: String).returns(T.nilable(Integer)) }
+  sig { params(work: Work, user: User, description: String).returns(T.untyped) }
```
mjgiarlo, line 12: The return value is either an integer or an array of procs, depending on what the ActionCable broadcast does. We don't care about the return value, though, so let Sorbet ignore it.

app/views/dashboards/show.html.erb
```diff
@@ -10,7 +10,7 @@
     </section>
 
     <% if @presenter.approvals.any? %>
-      <section class="mb-5">
+      <section class="mb-5" id="approvals">
```
mjgiarlo, line 13: Makes for a more targeted feature spec.

spec/features/mediated_deposit_spec.rb
```diff
@@ -0,0 +1,52 @@
+# typed: false
```
mjgiarlo, line 1: Since I was already writing a mediated deposit feature spec to test the versioning/state bugfix, I thought I'd toss a basic mediated deposit feature spec in too.

spec/features/mediated_deposit_versioning_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(page).to have_content(new_work_title)
+      expect(page).to have_content('Pending approval - Not deposited')
+
+      visit dashboard_path
+      within('section#approvals') do
+        click_link(new_work_title)
+      end
+      expect(page).to have_content('Review all details below, then approve or return this deposit')
+      find('label', text: 'Approve and deposit').click
+      click_button('Submit')
```
mjgiarlo, line 41: Prior to the bugfix, this would fail.

app/models/work.rb
jcoyne: Do you think this should go to a returned state? See https://github.com/sul-dlss/happy-heron/issues/576

mjgiarlo: I'll take a look.

mjgiarlo: @jcoyne can I defer this change to #576, and pick that up as my next ticket?

sorbet/rbi/gems/actionpack.rbi
```diff
@@ -1765,6 +1765,10 @@ class ActionDispatch::RemoteIp::GetIp
   def ips_from(header); end
   def to_s; end
 end
+module ActionController::ApiRendering
```
mjgiarlo, line 1768: I believe all the sorbet stuff was generated because I rebased on `main` with dep updates. I think it's safe to ignore.

mjgiarlo:
@jcoyne can you take another look? I think this is good to go. Then I will work on #576

app/models/work.rb
jcoyne: Sure

spec/features/mediated_deposit_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      fill_in 'Keywords', with: 'Springs'
+      blur_from 'work_keywords'
+      check 'I agree to the SDR Terms of Deposit'
+      click_button 'Deposit'
+
+      expect(page).to have_content(work_title)
+      expect(page).to have_content('Pending approval - Not deposited')
+
+      visit dashboard_path
+      within('section#approvals') do
```
jcoyne, line 37: another option you have available here is `within_table('Approvals') do`

app/views/dashboards/show.html.erb
jcoyne: consider https://github.com/sul-dlss/happy-heron/pull/607/files/ea86cbb9bf8bb5743dc89bea77341db965f78ab8#diff-171048dc70a2f5431a6ecc157b9b17eb786dc5bd1710c795c2c69551d7ac9a63 instead.

mjgiarlo: Nice, I dig it.

spec/features/mediated_deposit_spec.rb
mjgiarlo: @jcoyne done, thanks

jcoyne (approved):

PR #630, opened 2020-12-02
Display reason and status for Returned reviews

### LINKED ISSUE #576: Item returned for corrections displays wrong status in table

### LINKED ISSUE #581: Comments from Reviewer should be displayed at top of item edit page
In testing the review workflow, an item was returned for editing by the collection manager along with a comment.

When the depositor opened that item to make edits, the comment was not displayed on the page. https://sul-h2-qa.stanford.edu/works/38/edit.

ndushay:
Fixes #576
Fixes #581

## Why was this change made?

For #576:  display appropriate status for work rejected by reviewer

### After

image

For #581:  display comments from reviewer if work was rejected

### Before

image

### After

image

 
Amy reviewed the changes for #581 in stage;   tests added for both issues with their commits.

### DIFF
--- a/app/components/dashboard/collection_component.html.erb
+++ b/app/components/dashboard/collection_component.html.erb
@@ -29,7 +29,7 @@
               <% end %>
             <% end %>
           </td>
-          <td><%= work.state.humanize %></td>
+          <td><%= render Works::StateDisplayComponent.new(work: work) %></td>
           <td><%= I18n.l(work.updated_at.to_date, format: :abbr_month) %></td>
           <td><%= link_to work.purl, work.purl if work.purl %></td>
           <td><span class="fas fa-quote-left"></span> Cite</td>
--- a/app/components/works/state_display_component.rb
+++ b/app/components/works/state_display_component.rb
@@ -9,6 +9,7 @@ module Works
         'first_draft' => 'Draft - Not deposited',
         'version_draft' => 'New version draft - Not deposited',
         'pending_approval' => 'Pending approval - Not deposited',
+        'rejected' => 'Returned',
         'depositing' => 'Deposit in progress <span class="fas fa-spinner fa-pulse"></span>'.html_safe,
         'deposited' => 'Deposited'
       }.freeze,
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -48,6 +48,9 @@ class Work < ApplicationRecord
       BroadcastWorkChange.call(work: work, state: transition.to_name)
     end
 
+    # NOTE: there is no approval "event" because when a work is approved in review, it goes
+    # directly to begin_deposit event, which will transition it to depositing
+
     event :begin_deposit do
       transition %i[first_draft version_draft pending_approval] => :depositing
     end
@@ -57,12 +60,11 @@ class Work < ApplicationRecord
     end
 
     event :submit_for_review do
-      transition %i[first_draft version_draft] => :pending_approval
+      transition %i[first_draft version_draft rejected] => :pending_approval
     end
 
     event :reject do
-      transition pending_approval: :first_draft, if: ->(work) { work.druid.blank? }
-      transition pending_approval: :version_draft, if: ->(work) { work.druid.present? }
+      transition pending_approval: :rejected
     end
 
     event :update_metadata do
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -69,7 +69,7 @@ class Work < ApplicationRecord
 
     event :update_metadata do
       transition deposited: :version_draft
-      transition %i[first_draft version_draft] => same
+      transition %i[first_draft version_draft rejected] => same
     end
   end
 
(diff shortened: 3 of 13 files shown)


app/components/dashboard/collection_component.html.erb
```diff
@@ -29,7 +29,7 @@
               <% end %>
             <% end %>
           </td>
-          <td><%= work.state.humanize %></td>
+          <td><%= render Works::StateDisplayComponent.new(work: work) %></td>
```
ndushay, line 32: This is part of @mjgiarlo's efforts to normalize how we display state info.  He will have a PR forthcoming after _this_ PR is merged to finish this work.

app/components/works/state_display_component.rb
```diff
@@ -9,6 +9,7 @@ class StateDisplayComponent < ApplicationComponent
         'first_draft' => 'Draft - Not deposited',
         'version_draft' => 'New version draft - Not deposited',
         'pending_approval' => 'Pending approval - Not deposited',
```
ndushay, line 11: @amyehodge question as to whether you want this full string on the right to be displaying or not -- @mjgiarlo is hoping to normalize how we display state info.

app/models/work.rb
```diff
@@ -57,17 +60,16 @@ def event_context
...
-      transition pending_approval: :first_draft, if: ->(work) { work.druid.blank? }
-      transition pending_approval: :version_draft, if: ->(work) { work.druid.present? }
+      transition pending_approval: :rejected
     end
 
     event :update_metadata do
       transition deposited: :version_draft
-      transition %i[first_draft version_draft] => same
+      transition %i[first_draft version_draft rejected] => same
     end
```
ndushay, line 73: note that I specifically asked Amy to ensure that the rejection reason displayed at the appropriate times and that everything else seemed to work ok ...

app/views/dashboards/show.html.erb
```diff
@@ -29,7 +29,7 @@
               <td><%= link_to submission.title, work_path(submission) %></td>
               <td><%= submission.depositor.sunetid %></td>
               <td><%= submission.collection.name %></td>
-              <td><%= submission.state.humanize %></td>
+              <td><%= render Works::StateDisplayComponent.new(work: submission) %></td>
```
ndushay, line 32: This is part of @mjgiarlo's efforts to normalize how we display state info.  He will have a PR forthcoming after _this_ PR is merged to finish this work.

config/locales/en.yml
```diff
@@ -9,6 +9,12 @@ en:
     type:
       return: Changes requested
       begin_deposit: Deposited
+      rejected: Changes requested
+      pending_approval: Waiting for review
+      first_draft: First draft
+      version_draft: Version draft
+      depositing: Deposited
```
jcoyne, line 16: Do we need both `deposited` and `depositing`?  Do we need `rejected` and `return`?

ndushay: I wish we didn't need them all but my testing found we did.  Perhaps the forthcoming "normalization of event/state display code text" PR will address?   Perhaps wiping the databases will address?  Dunno.  I didn't do the coding that required these -- I was only a consumer stymied by their absence.

jcoyne: I pulled this branch down and removed `return` and `depositing` and there were no test failures.  Perhaps this code needs coverage?

mjgiarlo: @jcoyne @ndushay I would like to issue a follow-up PR that tweaks the events attached to state transitions and how we label them. Maybe I could take this on?

spec/components/works/detail_component_spec.rb
```diff
@@ -14,7 +14,7 @@
     end
   end
 
-  context 'when deposted' do
+  context 'when deposited' do
```
mjgiarlo, line 17: good eye

mjgiarlo (approved):
LGTM, pending my comment about follow-up work.

config/locales/en.yml
ndushay: FYI, I think "depositing" is what you get locally when you hit deposit and the thing spins forever unless you manually give it a druid;  having an extra translation for that isn't a travesty, IMO.

app/components/works/state_display_component.rb
amyehodge: I would say the "Not deposited" portion of the text does not need to display. And if I think about it, I wonder if that might make some people think that the "New version draft" somehow caused the original to be un-deposited? (not a thing, but they get very confused about the workflow sometimes, which is understandable).

mjgiarlo: This will be changing per post-stand-up discussion. Stay tuned.

mjgiarlo: @amyehodge for each of the six states here, could you write up the label you'd like to see in the app in a new issue, pls? It won't be a huge change, and it'd be good to get these right. Thanks!

PR #636, opened 2020-12-02
Use state machines graphviz library to generate diagrams

### LINKED ISSUE #629: Update state machine diagrams
@jcoyne Can I ask you to update the state machine diagrams you created for the inception deck? We've evolved the state machine a bit in code since the original design, and I'm hearing that it'd be valuable to keep these diagrams up to date so we have a visual way to make sense of our state machine flow. Once updated, can you link them in the [(new) dev wiki](https://github.com/sul-dlss/happy-heron/wiki)?

mjgiarlo:
Fixes #629

## Why was this change made?

These state machine images, plus instructions on how/when to regenerate them, will live in our dev docs (wiki): https://github.com/sul-dlss/happy-heron/wiki#state-machine-diagrams. They help us make better sense of our state transitions.

:man_shrugging: 

See above


app/services/broadcast_work_change.rb
```diff
@@ -1,4 +1,4 @@
-# typed: true
+# typed: strict
```
mjgiarlo, line 1: Sorbet flipped this for me on push.

jcoyne (changes requested):
How about generating svgs rather than PNGs? That would make them more useful at every resolution.

ndushay (approved):
Thank you for this!

ndushay (commented):
why does Collection have a transition "begin deposit"  from deposited to depositing without it going to version draft???

image

mjgiarlo:
@ndushay :speech_balloon: 
> why does Collection have a transition "begin deposit" from deposited to depositing without it going to version draft???
> 
> image

I don't know atm. Maybe the state machine diagrams are already proving worthwhile :laughing:

jcoyne (approved):

PR #641, opened 2020-12-02
blank work titles show render on dashboard in all places

peetucket:
## Why was this change made?

I noticed that the dashboard was not showing a link for draft works with no titles in the collection view area.  This fixes the problem across the dashboard where we create links to work without code duplication.

Before:

Screen Shot 2020-12-02 at 2 10 04 PM

After:

Screen Shot 2020-12-02 at 2 12 22 PM

In the browser

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -74,11 +74,6 @@ class Work < ApplicationRecord
   end
 
   sig { returns(T.nilable(String)) }
-  def name
-    title.presence || 'No title'
-  end
-
-  sig { returns(T.nilable(String)) }
   def purl
     return nil unless druid
 
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -74,6 +74,11 @@ class Work < ApplicationRecord
   end
 
   sig { returns(T.nilable(String)) }
+  def name
+    title.presence || 'No title'
+  end
+
+  sig { returns(T.nilable(String)) }
   def purl
     return nil unless druid
 
--- a/sorbet/rails-rbi/models/collection/habtm_managers.rbi
+++ b/sorbet/rails-rbi/models/collection/habtm_managers.rbi
@@ -7,6 +7,26 @@ module Collection::HABTM_Managers::ActiveRelation_WhereNot
   def not(opts, *rest); end
 end
 
+module Collection::HABTM_Managers::GeneratedAttributeMethods
+  sig { returns(Integer) }
+  def collection_id; end
+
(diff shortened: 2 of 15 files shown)


jermnelson (changes requested):
Can you add a test for this change?

app/models/work.rb
```diff
@@ -56,6 +56,11 @@ class Work < ApplicationRecord
     end
   end
 
+  sig { returns(T.nilable(String)) }
+  def name
```
jcoyne, line 60: I think this is a confusing method name. name/title are easily confused.  How about `title_or_default`?

jcoyne (changes requested):

app/models/work.rb
peetucket: Yeah, I was originally going to use "display_title" but wanted to be consistent with something similar being done for collections here: https://github.com/sul-dlss/happy-heron/blob/main/app/components/dashboard/collection_component.rb#L16-L19

We may want to consider renaming that method to be consistent if we change it for works here.

jcoyne: `display_title` would be okay too. If you want to be consistent, then putting it in the component (rather than the model) would be a good idea.  Then you could call it `title` as a component method wouldn't be expected to return exactly what a model method does.

peetucket: I just removed it from the model and added it as "title" to the `Works::DetailComponent`, which I then used on the Dashboard.  This also automatically fixes one more spot that I missed on the Work show page.

sorbet/rails-rbi/models/collection/habtm_managers.rbi
```diff
@@ -77,7 +57,6 @@ module Collection::HABTM_Managers::CustomFinderMethods
 end
 
 class Collection::HABTM_Managers < ActiveRecord::Base
-  include Collection::HABTM_Managers::GeneratedAttributeMethods
```
jcoyne, line 80: Did you run `bundle exec srb rbi hidden-definitions`?   I wonder why this disappeared?

peetucket: Nope.  Same comment: https://github.com/sul-dlss/happy-heron/pull/644#discussion_r535454919

I did do a rebase from  master (and on #644 too)

jcoyne (approved):

jcoyne (approved):

peetucket:
> Can you add a test for this change?

Two tests added: one for the dashboard view, and for the show work page view.

jermnelson (approved):

PR #642, opened 2020-12-02
Add collection show page component

### LINKED ISSUE #37: Collection view page
https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/426926002

aaron-collier:
## Why was this change made?

Fixes #37 

Adds the collection show page component.

Makes the collection name a link to the show page:

Screen Shot 2020-12-02 at 4 52 21 PM

The show page (events/history omitted):

Screen Shot 2020-12-02 at 3 43 59 PM

Deposits tab:

Screen Shot 2020-12-02 at 3 44 11 PM

New unit tests

N/A


app/components/works/work_type_modal_component.html.erb
```diff
@@ -11,7 +11,7 @@
   <button type="button" class="btn-close" data-dismiss="modal" aria-label="Close"></button>
 </div>
 <div class="modal-body">
-  <form method="get" action="/collections/1/work/new" data-target="work-type.form">
+  <form method="get" action="<%= collection_path %>" data-target="work-type.form">
```
jcoyne, line 14: I'm not sure I understand why this needed to change.

app/components/works/work_type_modal_component.rb
```diff
@@ -4,8 +4,18 @@
...
   class WorkTypeModalComponent < ApplicationComponent
+    def initialize(collection: nil)
+      @collection = collection
+    end
+
     def types
       WorkType.all
     end
+
+    def collection_path
```
jcoyne, line 15: I don't understand the need for these changes.

app/components/collections/information_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    sig { returns(String) }
+    def last_saved
+      format_date_string(collection.updated_at)
+    end
+
+    sig { returns(T.nilable(String)) }
+    def purl
+      return if collection.druid.nil?
+
+      "https://purl.stanford.edu/#{collection.druid.gsub(/druid:/, '')}"
```
jcoyne, line 32: Use a model method similar to `Work#purl`. It should use `Settings.purl_url`

app/controllers/collections_controller.rb
```diff
@@ -5,7 +5,7 @@
 class CollectionsController < ObjectsController
   before_action :authenticate_user!
   before_action :ensure_sdr_updatable
-  verify_authorized
+  verify_authorized except: [:show]
```
jcoyne, line 8: Why are we not authorizing the show method?

app/controllers/collections_controller.rb
```diff
@@ -49,6 +49,10 @@ def update
     end
   end
 
+  def show
+    @collection = Collection.find(params[:id])
```
jcoyne, line 53: This appears to be untested.

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        <%= render Collections::ParticipantsComponent.new(collection: @collection) %>
+        <%= render Collections::WorkflowReviewComponent.new(collection: @collection) %>
+      </div>
+
+      <div id="deposits" class="tab-pane fade">
+        <%= render Dashboard::CollectionComponent.new(collection: @collection) %>
+      </div>
+    </div>
+    <section data-controller="work-type">
+    <div class="modal fade" id="workTypeModal" tabindex="-1" aria-labelledby="content-type-prompt" aria-hidden="true">
```
jcoyne, line 35: It seems like this should be indented one more step.

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,45 @@
...
+        <%= render Collections::TermsOfUseComponent.new(collection: @collection) %>
+        <%= render Collections::ParticipantsComponent.new(collection: @collection) %>
+        <%= render Collections::WorkflowReviewComponent.new(collection: @collection) %>
+      </div>
+
+      <div id="deposits" class="tab-pane fade">
+        <%= render Dashboard::CollectionComponent.new(collection: @collection) %>
+      </div>
+    </div>
+    <section data-controller="work-type">
```
jcoyne, line 34: why do we need this section.  This seems like it would be confusing to a screen-reader.

config/settings.yml
```diff
@@ -31,3 +31,5 @@ authorization_workgroup_names:
 file_uploads_root: tmp/uploads
 
 purl_url: https://purl.stanford.edu
+
+terms_of_use: User agrees that, where applicable, content will not be used to identify or to otherwise infringe the privacy or confidentiality rights of individuals. Content distributed via the Stanford Digital Repository may be subject to additional license and use restrictions applied by the depositor.
```
jcoyne, line 35: Should this go in `config/locales/en.yml`?  This seems more like text rather than a setting.

app/components/collections/information_component.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    private
+
+    def format_date_string(date)
+      date.strftime('%B %e, %Y %I:%M%p')
```
jcoyne, line 38: This should use `I18n.l(date, format: :abbr_month)`

app/views/collections/show.html.erb
```diff
@@ -0,0 +1,45 @@
+<div class="container" id="collection" data-controller="collection-updates" dat-collection-updates-collection-id="<%= @collection.id %>">
+  <%= render 'nav' %>
+  <section class="details">
+    <header class="title">
+      <span class="header-text"><%= @collection.name %></span>
+      <span class="deposit-button">
+        <% if allowed_to?(:deposit?, @collection) %>
```
jcoyne, line 7: This is not a rule we have.  It should use `allowed_to?(:create?, Work.new(collection: @collection))`

spec/components/collections/participants_component_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+RSpec.describe Collections::ParticipantsComponent, type: :component do
+  let(:rendered) { render_inline(described_class.new(collection: collection)) }
+
+  context 'when displaying a collection' do
+    let(:depositors) { collection.depositors.pluck(:email).join(', ') }
+    let(:managers) { collection.managers }
+    let(:collection) { build_stubbed(:collection, :with_depositors) }
+
+    it 'renders the participant component' do
+      expect(rendered.css('table').to_html).to include(managers)
```
jcoyne, line 15: I'm not sure this is up to date. Collection#managers returns an association. So it's odd we'd see that in the html.  Perhaps do like you did with depositors.

spec/components/collections/workflow_review_component_spec.rb
```diff
@@ -0,0 +1,19 @@
+# typed: false
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Collections::WorkflowReviewComponent, type: :component do
+  let(:rendered) { render_inline(described_class.new(collection: collection)) }
+
+  context 'when displaying a collection' do
+    let(:review_enabled) { 'On' }
```
jcoyne, line 10: I'm not sure why this is in a let block.  It's only used the one time and I think having to reference up to the let blocks makes the test hard to read.

jcoyne (changes requested):

app/components/works/work_type_modal_component.html.erb
aaron-collier: Without changing this here, I was unable to get correct action is not set when reusing this model for the collection show. So the button always goes to `/collections/1/work/new` instead of the actual path for a new work in a collection. During pairing with @mjgiarlo this was our first pass at a possible solution.

Any idea why it wouldn't work as is in this context?

app/components/collections/information_component.rb
aaron-collier: Thanks! One question I had at the end of the day was going to be about locale for this and the terms of use. Thanks :-)

spec/components/collections/participants_component_spec.rb
aaron-collier: You're right, that changed before me. Fixing. Thanks.

app/views/collections/show.html.erb
aaron-collier: Without this, the work type modal displays without the type selection area:

without:

Screen Shot 2020-12-03 at 7 15 31 AM

With:

Screen Shot 2020-12-03 at 7 15 13 AM

app/controllers/collections_controller.rb
aaron-collier: This matches what we have in the WorksController... Without this I'm getting an unauthorized error.

Would our preference be to add a policy? Though I'm not seeing one for works either.

app/controllers/collections_controller.rb
aaron-collier: I added a request spec, this should be covered now.

app/components/works/work_type_modal_component.html.erb
aaron-collier: This is also the source of the type error. So perhaps there's a better way to do this.

jcoyne: I just think this is unnecessary, as https://github.com/sul-dlss/happy-heron/blob/main/app/javascript/controllers/work_type_controller.js#L9 takes care of overwriting this. I think this change is a no-op, but it adds extra code which we don't need to support, so I think it should be reverted.

app/views/collections/show.html.erb
jcoyne: I'm not saying to remove the ` data-controller="work-type"`, but that it shouldn't be a section.  Does it work if you move `data-controller` to the `workTypeModal` div or up to the parent `<section>`

spec/components/collections/information_component_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    let(:last_saved) { I18n.l(collection.updated_at, format: :abbr_month) }
+    let(:version) { collection.version.to_s }
+    let(:creator) { collection.creator.to_s }
+    let(:collection) { build_stubbed(:collection) }
+
+    it 'renders the information component' do
+      expect(rendered.css('table').to_html).to include(version)
+      expect(rendered.css('table').to_html).to include(creator)
+      expect(rendered.css('table').to_html).to include(created)
+      expect(rendered.css('table').to_html).to include(last_saved)
```
jcoyne, line 20: I'm not sure these are good tests.  You test is running the same code as the code being tested is running. Effectively you are asking does `x = x`?

spec/components/collections/terms_of_use_component_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+RSpec.describe Collections::TermsOfUseComponent, type: :component do
+  let(:rendered) { render_inline(described_class.new(collection: collection)) }
+
+  context 'when displaying a collection' do
+    let(:terms_of_use) { I18n.t(:terms_of_use) }
+    let(:license) { collection.default_license }
+    let(:collection) { build_stubbed(:collection) }
+
+    it 'renders the terms of use component' do
+      expect(rendered.css('table').to_html).to include(terms_of_use)
```
jcoyne, line 15: This also is an `x = x` test.

app/controllers/collections_controller.rb
jcoyne: Yes, we should have policies.  We do not want these show pages to be accessible to the general population.

jcoyne (changes requested):
Add a Collection show policy.

jcoyne (changes requested):
Can you add policy tests similar to what were added in https://github.com/sul-dlss/happy-heron/pull/649

app/models/collection.rb
```diff
@@ -21,6 +21,13 @@ def accessioned?
     %w[first_draft depositing].exclude?(state)
   end
 
+  sig { returns(T.nilable(String)) }
+  def purl
+    return nil unless druid
+
+    File.join(Settings.purl_url, T.must(druid).delete_prefix('druid:'))
```
jcoyne, line 28: Can you either add a test for this, or move this method to a module that Collection and Work share? I think the work already has a test for it.

aaron-collier: Done.

jcoyne (approved):

PR #644, opened 2020-12-02
add ability to delete draft works and collections

### LINKED ISSUE #469: Delete draft works and collections
On the dashboard I should be able to delete draft deposits on the dashboard and on the deposit form (for works and collections).

See https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/433047903

peetucket:
## Why was this change made?

Fixes #469 - allow the user to delete a draft collection or work.

~~Notes:~~
~~- I set this up so only draft collections that do not have any works associated with them can be deleted to prevent the user from mistakenly deleting a bunch of works too.~~

~~To Do:~~
~~- Tests~~

~~Does the `verify_authorized` (https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/collections_controller.rb#L8) and `authorize! collection` in the methods themselves use the policy methods qw have created to ensure we only delete a work or collection that is in draft?~~

- In browser
- New tests verifying draft works and collections have delete buttons and can be deleted
- Update existing tests for deposited works and collections to verify the delete button is not shown in those cases


app/policies/work_policy.rb
```diff
@@ -30,5 +31,10 @@ def review?
     record.pending_approval? && (administrator? || record.collection.reviewers.include?(user))
   end
 
+  sig { returns(T::Boolean) }
+  def destroy?
+    (administrator? || record.depositor == user) && record.state == 'first_draft'
```
jcoyne, line 36:
```suggestion
    (administrator? || record.depositor == user) && record.first_draft?
```

jcoyne (changes requested):
Looks good so far. Please add a couple of tests.

app/policies/collection_policy.rb
```diff
@@ -32,5 +33,10 @@ def deposit?
     NO_DEPOSIT_STATUS.exclude? record.state
   end
 
+  sig { returns(T::Boolean) }
+  def destroy?
+    (administrator? || collection_creator?) && record.state == 'first_draft' && record.works.count.zero?
```
jcoyne, line 38: A draft collection will not have any works, so no need to check for that.
```suggestion
    (administrator? || collection_creator?) && record.first_draft?
```

jcoyne (changes requested):

app/policies/collection_policy.rb
peetucket: huh, interesting -- I had some draft works in a draft collection on my localhost from previous testing, which is the reason I added this condition.  is that not true?  I mean, we can allow the draft works inside of the draft collection to be destroyed via `has_many :works, dependent: :destroy`, but I just thought it would be safer to not allow that in the UI

app/controllers/collections_controller.rb
```diff
@@ -49,6 +49,15 @@ def update
     end
   end
 
+  def destroy
+    collection = Collection.find(params[:id])
+    authorize! collection
```
peetucket, line 54: @jcoyne does this validate via the policy (to ensure the collection is a draft)?  and same question for the work

app/policies/collection_policy.rb
jcoyne: We have added restrictions that prevents works from being created in draft colletions recently: https://github.com/sul-dlss/happy-heron/blob/main/app/policies/work_policy.rb#L19

peetucket: got it, thanks

app/controllers/collections_controller.rb
jcoyne: Yes, it's using the `CollectionPolicy#destroy?` method you wrote which checks `first_draft?`

sorbet/rails-rbi/models/user/habtm_manages_collections.rbi
```diff
@@ -77,7 +57,6 @@ module User::HABTM_ManagesCollections::CustomFinderMethods
 end
 
 class User::HABTM_ManagesCollections < ActiveRecord::Base
-  include User::HABTM_ManagesCollections::GeneratedAttributeMethods
```
jcoyne, line 80: Did you run `bundle exec srb rbi hidden-definitions`?

peetucket: I did not - but I do have the pre-push hook, which does not include this.

peetucket:
> Looks good so far. Please add a couple of tests.

Tests added and a couple existing tests updated.

app/components/dashboard/collection_component.html.erb
```diff
@@ -5,6 +5,12 @@
         <span class="fas fa-pencil-alt"></span>
       <% end %>
     <% end %>
+
+    <% if allowed_to?(:delete?, collection) %>
+      <%= link_to collection_path(collection), method: :delete, id: "remove-collection-#{collection.id}", aria: { label: "Delete #{name}"}, data: {confirm: "Are you sure you want to delete this draft collection? It cannot be undone."} do %>
```
jcoyne, line 10: why does this need an id? Would you mind adding spaces around the brackets in the hash?

app/components/dashboard/collection_component.html.erb
```diff
@@ -28,6 +34,12 @@
                 <span class="fas fa-pencil-alt"></span>
               <% end %>
             <% end %>
+
+            <% if allowed_to?(:delete?, work) %>
+              <%= link_to work_path(work), method: :delete, aria: { label: "Delete #{work.title}"}, data: {confirm: "Are you sure you want to delete this draft work? It cannot be undone."} do %>
```
jcoyne, line 39: Please add spaces around the hash brackets.

app/views/dashboards/show.html.erb
```diff
@@ -5,6 +5,11 @@
       <% @presenter.drafts.each do |draft| %>
         <div>
           <%= Dashboard::CollectionComponent.new(collection: draft.collection).name %> &gt; <%= link_to Works::DetailComponent.new(work: draft).title, edit_work_path(draft) %>
+          <% if allowed_to?(:delete?, draft) %>
```
jcoyne, line 8: Could this section be it's own component and reused for views/dashboards/show.html.erb and in components/dashboard/collection_component.html.erb

spec/features/create_new_collection_spec.rb
```diff
@@ -30,6 +30,9 @@
       expect(page).to have_content(collection_attrs.fetch(:name))
       # The deposit button is not present until the collection is accessioned
       expect(page).not_to have_content('+ Deposit to this collection')
+
+      # We should not see the delete button for this collection since it is not a draft
+      expect(page).not_to have_selector("#remove-collection-#{Collection.last.id}")
```
jcoyne, line 35: Can you do this by querying on the aria-label rather than the id?  The aria-label is more important to the users.  The id serves no purpose that I can tell.

spec/features/create_new_work_spec.rb
```diff
@@ -130,6 +130,11 @@
       expect(page).to have_content('2020-03-06/2020-10-30')
       expect(page).to have_content('Whatever')
       expect(page).to have_content('CC-PDDC Public Domain Dedication and Certification')
+
+      visit dashboard_path
+
+      # We should not see the delete button for this work since it is not a draft
+      expect(page).not_to have_selector("#remove-work-#{Work.last.id}")
```
jcoyne, line 137: Can you do this by querying on the aria-label rather than the id? The aria-label is more important to the users. The id serves no purpose that I can tell.

spec/features/delete_draft_collection_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+
+      click_link '+ Create a new collection'
+
+      fill_in 'Collection name', with: collection_attrs.fetch(:name)
+      click_button 'Save as draft'
+
+      collection = Collection.last
+      expect(collection).to be_first_draft
+      accept_confirm do
+        find("#remove-collection-#{collection.id}").click
```
jcoyne, line 27: Can you do this by querying on the aria-label rather than the id?

spec/features/delete_draft_work_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  before do
+    sign_in user
+  end
+
+  context 'when draft' do
+    it 'allow users to delete the work and destroys the model' do
+      expect(work).to be_first_draft
+      visit dashboard_path
+      accept_confirm do
+        find("#remove-work-#{work.id}").click
```
jcoyne, line 19: Can you do this by querying on the aria-label rather than the id?

jcoyne (changes requested):

app/components/dashboard/collection_component.html.erb
mjgiarlo: Time to consider erb-lint?

spec/features/create_new_collection_spec.rb
peetucket: Yes, the id was added solely for the tests.  Do you know what the correct syntax to use for finding by aria-label?  i tried what was described here: https://github.com/teamcapybara/capybara/issues/1528 (setting `Capybara.enable_aria_label = true` and then `find("Delete #{work.title}").click`) but couldn't get it to work to find the delete link)

jcoyne: first i would try:
`click_link "Delete #{work.title}"` if that failed `find("[aria-label='Delete #{work.title}']").click`

app/components/dashboard/collection_component.html.erb
peetucket: fixed spacing, removed id in component as i got the aria-label finder to work in the test

spec/features/create_new_collection_spec.rb
peetucket: Well, waddya know, adding the capybara config and using click_link worked, not sure why I couldn't get that to work before (cause I did try it)

spec/features/delete_draft_collection_spec.rb
peetucket: yes, got it to work now

spec/features/create_new_work_spec.rb
peetucket: Done

jcoyne (approved):

PR #645, opened 2020-12-02
Add events to collections and orient event creation around events, not states

mjgiarlo:
## Why was this change made?

This commit follows a discussion in stand-up earlier today and aims to make the
following improvements around events:

* Create state machine transition events using the event as the type, not the resulting state
* Extract an `Eventable` model concern out of the Work and Collection model classes (a refactoring)
* Rename the `events.work_id` database column to `events.eventable`, using a polymorphic relationship allowing events to be associated with both works and collections
* Add event create to the collection state machine, for consistency with the work state machine, since we know we will need collection events soon
* Add the `Event#latest_by_type` scope to facilitate getting the latest event of a particular type by using database query mechanics
* Inject user information into event contexts in the work and collection controllers so that events are reflecting the accurate user associated with the event
* Regenerate sorbet and make it happy again

CI

None


app/services/description_generator.rb
```diff
@@ -88,7 +88,7 @@ def created_date
       type: 'creation',
       date: [
         {
-          value: date.respond_to?(:to_edtf) ? date.to_edtf : date.to_s,
+          value: date.respond_to?(:edtf) ? date.edtf : date.to_s,
```
mjgiarlo, line 91: Sorbet said so.

app/models/work.rb
```diff
@@ -98,18 +89,28 @@ def created_edtf=(edtf)
     end
   end
 
+  sig { returns(T.nilable(T.any(EDTF::Interval, Date))) }
```
mjgiarlo, line 92: Added to help bump from `true` to `strict` for Sorbet.

app/models/event.rb
```diff
@@ -3,6 +3,10 @@
 
 # Records events in the lifecycle of a deposit
 class Event < ApplicationRecord
-  belongs_to :work
+  belongs_to :eventable, polymorphic: true
```
mjgiarlo, line 6: Allow events to be attached to both works and collections.

config/locales/en.yml
```diff
@@ -7,14 +7,11 @@ en:
...
     type:
-      return: Changes requested
-      begin_deposit: Deposited
-      rejected: Changes requested
-      pending_approval: Waiting for review
-      first_draft: First draft
-      version_draft: Version draft
-      depositing: Deposited
-      deposited: Deposited
+      begin_deposit: Submitted for deposit
```
mjgiarlo, line 10: Only need events here now, not states.

spec/factories/collections.rb
```diff
@@ -13,7 +13,6 @@
     required_license { 'MyString' }
     default_license { 'MyString' }
     email_when_participants_changed { false }
-    state { 'deposited' }
```
mjgiarlo, line 16: Let it default to its default value, and use the object states traits in tests to vary the behavior, the same way we do with works.

spec/factories/events.rb
```diff
@@ -4,8 +4,8 @@
 FactoryBot.define do
   factory :event do
     description { 'MyString' }
-    event_type { 'MyString' }
-    work { nil }
+    event_type { 'update_metadata' }
```
mjgiarlo, line 7: Give it a valid event type since we validate them now.

spec/jobs/deposit_status_job_spec.rb
```diff
@@ -17,39 +17,40 @@
     allow(Honeybadger).to receive(:notify)
   end
 
-  context 'when the job is a successful work' do
+  context 'when the job is successful' do
```
mjgiarlo, line 20: This is a little test refactoring. No real changes here.

spec/policies/work_policy_spec.rb
```diff
@@ -29,7 +29,7 @@
     end
 
     failed 'when user is a depositor but the collection is depositing' do
-      let(:collection) { build_stubbed :collection, state: 'depositing' }
```
mjgiarlo, line 32: Prefer traits over property setting since we already have nifty traits.

app/models/concerns/eventable.rb
```diff
@@ -0,0 +1,40 @@
...
+# frozen_string_literal: true
+
+# For models that log events
+module Eventable
+  extend T::Sig
+  extend T::Helpers
+
+  abstract!
+
+  # NOTE: This is here to make Sorbet happy with `self.class` in `#valid_event?` :/
```
jcoyne, line 11: 🤢

app/models/event.rb
```diff
@@ -3,6 +3,10 @@
...
   belongs_to :user
+
+  validates :event_type, presence: true, event_type: true
+
+  scope :latest_by_type, ->(type) { where(event_type: type).order('created_at DESC').limit(1).first }
```
jcoyne, line 11: 👍

app/models/event.rb
```diff
@@ -3,6 +3,10 @@
 
 # Records events in the lifecycle of a deposit
 class Event < ApplicationRecord
-  belongs_to :work
+  belongs_to :eventable, polymorphic: true
   belongs_to :user
+
+  validates :event_type, presence: true, event_type: true
```
jcoyne, line 9: I don't understand the need for this.  Can you explain how an event could be set to something invalid?

app/services/description_generator.rb
jcoyne: I'm not sure I understand.  what was wrong with `#to_edtf`?

jcoyne (changes requested):

app/models/concerns/eventable.rb
mjgiarlo: Or we can flip to typed: false!

app/models/event.rb
mjgiarlo: With the other changes made in this commit, I'm not sure it can other than through manual event creation (in the console), which is prolly not worth caring about. I'll remove the validation.

app/services/description_generator.rb
mjgiarlo: I don't always understand what Sorbet does, either.

mjgiarlo:
@jcoyne can you take another look?

jcoyne:
Looks like the tests are borken now.

app/services/description_generator.rb
mjgiarlo: I thought we were OK with some folks using the pre-push hook automation, though?

mjgiarlo:
@jcoyne :speech_balloon: 
> Looks like the tests are borken now.

Tests fixed now.

app/services/description_generator.rb
jcoyne: My understanding was that only changed .rbi files.

mjgiarlo: :man_shrugging: I'm happy to twiddle Sorbet-related stuff in a follow-up PR, though.

jcoyne (approved):

PR #646, opened 2020-12-03
[Refactor] Move state labels to i18n

jcoyne:
## Why was this change made?
This is consistent with how we are handling workflow events


mjgiarlo (approved):
I was thinking of making this change too, so I'm glad to see it done! Looks good other than needing a rebase.

PR #649, opened 2020-12-03
Enforce restrictions on who may view a work

### LINKED ISSUE #648: Works should not show to the general population
There is currently no policy enforcement on `Work#show`

jcoyne:
## Why was this change made?
Fixes #648

### DIFF
--- a/spec/requests/show_work_spec.rb
+++ b/spec/requests/show_work_spec.rb
@@ -5,7 +5,6 @@ require 'rails_helper'
 
 RSpec.describe 'Show a work detail' do
   let(:work) { create(:work) }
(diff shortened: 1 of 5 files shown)


spec/requests/show_work_spec.rb
```diff
@@ -18,11 +18,26 @@
     end
   end
 
-  context 'with an authenticated user' do
+  context 'with an unauthentorized user' do
```
mjgiarlo, line 21: I think this should be `unauthorized`?

mjgiarlo (approved):
Small typo in a spec description, then free to merge.

PR #666, opened 2020-12-03
Show all collections to the admins on the dashboard

### LINKED ISSUE #531: Admin dashboard should show a compact list of the bottom of the collections the admin is not specifically listed on
The admin should have access to all collections and permission to edit any collection or item. However, the admin's dashboard should only display individual tables for those collections where the admin has been explicitly assigned a role as manager, reviewer, or depositor  -- or is the creator. 

All collections -- including those described above -- should be listed in a more compact table at the bottom of the dashboard. See https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/426926001

For example, I am not explicitly listed in this collection, but there is a table for it on my dashboard:

jcoyne:
## Why was this change made?

Fixes #531

Tested on QA.
Screen Shot 2020-12-03 at 5 52 19 PM

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -15,5 +15,7 @@ class DashboardsController < ApplicationController
                       .where('reviewers.user_id' => current_user),
       drafts: Work.with_state(:first_draft).where(depositor: current_user)
     )
+
+    @presenter.work_stats = StatBuilder.build_stats if user_with_groups.administrator?
   end
 end
--- a/app/services/stat_builder.rb
+++ b/app/services/stat_builder.rb
@@ -1,0 +1,33 @@
+# typed: true
+# frozen_string_literal: true
+
+# Runs queries and builds the data structure for the summary table.
+class StatBuilder
+  extend T::Sig
+
+  sig { returns(T::Hash[Collection, Hash]) }
+  def self.build_stats
+    collections_by_id = Collection.all.index_by(&:id)
+
+    by_state.each_with_object({}) do |(id, val), obj|
+      obj[collections_by_id.fetch(id)] = val.merge('total' => total.fetch(id))
+    end
+  end
+
+  sig { returns(T::Hash[Integer, T::Hash[String, Integer]]) }
+  def self.by_state
+    Work.all.group(:collection_id, :state).count.each_with_object({}) do |((id, state), count), obj|
+      obj[id] ||= {}
+      obj[id][state] = count
+    end
+  end
+  private_class_method :by_state
+
+  sig { returns(T::Hash[Integer, Integer]) }
+  def self.total
+    Work.all.group(:collection_id).count.each_with_object({}) do |(id, count), obj|
+      obj[id] = count
+    end
+  end
+  private_class_method :total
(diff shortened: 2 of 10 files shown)


app/controllers/dashboards_controller.rb
```diff
@@ -15,5 +15,7 @@ def show
                       .where('reviewers.user_id' => current_user),
       drafts: Work.with_state(:first_draft).where(depositor: current_user)
     )
+
+    @presenter.work_stats = StatBuilder.build_stats if user_with_groups.administrator?
```
mjgiarlo, line 19: This *feels* like a policy change but I don't feel strongly about whether that means this enforcement ought to move elsewhere.

app/services/stat_builder.rb
```diff
@@ -0,0 +1,33 @@
+# typed: true
+# frozen_string_literal: true
+
+# Runs queries and builds the data structure for the summary table.
+class StatBuilder
+  extend T::Sig
+
+  sig { returns(T::Hash[Collection, Hash]) }
+  def self.build_stats
+    collections_by_id = Collection.all.index_by(&:id)
```
mjgiarlo, line 10: Should we have performance concerns about `Collection.all` here?

app/services/stat_builder.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    by_state.each_with_object({}) do |(id, val), obj|
+      obj[collections_by_id.fetch(id)] = val.merge('total' => total.fetch(id))
+    end
+  end
+
+  sig { returns(T::Hash[Integer, T::Hash[String, Integer]]) }
+  def self.by_state
+    Work.all.group(:collection_id, :state).count.each_with_object({}) do |((id, state), count), obj|
```
mjgiarlo, line 19: Same question as above re: performance.

app/services/stat_builder.rb
```diff
@@ -0,0 +1,33 @@
...
+      obj[id] ||= {}
+      obj[id][state] = count
+    end
+  end
+  private_class_method :by_state
+
+  sig { returns(T::Hash[Integer, Integer]) }
+  def self.total
+    Work.all.group(:collection_id).count.each_with_object({}) do |(id, count), obj|
```
mjgiarlo, line 28: Same question as above re: performance.

mjgiarlo (commented):
Have some questions about performance and where responsibilities lie.

app/services/stat_builder.rb
jcoyne: No, because the requirement is to show all the collections on the page.

app/services/stat_builder.rb
jcoyne: This doesn't show all, because of the 'group' it just gets one row per collection/status.

mjgiarlo (approved):

app/controllers/dashboards_controller.rb
jcoyne: Yeah, I had considered this, but I wasn't convinced that we need a `StatsPolicy#show?` just yet.  If this gets more complicated, we could.

PR #677, opened 2020-12-04
Deposit: clear stale server-side form validation errors after valid data is entered

### LINKED ISSUE #560: work deposit: server-side validation errors aren't cleared when new data is entered in the field
The javascript in edit_deposit_controller.js doesn't clear out the old errors.

jmartin-sul:
## Why was this change made?

fixes #560

Manually on my laptop.  I was unable to get a feature test working for this, and jcoyne mentioned that writing one might be a pain and not worth the trouble, so i'm omitting it from this PR (but will write more detailed steps to reproduce on the ticket, should someone want to test, and i'm happy to also test on stage myself).  Though in trying to write that test, I found a couple seemingly easy touchups I could make to the test code i was using as a guide, and I left those changes in.

n/a

### DIFF
--- a/app/javascript/controllers/edit_deposit_controller.js
+++ b/app/javascript/controllers/edit_deposit_controller.js
@@ -31,6 +31,7 @@ export default class extends Controller {
     const [data, _status, xhr] = event.detail;
     switch (xhr.status) {
       case 400:
+        this.clearErrors()
         this.parseErrors(data)
         break
       default:
@@ -42,12 +43,27 @@ export default class extends Controller {
   parseErrors(data) {
     for (const [fieldName, errorList] of Object.entries(data)) {
       const key = `${fieldName}Field`
-      const target = this.targets.find(key)
-      if (target)
-        target.dispatchEvent(new CustomEvent('error', { detail: errorList }))
-      else
-        console.error(`unable to find target for ${key}`)
+      this.dispatchError(key, errorList)
     }
     window.scrollTo(0, 80)
+  }
+
+  clearErrors() {
+    const clearErrorsTargets = ["titleField", "fileField",
+                                "keywordsField", "contributorsField",
+                                "embargo-dateField"];
+
+    for (const key of clearErrorsTargets) {
+      this.dispatchError(key, [])
+    }
+  }
+
+  dispatchError(key, errorList) {
+    const target = this.targets.find(key)
+    if (target)
+      target.dispatchEvent(new CustomEvent('error', { detail: errorList }))
+    else
+      console.error(`unable to find target for ${key}`)
+
   }
 }
--- a/app/javascript/controllers/files_controller.js
+++ b/app/javascript/controllers/files_controller.js
@@ -5,7 +5,12 @@ export default class extends Controller  {
 
   // Triggered when edit-deposit controller sends an error event
   error(e) {
-    this.containerTarget.classList.add('is-invalid')
-    this.errorTarget.innerHTML = e.detail.join(' ')
+    if (e.detail === null || e.detail.length == 0) {
+      this.containerTarget.classList.remove('is-invalid')
+      this.errorTarget.innerHTML = ''
+    } else {
+      this.containerTarget.classList.add('is-invalid')
+      this.errorTarget.innerHTML = e.detail.join(' ')
+    }
   }
 }
--- a/app/javascript/controllers/files_controller.js
+++ b/app/javascript/controllers/files_controller.js
@@ -6,11 +6,19 @@ export default class extends Controller  {
   // Triggered when edit-deposit controller sends an error event
   error(e) {
     if (e.detail === null || e.detail.length == 0) {
-      this.containerTarget.classList.remove('is-invalid')
-      this.errorTarget.innerHTML = ''
+      this.clearErrors()
     } else {
-      this.containerTarget.classList.add('is-invalid')
-      this.errorTarget.innerHTML = e.detail.join(' ')
+      this.setErrors(e.detail)
     }
+  }
+
+  clearErrors() {
+    this.containerTarget.classList.remove('is-invalid')
+    this.errorTarget.innerHTML = ''
+  }
+
+  setErrors(detail) {
+    this.containerTarget.classList.add('is-invalid')
+    this.errorTarget.innerHTML = detail.join(' ')
   }
 }
--- a/spec/features/create_new_work_spec.rb
+++ b/spec/features/create_new_work_spec.rb
@@ -13,8 +13,6 @@ RSpec.describe 'Create a new work in a deposited collection', js: true do
   end
 
   context 'when successful deposit' do
-    let(:collection_attrs) { attributes_for(:collection) }
-
     it 'deposits and renders work show page' do
       visit dashboard_path
 
@@ -35,7 +33,7 @@ RSpec.describe 'Create a new work in a deposited collection', js: true do
       expect(page).to have_content 'Deposit your content'
 
       # Test client-side validation messages
-      fill_in 'Publication year', with: '2021'
+      fill_in 'Publication year', with: '2049'
       fill_in 'Created year', with: '999'
       click_button 'Deposit'
       expect(page).to have_content(
@@ -86,7 +84,6 @@ RSpec.describe 'Create a new work in a deposited collection', js: true do
       fill_in 'Abstract', with: 'Whatever'
       check 'Musical notation'
 
-      # fill_in 'Citation', with: 'Whatever'
       check 'I agree to the SDR Terms of Deposit'
 
       # Test remote form validation which only happens once client-side validation passes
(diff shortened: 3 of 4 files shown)


spec/features/create_new_work_spec.rb
```diff
@@ -35,7 +33,7 @@
       expect(page).to have_content 'Deposit your content'
 
       # Test client-side validation messages
-      fill_in 'Publication year', with: '2021'
+      fill_in 'Publication year', with: '2049'
```
jmartin-sul, line 36: because otherwise i think the test would stop working when we get back from break?

spec/features/create_new_work_spec.rb
```diff
@@ -125,7 +122,6 @@
       expect(page).to have_content(user.email)
       expect(page).to have_content('sound')
       expect(page).to have_content('Course/instruction, Musical notation, Poetry reading')
-      expect(page).to have_content(user.email)
```
jmartin-sul, line 128: dupe of line 125

spec/features/create_new_work_spec.rb
```diff
@@ -86,7 +84,6 @@
       fill_in 'Abstract', with: 'Whatever'
       check 'Musical notation'
 
-      # fill_in 'Citation', with: 'Whatever'
```
jmartin-sul, line 89: seemed like cruft?

app/javascript/controllers/edit_deposit_controller.js
```diff
@@ -42,12 +43,27 @@ export default class extends Controller {
...
-        target.dispatchEvent(new CustomEvent('error', { detail: errorList }))
-      else
-        console.error(`unable to find target for ${key}`)
+      this.dispatchError(key, errorList)
     }
     window.scrollTo(0, 80)
   }
+
+  clearErrors() {
+    const clearErrorsTargets = ["titleField", "fileField",
```
jmartin-sul, line 52: discussed w/ @jcoyne when i was running my WIP by him yesterday, listing these because they don't include `"title"` or `"file"` from `targets`, and also because the `static` class property is hard to get at in this context (an object instance without an easy reference to the class afaict).

jmartin-sul:
in case this is reviewed before circle is back online:  ran `be srb tc && be rubocop && be rspec` (`be` is my `bundle exec` alias) locally, and all passed.

```
No errors! Great job.
Inspecting 238 files
..............................................................................................................................................................................................................................................

238 files inspected, no offenses detected
### snip ###
Finished in 26.23 seconds (files took 14.38 seconds to load)
308 examples, 0 failures, 10 pending

Randomized with seed 19509
```

spec/features/create_new_work_spec.rb
mjgiarlo: :laughing: good call

spec/features/create_new_work_spec.rb
mjgiarlo: Very likely, yes. :clinking_glasses:

spec/features/create_new_work_spec.rb
mjgiarlo: :eyeglasses:

app/javascript/controllers/edit_deposit_controller.js
```diff
@@ -31,6 +31,7 @@ export default class extends Controller {
     const [data, _status, xhr] = event.detail;
     switch (xhr.status) {
       case 400:
+        this.clearErrors()
```
mjgiarlo, line 34: So we clear the errors before they're parsed? I'm a little surprised to see us clearing errors when we get a 400 back, but maybe I'm just not following closely enough.

app/javascript/controllers/edit_deposit_controller.js
mjgiarlo: Would it be too automagical to, say, filter `this.targets` based on the name of the target (we only want the ones that end in `Field`, it looks like)?

Maybe I'm not grokking why it's challenging to access the `static` class property from here. `clearErrors()` is called right before `parseErrors()` and the latter is using the targets fine (apparently?) judging by line 62.

app/javascript/controllers/edit_deposit_controller.js
```diff
@@ -42,12 +43,27 @@ export default class extends Controller {
...
+
+    for (const key of clearErrorsTargets) {
+      this.dispatchError(key, [])
+    }
+  }
+
+  dispatchError(key, errorList) {
+    const target = this.targets.find(key)
+    if (target)
+      target.dispatchEvent(new CustomEvent('error', { detail: errorList }))
```
mjgiarlo, line 64: So passing an empty list to `errorList` clears the errors? Huh, OK!

app/javascript/controllers/files_controller.js
```diff
@@ -5,7 +5,20 @@ export default class extends Controller  {
 
   // Triggered when edit-deposit controller sends an error event
   error(e) {
+    if (e.detail === null || e.detail.length == 0) {
```
mjgiarlo, line 8: Can you say more about needing both of these checks? (And should the latter be `===` instead of `==` or nah?)

app/javascript/controllers/edit_deposit_controller.js
jmartin-sul: ah right, so not realizing i should've even be filtering, iterating over `this.targets` as a list of strings was the first thing i thought i'd try, but that didn't work, because `this.targets` in the (instantiated object) context of the `parseErrors` or `clearErrors` methods is actually this complex object from Stimulus, not the list of strings defined by the `static` property.  and i couldn't find a way to get at the class fields, even through `this.targets__proto_` (i am also not the most well versed in javascript class mechanics).

i'm not grokking why it's challenging either 😆.  i mean, i get that the instance and the class are two different things.  but i find the magic allowed by this name reuse to be more confusing than helpful.

example screenshots from the debugger:
Screen Shot 2020-12-04 at 5 06 55 PM
Screen Shot 2020-12-04 at 5 10 45 PM

i did also try to use the magic [camel case](https://stimulusjs.org/reference/targets#properties) properties to reference the fields i wanted, but i couldn't get that to work for `embargo-dateField`.  but that was for the best anyway, because switching `clearErrors` to using `this.targets.find` made me realize i could consolidate the error detail dispatching code and its checking for the target's existence to a centralized method shared by `parseErrors` and `clearErrors`.

jmartin-sul: `this.targets.element` is, i think, the form element handled by the controller instance, and it appears to have all its child DOM elements.  in my initial implementation, i iterated over that list of children and sent an error clearing event to each one, but i discussed w/ @jcoyne yesterday, and he thought that was overly broad.  we both figured the duplicative listing of fields here was tolerable, since it's pretty short.  i had actually meant to add a comment on each about keeping the two lists in sync, happy to do that, if you think the benefit might be worth the noise.
Screen Shot 2020-12-04 at 5 21 33 PM

(you can see the file element at index `2`)

app/javascript/controllers/edit_deposit_controller.js
jmartin-sul: from discussing w/ jcoyne and testing locally, it seemed to me that we got all the errors back at once when any server-side errors were transmitted, and so it was safe to clear everything, and then re-flag the things that are still in error based on the `400`.

app/javascript/controllers/edit_deposit_controller.js
jmartin-sul: yeah, that allowed me to use the existing error methods to automatically clear error messages.  it is a bit confusing in that it's an error event, but also sending an empty error list doesn't seem like a terrible way to signal "no errors", so 🤷.  i wasn't wild about this, but i didn't give it a ton of consideration, and am happy to do something with more informative method naming if you'd rather.

app/javascript/controllers/files_controller.js
jmartin-sul: seemed nice to be able to tolerate either a `null` error list or an empty`[]` error list as meaning the same thing in this context.  i had initially passed `null`, but then switched over to `[]`, and so you could argue this is cruft from that.  happy to simplify to the the sole actual usage pattern, if you'd rather, don't feel strongly.

jmartin-sul: oh also, yes, possible that second `==` should be a `===`.  does it make a difference for integer comparison?

app/javascript/controllers/edit_deposit_controller.js
mjgiarlo: Yep, I'd be fine with adding the comments you mention.

app/javascript/controllers/edit_deposit_controller.js
mjgiarlo: OK.

app/javascript/controllers/edit_deposit_controller.js
mjgiarlo: if it works, it works!

app/javascript/controllers/files_controller.js
mjgiarlo: I... am not sure.

jcoyne (approved):

PR #679, opened 2020-12-04
Add labels for the citation and the auto-citation fields

### LINKED ISSUE #669: Accessibility: Labels must be explicitly associated with relevant form field
From AMP tool: 

Developers should define form labels using the label element whenever an on-screen label is present. The label element's value is then explicitly associated with the relevant form field using the for and id attributes. Use of the label element provides an added benefit as it increases the target area that users can click to focus the form field. In certain instances, however, it is impractical or 'undesirable' to use an explicit label for a form field; this is largely driven by the fact that the value between the opening and closing label tags must, by definition, be a block of displayed text.

When on-screen text is not available or can not programmatically be contained in a label (such as when a field is placed in a data table with with a column label and a row label - developers can use another method to provide an accessible name such as by using the title, aria-label, or aria-labelledby attribute to provide the label (see below for details on using aria-labelledby and aria-label). Note: When an accessible name exists from an explicit label, aria-label, or aria-labelledby AND a title attribute is used -- the title attribute will act as the accessible description for the input field and does not override the accessible name. The aria-describedby property can also be used to provide a description for the input field such as to associate instructions or error messages with a field.

The accessible name must sufficiently indicate the purpose/usage of the form field as conveyed by the visual label and/or the visual layout. This is important because some users of assistive technology such as speech recognition software may interact with a form field by speaking the field's visual label. When the programmatic label does not match the visual label or obstacle may be created for users of speech recognition software. Furthermore, to prevent confusion with AT to know which programmatic information to use as the accessible name, title, label, or ARIA based names should be used in a mutually exclusive fashion unless title is acting as the accessible description..

The aria-label and aria-labelledby attributes can also be used to provide a label or labels for form fields to assistive technology that supports ARIA. The aria-labelledby attribute must contain one or more ids of elements that label the form field. Note: for multiple labels to properly be rendered in older versions of Internet Explorer it may be necessary to set tabindex="-1" on the elements designated as labels. The aria-label attribute should only be used when on-screen labels can not be used. Developers should place the string that labels the form field in the aria-label attribute. Note: The spelling of aria-labelledby is the British spelling with two "l"s.

Notes:

Whenever a form field is required, or some other form of constraint applies, developers should ensure the "required" indicator (typically an asterisk *) is included in the accessible name, the label, title attribute, aria-label, or content referenced by aria-labelledby. A better solution is to use the aria-required attribute to indicate that a form field is required.
Also, assistive technologies such as screen readers can become confused by the presence of formatting markup between the opening and closing label tags, resulting in the incorrect reading of its text value. Examples include the emphasis elements strong and b, and the line break element br. Developers should move any desired formatting markup outside the label block, rather than inside it, or verify that the usage is rendered correctly in screen readers. (This does not apply to the use of form input labels put in the input's title attribute.)
There should be no more than one label explicitly associated to a given form field.
HTML5 placeholder attribute should not be used as a substitute for the label of the form field. If the HTML5 placeholder is used as a label of the form field, an alternative method must be provided. For example, an explicit label, a title attribute or the CSS off-screen text.
Some assistive technology may not support implicit form labels. For assistive technology such as speech recognition that does not support this another method should be provided. An implicit label is a label that wraps label text and a form field but that does not contain an explicit for and id relationship.

jcoyne:
## Why was this change made?

Fixes #669

Tested locally by using the Firefox accessability tool. There are no user visible changes aside from margins below the subtypes, which more closely resemble the mockups:

Screen Shot 2020-12-05 at 10 03 32 AM

Before:

Screen Shot 2020-12-05 at 10 05 59 AM

### DIFF
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -25,7 +25,7 @@
     </div>
   </div>
 
-  <div class="mb-3 row">
+  <div class="mb-5 row">
     <% if other_type? %>
       <%= form.label :subtype, 'Other', class: 'col-sm-2 col-form-label' %>
       <div class="col-sm-10">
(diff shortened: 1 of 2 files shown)


mjgiarlo (changes requested):
Tests are busted. And two requests:

1. Can you include before/after screenshots?
2. Can you comment on how this was tested per the PR template?

jcoyne:
@mjgiarlo changes made.

mjgiarlo (approved):

PR #688, opened 2020-12-04
send email to depositors when added to collection

### LINKED ISSUE #588: Email notifications: "Depositor: Invitation to deposit to a collection"
https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#heading=h.cbc433yprhtx

Example implementation: https://github.com/sul-dlss/happy-heron/pull/673

peetucket:
## Why was this change made?

Fixes #588: send notifications to new depositors on a collection

~~Problem: what happens when a user saves a draft (thus not sending notifications), then later comes back and submits the collection, without making any changes to the list of depositors.  Currently the logic in this PR would not send out notifications because no changes to depositors were made at the time of actual deposit.~~  See comments in #588 for how this is not a problem based on the implementation

New tests


app/controllers/collections_controller.rb
```diff
@@ -80,6 +82,20 @@ def after_save(collection)
     redirect_to dashboard_path
   end
 
+  def send_depositor_notifications(collection, new_depositors)
```
aaron-collier, line 85: perhaps nitpicky, but does `new_depositors` makes sense here since we send to all depositors on line 33? While not a reason to hold this up, while reading this, the name made me go back and see how it was used - so a bit confusing.

peetucket: good call, i'll switch this to `depositors`

app/mailers/notification_mailer.rb
```diff
@@ -32,4 +32,10 @@ def submitted_for_review_email
     @work = params[:work]
     mail(to: @user.email, subject: "A Depositor has submitted a deposit in the #{@work.collection_name} collection")
   end
+
+  def invitation_to_deposit_email
```
jcoyne, line 36: Can you move this out of the NotificationMailer and move it to a "CollectionsMailer"?  Currently the notifications mailer is all about events on a work, and this new message is about events on a collection.  It seems like this would make it easier to find later.

jcoyne (changes requested):

spec/requests/create_collection_spec.rb
```diff
@@ -108,34 +115,39 @@
...
+        context 'with mostly empty fields' do
           let(:draft_collection_params) do
             {
               name: '',
               description: '',
               contact_email: '',
               manager_sunets: user.sunetid,
               access: 'world',
-              depositor_sunets: ''
+              depositor_sunets: 'maya.aguirre,jcairns, cchavez, premad, giancarlo, zhengyi'
```
jcoyne, line 134: This does not belong in the "mostly empty fields". This was a test to show it was okay to provide nothing, and now we don't have it.

jcoyne (changes requested):

spec/requests/create_collection_spec.rb
peetucket: ok, i'll revert this test and add a new context block with a new test

app/mailers/notification_mailer.rb
peetucket: yeah, i can do that -- though then we want to rename `NotificationMailer` to `WorksMailer` for consistency.

jcoyne: That sounds like a splendid idea.

jcoyne (approved):

app/mailers/notification_mailer.rb
peetucket: Done in #747

PR #703, opened 2020-12-07
Add a page to show the works for a collection

### LINKED ISSUE #687: Collection deposits page should show all items in the collection
This page, unlike the dashboard, should list all the items in the collection without needing to ask for them. The "See all deposits" text should be removed as it is not needed with all items displayed all the time.

### LINKED ISSUE #693: Refreshing Collection Deposits page should keep me on that page
When I refresh the collection deposits page seen below:

it reloads the collections details tab instead:

https://sul-h2-stage.stanford.edu/collections/79

### LINKED ISSUE #694: "See all deposits" text on dashboard to take user to Collection Deposits page
Now that the collection deposits page is implemented, these links on the dashboard -- which should appear for every collection with more than 4 items -- should go to the Collections Details page so they can see the full list of items.

(Note that the full list of items needs to be displayed on that page as well. That's ticket #687).

jcoyne:
Using the Dashboard::WorksComponent was not sufficient.

Fixes #693
Fixes #694
Fixes #687

Screen Shot 2020-12-07 at 12 24 19 PM

### DIFF
--- a/app/components/collections/header_component.html.erb
+++ b/app/components/collections/header_component.html.erb
@@ -1,0 +1,23 @@
+<header class="title">
+  <span class="header-text"><%= @collection.name %></span>
+  <% if helpers.allowed_to?(:create?, Work.new(collection: @collection)) %>
+    <span class="deposit-button" data-controller="work-type">
+      <%= button_tag '+ Deposit to this collection',
+            data: {
+              destination: new_collection_work_path(@collection),
+              toggle: 'modal',
+              target: '#workTypeModal',
+              action: "work-type#set_collection"
+            },
+            class: "btn btn-primary" %>
+
+      <div class="modal fade" id="workTypeModal" tabindex="-1" aria-labelledby="content-type-prompt" aria-hidden="true">
+        <div class="modal-dialog modal-dialog-centered modal-lg">
+          <div class="modal-content">
+            <%= render Works::WorkTypeModalComponent.new %>
+          </div>
+        </div>
+      </div>
+    </span>
+  <% end %>
+</header>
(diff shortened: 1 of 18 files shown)


app/components/collections/tabs_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<section class="collection-nav">
```
jcoyne, line 1: This used to be `views/collections/_nav.html.erb`

app/components/collections/header_component.html.erb
```diff
@@ -0,0 +1,23 @@
+<header class="title">
```
jcoyne, line 1: extracted from `views/collections/show.html.erb` since it's now used by Collections#show and Works#index

app/policies/work_policy.rb
```diff
@@ -7,8 +7,12 @@ class WorkPolicy < ApplicationPolicy
   alias_rule :delete?, to: :destroy?
 
   relation_scope :edits do |scope|
-    scope.where(depositor: user)
-         .or(scope.where(collection_id: [user.manages_collection_ids + user.reviews_collection_ids]))
+    if administrator?
```
mjgiarlo, line 10: Should we add policy spec coverage for this or do we already have it?

mjgiarlo (commented):
Have a question about policy spec coverage but otherwise looks good. Can you also include a screenshot of the page you added?

app/policies/work_policy.rb
jcoyne: It's covered by the request spec, which is why it didn't seem necessary to have a unit test.

jcoyne:
@mjgiarlo I added a screenshot, but it's no different than what was there before, it's only at a new route.

mjgiarlo (approved):

PR #721, opened 2020-12-08
Remove email domain from display of email usernames

### LINKED ISSUE #692: Collection view page should display only SUNet IDs in fields with users
Currently the display is of email addresses. Please make these just the SUNet. 

Fields I see affected:

- created by
- managers
- depositors
- reviewers

aaron-collier:
## Why was this change made?

Fixes #692 

Before:

Screen Shot 2020-12-08 at 2 35 12 PM

After:

Screen Shot 2020-12-08 at 2 35 28 PM

Updated unit and request tests

### DIFF
--- a/app/components/collections/information_component.html.erb
+++ b/app/components/collections/information_component.html.erb
@@ -16,7 +16,7 @@
     </tr>
     <tr>
       <th>Created by</th>
-      <td><%= creator %></td>
+      <td><%= creator.sunetid %></td>
     </tr>
     <tr>
       <th>Collection created</th>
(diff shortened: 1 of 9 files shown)


jcoyne (changes requested):
Can we use `collection.creator.sunetid` throughout?

aaron-collier:
@jcoyne yeah, I just saw that when working on another PR. Will do. I wasn't entirely sure if they would always be sunets, but I suspect that only makes sense.

aaron-collier:
@jcoyne done.

mjgiarlo (approved):

jcoyne (approved):

PR #725, opened 2020-12-08
Use remote forms consistently across app

mjgiarlo:
## Why was this change made?

Consistency is good. In doing this work, I replaced a single usage of `form_for` with `form_with` so the app does not use a mix of local and remote forms, and removed the model argument to this call which is/was unused. This commit also explicitly marks all of our forms as remote forms for future-proofing, as we know an upcoming Rails 6.1 change will make `form_with` default back to local forms.

I found that the mediated deposit versioning spec was flappy locally because there can be two links on the page with the work title, so I changed the spec to look for the link only in the scope of the collection name (the deposits in progress section). To do this, I moved the collection name from a section header to a table caption. I do not notice any UI difference with this change.

CI, local browser testing

None

### DIFF
--- a/app/components/dashboard/collection_component.html.erb
+++ b/app/components/dashboard/collection_component.html.erb
@@ -1,14 +1,13 @@
 <section>
-  <header><%= link_to name, collection %>
-    <% if allowed_to?(:edit?, collection) %>
-      <%= link_to edit_collection_path(collection), aria: { label: "Edit #{name}"} do %>
-        <span class="fas fa-pencil-alt"></span>
-      <% end %>
-    <% end %>
-    <%= render Collections::DeleteComponent.new(collection: collection) %>
-  </header>
-
   <table class="table">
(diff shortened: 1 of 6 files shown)


app/components/help/form_component.html.erb
```diff
@@ -7,7 +7,7 @@
       </div>
       <div class="modal-body" data-controller="help">
         <%= form_with(url: help_path,
-                      method: 'post',
```
mjgiarlo, line 10: POST is the default method.

app/components/dashboard/collection_component.html.erb
```diff
@@ -1,14 +1,13 @@
 <section>
-  <header><%= link_to name, collection %>
```
mjgiarlo, line 2: See PR description for rationale behind this change.

jcoyne (approved):

PR #726, opened 2020-12-08
use factory trait for deposited instead of hardcoding string

peetucket:
## Why was this change made?

Small refactor to use factory traits where possible

Test suite


mjgiarlo (approved):
Thank you! I've been wanting to do this for some time.

PR #735, opened 2020-12-09
Reduce number of queries to render dashboard without sacrificing readability

mjgiarlo:
## Why was this change made?

See above.

CI

None

### DIFF
--- a/app/components/dashboard/collection_component.html.erb
+++ b/app/components/dashboard/collection_component.html.erb
@@ -19,7 +19,7 @@
       </tr>
     </thead>
     <tbody>
-      <% visible_deposits.each do |work| %>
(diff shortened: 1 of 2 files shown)


jcoyne (approved):

app/components/dashboard/collection_component.html.erb
```diff
@@ -40,7 +40,7 @@
     </tbody>
   </table>
 
-  <% if total_deposits_count > MAX_DEPOSITS_TO_SHOW %>
+  <% if visible_deposits.count > MAX_DEPOSITS_TO_SHOW %>
```
jcoyne, line 43: Can total_deposits_count be removed now?

jcoyne (changes requested):

jcoyne (approved):

PR #740, opened 2020-12-09
add delete button to works and collection edit page

### LINKED ISSUE #705: I should be able to delete draft deposits on the deposit forms for both works and collections.
Item deposit form, at the bottom left:

https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403

Collection deposit form, at the bottom left: 

https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930437

peetucket:
## Why was this change made?

Fixes #705 (add delete button to draft works and collection edit pages)
Note: this discard button will only show up on the edit page when its possible to delete (i.e. on a first draft of a work or a collection that has been saved as a draft previously at least once)

**Item edit page:**

Screen Shot 2020-12-09 at 5 11 04 PM

------

**Collection edit page:**

Screen Shot 2020-12-09 at 5 10 28 PM

Added new unit tests and localhost browser


jcoyne:
Please add screenshots

peetucket:
> Please add screenshots

Added to the PR description

app/components/buttons_component.html.erb
```diff
@@ -1,6 +1,2 @@
-<div class="row justify-content-end">
-  <div class="col-auto">
-    <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button' %>
-    <%= form.submit submit_button_label, class: 'btn btn-primary' %>
-  </div>
-</div>
+<%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button' %>
```
jcoyne, line 1: Rather than strip out this HTML, can we add the delete button into the ButtonsComponent?  These seems to appear the same between on both the collection and works edit pages.

jcoyne (changes requested):

app/components/buttons_component.html.erb
peetucket: We could do that, though we'd then end up duplicating the delete link since we'd we still need the separate `DeleteComponent` for the dashboard (same link but styled differently).  This was an attempt to centralize the delete links from any location into a single component.

jcoyne: Could `DeleteComponent` be called by the `ButtonComponent`?  The ButtonsComponent was intended to be the full last row of the form with all the buttons on it.

peetucket: Yeah, I could do that, though I'd have do to some logic in the `ButtonComponent` to pick the correct one for the form we are on, since the `DeleteComponent` is specific to a work of a collection.  Let me give it a try.

app/components/buttons_component.html.erb
```diff
@@ -1,4 +1,11 @@
-<div class="row justify-content-end">
+<div class="row justify-content-between">
+  <div class="col-auto">
+    <% if model_type == 'Work' %>
```
jcoyne, line 3: Can you move this into `ButtonsComponent` (the `.rb` file)?

app/components/collections/delete_component.html.erb
```diff
@@ -1,3 +1,7 @@
 <%= link_to collection_path(collection), method: :delete, aria: { label: "Delete #{collection.name}" }, data: { confirm: "Are you sure you want to delete this draft collection? It cannot be undone." } do %>
-  <span class="far fa-trash-alt"></span>
+  <% if @style == :icon %>
```
jcoyne, line 2: Can you move this into a method in `DeleteComponent`?

spec/components/works/work_form_component_spec.rb
```diff
@@ -6,7 +6,7 @@
 RSpec.describe Works::WorkFormComponent do
   let(:component) { described_class.new(work_form: work_form) }
   let(:form) { ActionView::Helpers::FormBuilder.new(nil, work_form, controller.view_context, {}) }
-  let(:work) { build(:work, collection: build(:collection, id: 7)) }
+  let(:work) { create(:work, collection: build(:collection, id: 7)) }
```
jcoyne, line 9: Can you use `build_stubbed` instead? This will keep the test suite fast.

spec/features/delete_draft_collection_spec.rb
```diff
@@ -5,29 +5,38 @@
 
 RSpec.describe 'Delete a draft collection', js: true do
   let(:user) { create(:user) }
+  let(:collection_attrs) { attributes_for(:collection) }
+  let(:collection) { Collection.last }
 
   before do
     sign_in user, groups: ['dlss:hydrus-app-collection-creators']
+    visit dashboard_path
```
jcoyne, line 13: Can we just `create(:collection)` here instead of going through the UI?  Presumably create a collection is tested elsewhere.

spec/features/delete_draft_work_spec.rb
```diff
@@ -12,13 +12,22 @@
...
       expect(work).to be_first_draft
       visit dashboard_path
       accept_confirm do
         click_link "Delete #{work.title}"
       end
       expect(Work.find_by(id: work.id)).to be(nil)
     end
+
+    it 'allow users to delete the work and destroys the model from the work edit page' do
+      expect(work).to be_first_draft
```
jcoyne, line 25: I don't think this is a useful assertion.  We told it to be a first_draft when we created it.

spec/features/delete_draft_work_spec.rb
```diff
@@ -12,13 +12,22 @@
...
+      visit edit_work_path(work)
+      accept_confirm do
+        click_link 'Discard draft'
+      end
+      expect(Work.find_by(id: work.id)).to be(nil)
```
jcoyne, line 30: `expect(Work.exists?(work.id)).to be false` ?

jcoyne (changes requested):

app/components/buttons_component.html.erb
jcoyne: Looks much better now. Thank you.

spec/features/delete_draft_collection_spec.rb
peetucket: I originally had used the factory when I first create this test in a previous PR, but it left the collection in a state where it couldn't be accessed by the user logging in (and therefore the delete button was not able to be rendered).  After trying various things and pairing with Naomi, we got the test to work by just creating it via the UI.  Maybe you have some ideas on how to get this work?

spec/features/delete_draft_work_spec.rb
peetucket: Will remove

app/components/buttons_component.html.erb
peetucket: Not sure what I should move (and same comment for the delete component)?  Can you render HTML from within the component class?  Or just create a boolean method instead of the comparison here?

jcoyne:
Looks like a rebase is needed and then turn off the typechecking in the component

peetucket:
@jcoyne all changes made

jcoyne (approved):

PR #741, opened 2020-12-09
Embargo Date Validation

### LINKED ISSUE #556: When "On this date" is selected, full embargo date must be provided
Currently it doesn't submit the form, but it provides no visual indication that something is wrong.

jermnelson:
## Why was this change made?
Fixes #556 

Displays the following when month and day are not set:
Screen Shot 2020-12-10 at 11 04 48 AM

New and existing tests.

n/a

### DIFF
--- a/app/validators/embargo_date_parts.rb
+++ b/app/validators/embargo_date_parts.rb
@@ -4,7 +4,8 @@
 # Validates embargo month and day
 class EmbargoDateParts < ActiveModel::Validator
   def validate(record)
-    return if record.send('embargo_date').instance_of?(Date)
+    return if record.embargo_date.instance_of?(Date)
+    return if record.release == 'immediate'
 
     month = record.send('embargo_date(2i)')
     day = record.send('embargo_date(3i)')
--- a/app/validators/embargo_date_parts.rb
+++ b/app/validators/embargo_date_parts.rb
@@ -1,0 +1,13 @@
+# typed: true
+# frozen_string_literal: true
+
+# Validates embargo month and day
+class EmbargoDateParts < ActiveModel::Validator
+  def validate(record)
+    return if record.send('embargo_date').instance_of?(Date)
+
+    month = record.send('embargo_date(2i)')
+    day = record.send('embargo_date(3i)')
+    record.errors.add('embargo-date', 'Must provide all parts') if month.nil? || day.nil?
+  end
+end
--- a/app/validators/embargo_date_parts.rb
+++ b/app/validators/embargo_date_parts.rb
@@ -4,11 +4,10 @@
 # Validates embargo month and day
 class EmbargoDateParts < ActiveModel::Validator
   def validate(record)
+    # If we've already successfully deserialized the date, there's no point to continuing
     return if record.embargo_date.instance_of?(Date)
-    return if record.release == 'immediate'
 
-    month = record.send('embargo_date(2i)')
-    day = record.send('embargo_date(3i)')
-    record.errors.add('embargo-date', 'Must provide all parts') if month.nil? || day.nil?
+    # Adds error because embargo date is nil (may be missing month and day values in form)
+    record.errors.add('embargo-date', 'Must provide all parts')
   end
 end
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -29,7 +29,7 @@ class DraftWorkForm < Reform::Form
   validates :created_edtf, created_in_past: true
   validates :published_edtf, created_in_past: true
   validates :embargo_date, embargo_date: true
-  validates_with EmbargoDateParts
+  validates_with EmbargoDateParts, unless: proc { |form| form.release == 'immediate' }
 
   def deserialize!(params)
     # Choose between using the user provided citation and the auto-generated citation
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -29,6 +29,7 @@ class DraftWorkForm < Reform::Form
   validates :created_edtf, created_in_past: true
   validates :published_edtf, created_in_past: true
   validates :embargo_date, embargo_date: true
+  validates_with EmbargoDateParts
 
   def deserialize!(params)
     # Choose between using the user provided citation and the auto-generated citation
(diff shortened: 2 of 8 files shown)


jermnelson:
Currently failing in the following two specs:

- `spec/requests/edit_work_spec.rb`
- `spec/requests/create_work_spec.rb`

``` ruby  
  Failure/Error: render 'errors', status: :bad_request                                                      
                                                                                                               
     ActionView::MissingTemplate:                                                                              
       Missing template works/errors, objects/errors, application/errors with {:locale=>[:en], :formats=>[:html
], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby, :jbuilder]}. Searched in:                    
         * "/Users/jpnelson/2020/sul-dlss/happy-heron/app/views"                                               
         * "/Users/jpnelson/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/devise-4.7.3/app/views"             
         * "/Users/jpnelson/.rbenv/versions/2.7.1/lib/ruby/gems/2.7.0/gems/view_component-2.22.1/app/views"    
     # ./app/controllers/works_controller.rb:32:in `create'    
```

On line 32 in the`works_controller.rb` the render call is not finding the correct jbuilder template at `app/view/works/errors.json.jbuilder` view https://github.com/sul-dlss/happy-heron/blob/04f7da46fdf6fe4d85c6549965c5045b6af056b2/app/controllers/works_controller.rb#L26-L33

app/validators/embargo_date_parts.rb
```diff
@@ -0,0 +1,14 @@
+# typed: true
+# frozen_string_literal: true
+
+# Validates embargo month and day
+class EmbargoDateParts < ActiveModel::Validator
+  def validate(record)
+    return if record.send('embargo_date').instance_of?(Date)
```
jcoyne, line 7: Can you explain why this clause is necessary?

app/validators/embargo_date_parts.rb
```diff
@@ -0,0 +1,14 @@
...
+    return if record.send('release') == 'immediate'
```
jcoyne, line 8: Can you move this up into the form?  E.g. `validates_with EmbargoDateParts, if: Proc.new { |form| form.release == 'immediate' }`

app/validators/embargo_date_parts.rb
jcoyne: Also, why `send` and not just `record.embargo_date`?

jcoyne (changes requested):
Can you add a screenshot showing what the validation looks like to the user?

app/validators/embargo_date_parts.rb
jermnelson: Thanks, changed to  use `record.embargo_date`. The reason for this guard clause is that this test https://github.com/sul-dlss/happy-heron/blob/1190a978784dc06e275873c6970eceb5cededbcd/spec/requests/create_work_spec.rb#L261, fails because a valid Date is being passed into the `EmbargoDateParts` validator and we removed the month and day parameters in the filter https://github.com/sul-dlss/happy-heron/blob/1190a978784dc06e275873c6970eceb5cededbcd/app/forms/embargo_date_params_filter.rb#L20-L21
causing the validation to fail as we would expect. Both this guard clause and the next line are because of test failures (see also my comment below).

app/validators/embargo_date_parts.rb
jermnelson: Moving this up to the draft_work_form.rb causes four failing tests in the `spec/requests/create_work_spec.rb` (test on lines 328, 358, 396, 463).

app/validators/embargo_date_parts.rb
jcoyne: Can you add a comment to the effect of, `# If we've already successfully deserialized the date, there's no point to continue with this`

app/validators/embargo_date_parts.rb
jcoyne: sorry, the `if` should be `unless`.  Does that fix it?

jermnelson: Yup! That fixed the issues with the failing tests. Just pushed up a new commit.

jcoyne (changes requested):
I think this ought to have an integration test (in `spec/requests/create_work_spec.rb`) where we can see this error.

app/validators/embargo_date_parts.rb
```diff
@@ -0,0 +1,14 @@
+# typed: true
+# frozen_string_literal: true
+
+# Validates embargo month and day
+class EmbargoDateParts < ActiveModel::Validator
+  def validate(record)
+    # If we've already successfully deserialized the date, there's no point to continue with this
+    return if record.embargo_date.instance_of?(Date)
+
+    month = record.send('embargo_date(2i)')
```
jcoyne, line 10: Are these ever not nil?  Haven't they been deleted in EmbargoDateParams filter?

jcoyne (changes requested):

jermnelson:
> I think this ought to have an integration test (in `spec/requests/create_work_spec.rb`) where we can see this error.

Do you mean an integration test in the `spec/features/create_work_spec.rb`?

jcoyne:
I think in the requests spec should be sufficient. I'm not so concerned about the front end.

app/validators/embargo_date_parts.rb
jermnelson: Changed to just check if `embargo_date` is nil

app/validators/embargo_date_parts.rb
```diff
@@ -0,0 +1,12 @@
+# typed: true
+# frozen_string_literal: true
+
+# Validates embargo month and day
+class EmbargoDateParts < ActiveModel::Validator
+  def validate(record)
+    # If we've already successfully deserialized the date, there's no point to continue with this
+    return if record.embargo_date.instance_of?(Date)
+
+    record.errors.add('embargo-date', 'Must provide all parts') if record.embargo_date.nil?
```
jcoyne, line 10: I don't think this conditional is necessary, it's either a date or its nil.

jcoyne (changes requested):

jcoyne (approved):

PR #675, opened 2020-12-04
Add a link to show the help form

jcoyne:
## Why was this change made?

Fixes #672

PR #743, opened 2020-12-09
Update citation after deposit

### LINKED ISSUE #657: PURL needs to be added to user-edited citation when placeholder text is left in.
If the user switches off the default citation to manually edit the citation, but leaves in the `:link:` text for the PURL, then the PURL should be automatically entered when it is available. Currently it's showing this as "Available at :text:".

Alternatively, if this is can't be done once it's been switched to the manual edit, then I propose when the switch is clicked that the "Available at :link:" text be removed entirely for manually editing.

jcoyne:
## Why was this change made?
Fixes #657

Tested on QA

### DIFF
--- a/app/components/works/detail_component.html.erb
+++ b/app/components/works/detail_component.html.erb
@@ -29,7 +29,7 @@
     <tbody>
       <tr>
         <th scope="row">Persistent link</th>
-        <td data-work-updates-target="url">
+        <td data-work-updates-target="purl">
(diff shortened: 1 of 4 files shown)


peetucket:
I think the linked issue above (#675) is not the correct one for this PR?

jcoyne:
Thanks @peetucket, updated to #657

peetucket (approved):

PR #754, opened 2020-12-10
Show when a collection is depositing

### LINKED ISSUE #653: Add "Depositing" and spinner icon when collection deposit is in progress
I would like to see this added to the dashboard and the collection view page as well -- similar to the items -- for when a deposit is in progress. This will make it less confusing for the user when they don't see the edit icon or the button to add items to the collection.

jcoyne:
## Why was this change made?
Fixes #653

Tested on QA

Dashboard
Screen Shot 2020-12-10 at 8 35 17 AM

Show page

Screen Shot 2020-12-10 at 8 37 04 AM

### DIFF
--- a/app/components/collections/information_component.html.erb
+++ b/app/components/collections/information_component.html.erb
@@ -8,7 +8,13 @@
   <tbody>
     <tr>
       <th>PURL</th>
-      <td><%= link_to purl, purl unless purl.nil? %></td>
+      <td data-state-updates-target="purl">
+        <% if purl %>
+          <%= link_to purl, purl %>
+        <% else %>
+          <em>Link will become available once the collection has been deposited.</em>
+        <% end %>
+      </td>
     </tr>
     <tr>
       <th>Version details</th>
--- a/app/components/dashboard/collection_component.html.erb
+++ b/app/components/dashboard/collection_component.html.erb
@@ -25,7 +25,12 @@
             <% end %>
             <%= render Works::DeleteComponent.new(work: work) %>
           </td>
-          <td><%= render Works::StateDisplayComponent.new(work: work) %></td>
+          <% if work.depositing? %>
+            <td data-controller="state-updates" data-state-updates-work-id="<%= work.id %>" data-state-updates-target="state">
+          <% else %>
+            <td>
+          <% end %>
+          <%= render Works::StateDisplayComponent.new(work: work) %></td>
           <td><%= I18n.l(work.updated_at.to_date, format: :abbr_month) %></td>
           <td><%= link_to work.purl, work.purl if work.purl %></td>
           <td><span class="fas fa-quote-left"></span> Cite</td>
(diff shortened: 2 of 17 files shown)


app/components/collections/information_component.html.erb
```diff
@@ -8,7 +8,13 @@
   <tbody>
     <tr>
       <th>PURL</th>
-      <td><%= link_to purl, purl unless purl.nil? %></td>
+      <td data-state-updates-target="purl">
+        <% if purl %>
+          <%= link_to purl, purl %>
+        <% else %>
+          <em>Link will become available once the work has been deposited.</em>
```
mjgiarlo, line 15: `s/work/collection/` ?

app/components/dashboard/collection_component.html.erb
```diff
@@ -30,7 +25,12 @@
             <% end %>
             <%= render Works::DeleteComponent.new(work: work) %>
           </td>
-          <td><%= render Works::StateDisplayComponent.new(work: work) %></td>
+          <% if work.depositing? %>
+            <td data-controller="state-updates" data-state-updates-work-id="<%= work.id %>" data-state-updates-target="state">
```
mjgiarlo, line 29: I wonder if we should rename work-id to object-id so it's more clearly labeled as being used for both works and collections.

app/components/dashboard/collection_header_component.rb
```diff
@@ -0,0 +1,27 @@
...
+    delegate :allowed_to?, to: :helpers
+    delegate :depositing?, to: :collection
+
+    sig { returns(String) }
+    def name
+      collection.name.presence || 'No Title'
+    end
+
+    sig { returns(String) }
+    def spinner
```
mjgiarlo, line 23: I wonder if this could move to the ApplicationComponent but I don't think it matters too much right now.

app/components/dashboard/collection_component.html.erb
mjgiarlo: Ah, nevermind.

app/services/broadcast_collection_change.rb
```diff
@@ -0,0 +1,15 @@
+# typed: strict
+# frozen_string_literal: true
+
+# Sends messages about a collection changing state to the CollectionUpdatesChannel
+class BroadcastCollectionChange
+  extend T::Sig
+
+  sig { params(collection: Collection, state: Symbol).void }
+  def self.call(collection:, state:)
+    return unless state == :deposited
```
mjgiarlo, line 10: Can you say more about the differences between this service and the one for works? E.g., the guard clause on line 10, and broadcasting the state as `''`?

jcoyne: Collections only show "depositing" otherwise they don't show a stare

app/components/dashboard/collection_header_component.rb
jcoyne: I would not put it there, because this isn't something for every component.  If anything it could be its own component.

app/services/broadcast_collection_change.rb
jcoyne: I've added a comment to this effect.

mjgiarlo (approved):

Issue #589, opened 2020-11-24
Email notifications: "Depositor: Draft in progress, not submitted/published"


peetucket:
https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#heading=h.lj2h48ox03kd

Example implementation: https://github.com/sul-dlss/happy-heron/pull/673

jmartin-sul:
closing this for now, but will note here the possible delta between the implementation that was merged last month and the spec...

from the PR description
> This PR as is has a bit of a delta from [the spec]: I'd originally interpreted as notify at two weeks, then every four weeks after that... but upon re-reading, I think the intent is notify at two weeks, again at four weeks, then every four weeks after that? If so, it should be easy enough to make that change in this PR or in a follow on (happy to do either). [_PR is now merged, so would need a follow-on now)_]

the comments on the spec doc linked from this issue say:
> @hannahfrost -- How much time has elapsed before the reminder is issued? How many reminders are sent before they cease? When a recipient's SUNet account becomes inactive before completing a deposit, can we detect this and cease sending the notifications?
>
> @amyehodge -- My inclination is to go with two weeks, then at a month, and monthly thereafter. I wouldn't stop as long as the SUNet is valid, but it would be nice if they had an option to request that they stop. That might be out of scope right now.

there is, indeed, no attempt at present to verify whether the SUNet is still valid.

if we want to adjust the interval between reminders, or validate that the SUNet is still valid, i'd be slightly inclined to open separate new issues for either/each of those things, rather than reopening this one (but i don't feel strongly about that).

amyehodge:
@jmartin-sul I think that notification interval is fine.

It would be great to have verification if SUNet IDs are still valid, but perhaps we would want that for more than just this notification. Question: what happens if the email is sent and the SUNet isn't valid? Would that bounce back to somewhere, basically letting us know it isn't valid? That could be a way of getting this info indirectly.

amyehodge:
I would have expected to have received a notification about [this work](https://sul-h2-stage.stanford.edu/works/13373) that has been in draft since April 20, 2021, but I haven't. Can someone check on this?

amyehodge:
filed new ticket #1419

Issue #592, opened 2020-11-24
Email notifications: "Depositor: New version in progress, not submitted/published"


peetucket:
https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#heading=h.awan2wkp6vwt

Example implementation: https://github.com/sul-dlss/happy-heron/pull/673

amyehodge:
Leah should have received this notification about [this work](https://sul-h2-stage.stanford.edu/works/13273) that has been in draft state since April 5, 2021, but she hasn't. Can someone check on why this isn't going out on schedule? Thanks.

amyehodge:
filed new ticket #1420

amyehodge:
filed new ticket #1420

PR #756, opened 2020-12-10
send reminder emails for draft versions of deposits

jmartin-sul:
## Why was this change made?

implement #589.  from the [spec](https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#heading=h.lj2h48ox03kd):
> "How much time has elapsed before the reminder is issued?  How many reminders are sent before they cease?"
>
> "My inclination is to go with two weeks, and then a month, and then monthly thereafter.  I wouldn't stop as long as the SUNet is valid, but it would be nice if they had an option to request that they stop.  That might be out of scope right now."

This PR as is has a bit of a delta from that description:  I'd originally interpreted as notify at two weeks, then every four weeks after that... but upon re-reading, I think the intent is notify at two weeks, again at four weeks, then every four weeks after that?  If so, it should be easy enough to make that change in this PR or in a follow on (happy to do either).

unit tests.  some manual testing locally on rails console and using the mailer preview, e.g.:
Screen Shot 2020-12-10 at 9 38 02 AM

Added the `whenever` gem and an initial daily cron job to run the new functionality.


app/mailers/works_mailer.rb
```diff
@@ -15,6 +15,15 @@ def deposited_email
     mail(to: @user.email, subject: "Your deposit, #{@work.title}, is published in the SDR")
   end
 
+  def first_draft_reminder_email
+    @work = params[:work]
+    @user = @work.depositor
+
+    subject = "Reminder: New version of a deposit to the #{@work.collection.name} collection in the SDR is in progress"
```
jcoyne, line 22: can you use `@work.collection_name` here so that this object doesn't need to depend on a deep graph of dependencies?

app/models/work.rb
```diff
@@ -88,6 +88,30 @@ class Work < ApplicationRecord
     end
   end
 
+  sig { params(first_interval: ActiveSupport::Duration, subsequent_interval: ActiveSupport::Duration).void }
+  def self.send_first_draft_reminders(first_interval: nil, subsequent_interval: nil)
```
jcoyne, line 92: Can you move all of this to a service object (or background job)?  A models responsibility is about persistence, and this code is about sending mails.

jcoyne (changes requested):
I believe this also needs to handle version_drafts too.

app/mailers/works_mailer.rb
jmartin-sul: will do...

app/models/work.rb
jmartin-sul: ah yeah, will do, had intended to do that and then forgot to refactor...

jmartin-sul:
@jcoyne i think i've addressed all the suggestions.  had to re-gen `.rbi` files to get the type-checking to pass after i extracted `send_first_draft_reminders` and `send_first_draft_reminder?` to a service class.  sorry for the churn, stuck it in a commit of its own.

app/services/work_reminder_generator.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+  sig do
+    params(work: Work, first_interval: ActiveSupport::Duration, subsequent_interval: ActiveSupport::Duration)
+      .returns(T::Boolean)
+  end
+  def self.send_first_draft_reminder?(work, first_interval, subsequent_interval)
+    today = DateTime.now.utc.beginning_of_day
+    first_reminder_day = work.created_at.utc.beginning_of_day.to_datetime + first_interval
```
jcoyne, line 33: Should this be `updated_at`?  Since we're looking at a time since they last touched this item.  Is there a reason we can't run this query in the database (e.g. show me the works that are in a draft state and haven't been touched for 2 weeks?)

jcoyne (changes requested):

app/services/work_reminder_generator.rb
jmartin-sul: Hmm... so I was initially using `created_at` when I thought this only applied to the `:first_draft` state.  But additionally, I was also under the impression that the reminder interval should be based on when the work went _into_ the draft state, and not when it was last edited.

But looking back at the issue and the spec it links, that now seems ambiguous to me, and I don't think my assumption there was clearly founded.  @jcoyne or @amyehodge, could you clarify whether the desired behavior is for the notification interval to be based on how long the work has been in draft state, vs basing it on the last edit?  Even if it's the former (how long it's been in draft state), I think I could join to `events` to get that info, instead of looking at ~~`created_at`~~`updated_at`.  But, now that I'm not just looking at things in `:first_draft`, using `created_at` does clearly feel wrong no matter what.

@jcoyne, I had briefly considered limiting by date in the query, but I didn't have a strong preference about that when mulling it over.  After getting into the implementation, it seemed a bit easier to do the date math that would let me get everything from an entire given day in the Ruby code instead of in the query.  But it would also be easy enough to filter in the query with a cutoff that's one day later than it should be, and then to keep the same filtering in the Ruby code.  Or, I could just do the date math in SQL afterall 🙂   Do you have a strong preference on this front?

jmartin-sul: oh actually, i suppose i could also just calculate the cutoff date in ruby code and pass that into the query.  so yeah, easy enough to do some date filtering in the query, if you'd prefer.

jcoyne: It's not a big deal now, but I can imagine there might be thousands of objects in a draft states, so a sql query would prevent us from having to "fix" this down the road.

jmartin-sul: yeah, i didn't have a great sense of what the expected scale would be, so i wasn't sure how important it'd be to optimize that part of things.  will fix the query.

app/services/work_reminder_generator.rb
```diff
@@ -0,0 +1,40 @@
...
+  # by a cron job, but the optional manual override for notification interval is useful in case a
+  # human has to manually call the method to send dropped notifications, e.g. in the event that
+  # something goes awry with the cron job on a particular day. For example, if the cron job got
+  # wedged 2 days ago, you could call this method with the the default values plus 2.
+  sig { params(first_interval: ActiveSupport::Duration, subsequent_interval: ActiveSupport::Duration).void }
+  def self.send_first_draft_reminders(first_interval: nil, subsequent_interval: nil)
+    first_interval ||= Settings.notifications.first_draft_reminder.first_interval.days
+    subsequent_interval ||= Settings.notifications.first_draft_reminder.subsequent_interval.days
+
+    Work.with_state(:first_draft, :version_draft).find_each do |work|
```
jcoyne, line 19: Sorry, It looks like `version_draft` should be part of #592

jmartin-sul: ah ok, will back that change out, no worries

jmartin-sul:
After playing with it a bit more and looking up the applicable date functions, I ended up pushing all the selection logic into the query.  For whatever reason, I was thinking it'd be uglier, with more casting or formatting, but the SQL version is actually a lot more concise than the ruby version was.  To support that shift, I indexed `works.created_at` and `works.updated_at`.  I left that change in its own commit (only required a very minor test adjustment, since passing the intervals in as `Integer`s started to seem more sensible than using `ActiveSupport::Duration`s, since the SQL wants ints anyway).

jmartin-sul:
I also tacked on a commit that switched back to using `created_at`, since that was easy to do now that `:first_draft` is the only state in consideration.  It seemed to make sense given the length of the notification interval.  But it's super easy to revert back to using `updated_at`, should that be the way to go.

jcoyne (approved):

jcoyne:
@jmartin-sul You have a small conflict in Gemfile.lock to resolve, then please merge.

PR #764, opened 2020-12-11
Send a mail when deposit access is revoked

jcoyne:
## Why was this change made?

For consistency with where other mails are sent from.

Test suite.

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -47,9 +47,6 @@ class WorksController < ObjectsController
 
     @form = work_form(work)
     if @form.validate(work_params) && @form.save
-      work.event_context = { user: current_user }
-      work.update_metadata!
-
       after_save(work)
     else
       # Send form errors to client in JSON format to be parsed and rendered there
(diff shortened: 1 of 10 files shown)


app/controllers/works_controller.rb
```diff
@@ -48,8 +48,6 @@ def update
     @form = work_form(work)
     if @form.validate(work_params) && @form.save
       work.event_context = { user: current_user }
```
mjgiarlo, line 50: Should this line also move into `#after_save` as in the `CollectionsController`?

mjgiarlo (approved):

spec/requests/edit_collection_spec.rb
```diff
@@ -3,87 +3,14 @@
...
-  let(:collection) { create(:collection) }
+RSpec.describe 'Updating an existing collection' do
   let(:deposit_button) { 'Deposit' }
   let(:save_draft_button) { 'Save as draft' }
 
   before do
     allow(Settings).to receive(:allow_sdr_content_changes).and_return(true)
   end
 
-  context 'with unauthenticated user' do
```
jcoyne, line 15: These tests moved to spec/requests/create_collection_spec.rb

PR #775, opened 2020-12-14
Allow collection values to be nil 

jcoyne:
## Why was this change made?

We want to allow drafts to be saved without requiring everything.


mjgiarlo (approved):
And we want these to help with migrations, which we don't expect to use form objects, hence the validations being defined on the model rather than the form? Assuming that, and approving.

PR #817, opened 2021-01-11
[DRAFT] Adds sorting and paging to the collection deposits page

aaron-collier:
## Why was this change made?

Screen Shot 2021-01-11 at 1 45 39 PM

TDB

N/A


app/views/layouts/application.html.erb
```diff
@@ -9,8 +9,14 @@
 
     <%= stylesheet_link_tag "https://www-media.stanford.edu/su-identity/css/su-identity.css" %>
     <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700" %>
+    <%= stylesheet_link_tag "https://cdn.datatables.net/1.10.23/css/dataTables.bootstrap5.min.css" %>
     <%= stylesheet_pack_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <script src="https://code.jquery.com/jquery-3.5.1.js"></script>
```
jcoyne, line 15: Please don't add jquery. It's a very big library and we don't need to add all that weight just to sort some tables.  It's not very challenging to just roll your own here.

jcoyne: take a look at https://medium.com/javascript-in-plain-english/easy-table-sorting-with-javascript-370d8d97cad8

aaron-collier: Thanks for the tip, that was going to be my next question based on something I read today.

Issue #786, opened 2021-01-07
Attempt to implement hotwire for form validation


peetucket:
Spike to see if hotwire is a good improvement for form validations.  Try to implement on item deposit page.

https://hotwire.dev/

mjgiarlo:
# Spike report

This PR begins by [installing the hotwire-rails gem](https://github.com/sul-dlss/happy-heron/pull/820/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55fR59), the rationale for which is above. The hotwire-rails gem rolls up two other gems, turbo-rails and stimulus-rails. These contain the javascript that's needed on the client side to make the hotwire approach functional and they also contain handy helper methods for using in views, models, controllers, etc. 

A potential downside is that they require sprockets for the asset pipeline: H2 doesn't currently use sprockets for assets (it's pure webpacker). No problem, it's easy to enable [sprockets](https://github.com/sul-dlss/happy-heron/pull/820/files#diff-c1fd91cb1911a0512578b99f657554526f3e1421decdb9e908712beab57e10f9R16) for [JavaScript assets](https://github.com/sul-dlss/happy-heron/pull/820/files#diff-12fd523cb95d119382c3743591e40c90ae62ed20b24a7012b37aed54031009b8R3).

When sprockets is turned on for JavaScript assets, the Stimulus controllers that fire in the feature specs stop working. No big problem: we can copy/move the Stimulus controllers from `app/javascript/controllers/` to `app/assets/javascripts/controllers/`.

Apparently the current stimulus-rails autoloader doesn't know how to deal with Stimulus controllers that have dashes in their name ([pending PR 1](https://github.com/hotwired/stimulus-rails/pull/3) [pending PR 2](https://github.com/hotwired/stimulus-rails/pull/11)). At any rate, we can patch that pretty easily in https://github.com/hotwired/stimulus-rails/blob/v0.1.2/app/assets/javascripts/stimulus/loaders/autoloader.js#L10 by changing that line to:
```javascript
     let controllerFilename = `${controllerName.replaceAll('-', '_')}_controller`
```

At this point, the Stimulus controllers still aren't wired up correctly as target fields cannot be found:

```
error: unable to find target for subtypeField                                                                                                         
error: unable to find target for subtypeField                                                                                                         
error: unable to find target for abstractField                                                                                                        
error: unable to find target for contact_emailField                                                                                                   
error: unable to find target for abstractField                                                                                                        
[...]

PR #820, opened 2021-01-12
[DRAFT] [SPIKE] Use hotwire for work form validation

mjgiarlo:
Note that installation requires the asset pipeline, which requires the sprockets railtie. Why not just install via `yarn` and rely on webpacker? We could, but then we would not get any of the hotwire helpers that come from installing via the hotwire-rails gem, e.g., `broadcasts_to`, `turbo_frame_tag`, `turbo_stream_action_tag`.

Note as well that we should ultimately be able to replace remote forms entirely with the hotwire approach, hence the new TODO and commented-out configuration option in `config/application.rb`.

Connects to #786

## Why was this change made?

To experiment with using hotwire to improve H2 form validation, in terms of user experience and developer understanding.

CI

None


mjgiarlo:
Closing. See rationale: https://github.com/sul-dlss/happy-heron/issues/786#issuecomment-759052241

PR #822, opened 2021-01-12
[DRAFT] Add basic table sorting and disable unsortable columns

aaron-collier:
## Why was this change made?

Option #2 for table sorting based on comments in other PR.

Default display without any sort indicators:

Screen Shot 2021-01-12 at 3 17 16 PM

Once a heading has been clicked showing a sort indicator:

Screen Shot 2021-01-12 at 3 17 22 PM

And clicked again, the indicator changes direction:

Screen Shot 2021-01-12 at 3 17 28 PM

One area where I am currently blocked is the ability to clear the existing sort indicator without causing problems with the event due to reloading of the dom (see: https://stackoverflow.com/questions/5113105/manipulating-innerhtml-removes-the-event-handler-of-a-child-element):

Screen Shot 2021-01-12 at 3 17 35 PM

### DIFF
--- a/app/components/collections/works_component.html.erb
+++ b/app/components/collections/works_component.html.erb
@@ -1,11 +1,11 @@
-<table class="table sortable">
+<table id="works" class="table sortable">
 <thead class="table-light">
-  <tr>
-    <th><%= I18n.t 'collection.deposits' %></th>
+  <tr class="work-headers">
+    <th width="18%"><%= I18n.t 'collection.deposits' %></th>
     <th><span class="visually-hidden"><%= I18n.t 'collection.actions' %></span></th>
-    <th><%= I18n.t 'collection.depositor' %></th>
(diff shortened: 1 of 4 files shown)


mjgiarlo:
Depending on how you're feeling about the current approach and trajectory, we may want to consider this datatables library that does not use JQuery: https://github.com/fiduswriter/Simple-DataTables I assume it's not as big as JQuery, but before using this, we should check to see how much it'd inflate our overall page size. 

If we *do* want all custom code around table sorting, though, I wonder if using Stimulus might make it easier to manipulate individual column headers as named elements without surprises? Maybe, maybe not.

Do you have druthers, @aaron-collier @jcoyne ?

aaron-collier:
I like the idea of Simple-DataTables, and happy to put that forward if we want that approach, perhaps short discussion at standup?

PR #819, opened 2021-01-11
bold and indent rejection reason in email

peetucket:
## Why was this change made?

Fixes #722 - format email change

## How was this change tested?

In browser

## Which documentation and/or configurations were updated?

N/A

PR #823, opened 2021-01-12
Display units for file sizes

### LINKED ISSUE #770: add units to Deposit Size column in Collection Deposits table
I assume these are in bytes, so can this read "Deposit size (bytes)"? 

Even better would be if the numbers were made more readable and said 11kb, 28MB, 2GB, etc as opposed to having a standard unit at the top. 

E.g. https://sul-h2-stage.stanford.edu/collections/105/works

jcoyne:
## Why was this change made?
Fixes #770
Screen Shot 2021-01-12 at 12 16 54 PM

### DIFF
--- a/app/components/collections/work_row_component.rb
+++ b/app/components/collections/work_row_component.rb
@@ -1,0 +1,23 @@
+# typed: false
+# frozen_string_literal: true
+
+module Collections
+  # Displays a single table row representing a work in a collection.
+  class WorkRowComponent < ApplicationComponent
+    with_collection_parameter :work
+
+    def initialize(work:)
+      @work = work
+    end
+
+    attr_reader :work
+
+    # Returns the size of the attached files
+    def size
+      number_to_human_size(attached_files.sum(&:byte_size))
+    end
+
+    delegate :attached_files, to: :work
+    delegate :allowed_to?, to: :helpers
+  end
+end
--- a/spec/components/collections/work_row_component_spec.rb
+++ b/spec/components/collections/work_row_component_spec.rb
@@ -1,0 +1,16 @@
+# typed: false
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Collections::WorkRowComponent, type: :component do
+  pending "add some examples to (or delete) #{__FILE__}"
+
+  # it "renders something useful" do
+  #   expect(
+  #     render_inline(described_class.new(attr: "value")) { "Hello, components!" }.css("p").to_html
+  #   ).to include(
+  #     "Hello, components!"
+  #   )
+  # end
+end
--- a/app/components/collections/work_row_component.html.erb
+++ b/app/components/collections/work_row_component.html.erb
@@ -1,0 +1,18 @@
+<tr>
+  <td class="work-title"><%= link_to Works::DetailComponent.new(work: work).title, work %></td>
+  <td>
+    <% if allowed_to? :edit?, work %>
+      <%= link_to [:edit, work], aria: { label: "Edit #{Works::DetailComponent.new(work: work).title}" } do %>
+        <span class="fas fa-pencil-alt"></span>
+      <% end %>
+    <% end %>
+    <%= render Works::DeleteComponent.new(work: work) %>
+  </td>
+  <td><%= work.depositor.sunetid %></td>
+  <td><%= render Works::StateDisplayComponent.new(work: work) %></td>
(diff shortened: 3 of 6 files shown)


app/components/collections/work_row_component.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def initialize(work:)
+      @work = work
+    end
+
+    attr_reader :work
+
+    # Returns the size of the attached files
+    def size
+      number_to_human_size(attached_files.sum(&:byte_size))
```
mjgiarlo, line 17: I guess you don't think we need to change the tests to ensure sizes are displayed in the format that `number_to_human_size` returns?

jcoyne: not really.  If that were to change, most of the integration tests would fail, right?

spec/components/collections/work_row_component_spec.rb
```diff
@@ -0,0 +1,16 @@
+# typed: false
```
aaron-collier, line 1: Do we want this blank test?

mjgiarlo: We already have many such blank component tests, unfortunately, so this is at least consistent with current practice.

app/components/collections/work_row_component.rb
mjgiarlo: I don't know. This line is the main change in this PR, right? I guess I'm surprised we needed to add it and it didn't change test behavior.

jcoyne: Do we need a test for every last detail?  We could write tests down to the pixel level if we want to, but there are diminishing returns here.  Both before and after we are presenting the same data to the user, this just has a format that is slightly more readable.

mjgiarlo: I'm not advocating for pixel-level testing. What I think we target: test behaviors that our users care about so that we are sure they work as expected and that regressions in those behaviors are prevented (or at least detected). For this PR, despite all the machinery changes, I thought adding `number_to_human_size` was a/the main user-facing change, so I thought it reasonable to ask about whether we're testing it. If it's already tested by feature specs or integration tests, great (though if they were already passing, why/how was this change needed?). If not, then in an instance like this when I don't see a test for the main behavioral change, I can only assume that means we don't care too much if it regresses. True/false?

aaron-collier: It seems like this could be a point of possible regression like we recently saw in google books with the output of `average` changing, and a test caught that. While it should be a rather simple test, we probably want to at least have a way to catch that kind of change.

jcoyne: @aaron-collier What I hear you saying is that we should test that the libraries we are using are not changing behavior.  That is not my understanding of what tests should be for.  The contract is defined by the API and it is up to the library maintainers not to break such contracts.

jcoyne: I see this change as akin to #819, which also affects behavior but needs no tests.

aaron-collier: @jcoyne Thinking about it, I think you're right. In the case of `average` we were tasting the output of our code and the result just happened to change. In this case it would be a test for specifically for `number_to_human_size` and not something we're doing with the result. Good point.

mjgiarlo: The point to my mind was not to test ActiveSupport here; I trust the Rails maintainers to make sure their code is relatively well tested. The point, rather, is that we have a user-facing requirement that file sizes display in a more human-readable form vs. a bare integer, and there's a *lot* of machinery required to satisfy that requirement. The machinery itself is well tested but the way we compose all of this machinery to satisfy this particular requirement may not be.

We do refactorings all the time and it's not unheard of for us to introduce bugs in our refactorings, particularly when requirements are not tested (in e.g., a feature spec). What if in our next refactoring of these components, someone winds up yanking the `number_to_human_size` call? In this case, it would be sufficient to toss a one-line `expect` into an existing feature spec to make sure the size renders in human-readable form. I don't *think* that suggests we need to test every pixel or test every API in every library used in H2.

mjgiarlo: I'm open to being convinced that I am off-base or overthinking this, too.

At any rate, this is a small PR with a change we want to see in sprint 0, so let's commit to resolving this one way or the other today?

jcoyne: added a test.

mjgiarlo (approved):

PR #825, opened 2021-01-12
Be informed of state changes by RabbitMQ

jcoyne:
## Why was this change made?

Testing the suitability of RabbitMQ for our purposes. 

### Findings:
This seems to be a reasonable way of getting the events we need. We will need to add https://github.com/sul-dlss/cocina-models/issues/202 if we want this to work for collections.  We should also get rid of sidekiq and use Sneakers to deal with the DepositJob and DepositCollectionJob.


jermnelson (approved):
LGTM; Will a follow-up PR update `docker-compose.yml` with a RabbitMQ instance?

justinlittman (approved):

PR #842, opened 2021-01-14
Allow for multiple Contact Emails

### LINKED ISSUE #731: Allow for entry of more than one email address.
Currently, I can only enter one email address. We had hoped to make it possible for more than one to be entered to increase the chance of one of them being an active email for a longer period of time in case someone in the future has questions. 

I realize this was not explicitly spelled out in the spec, though it is in the tooltip text. Is this possible, or does it make it impossible to do the validation? 

Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.r28y1sknl4gx

Note: this should be possible by any user in both the item deposit form and the collection deposit form for the contact fields.

jermnelson:
## Why was this change made?
Fixes #731 

New **+ Add another email** button and row to Collection Form:
Screen Shot 2021-01-20 at 9 26 44 AM

New **+ Add another email** button and row to Work Form:
Screen Shot 2021-01-20 at 9 24 50 AM

Testing suite and new tests

n/a


app/forms/contact_emails_populator.rb
```diff
@@ -0,0 +1,16 @@
...
+
+# Populates the nested contact email form
+class ContactEmailsPopulator < ApplicationPopulator
+  # The fragment represents one row of the attached file data from the HTML form
+  def call(form, fragment:, **)
+    item = existing_record(form: form, id: fragment['id'])
+
+    if fragment['_destroy'] == '1'
+      form.contact_emails.delete(item)
+      return skip!
```
mjgiarlo, line 12: It looks like these two lines may have caused the coverage drop.

jermnelson: Thanks @mjgiarlo! Adjusted test for `contact_emails_populator_spec.rb` and now passes with specific test for removal.

app/forms/draft_collection_form.rb
```diff
@@ -33,6 +32,13 @@ class DraftCollectionForm < Reform::Form
     self.reviewer_sunets = reviewer_sunets_from_model.join(', ')
   }
 
+  collection :contact_emails, populator: ContactEmailsPopulator.new(:contact_emails, ContactEmail),
+                              prepopulator: ->(*) { contact_emails << ContactEmail.new if contact_emails.blank? } do
+    property :id
+    property :email
+    property :_destroy, virtual: true
```
jermnelson, line 39: Add similar `{ with: Devise.email_regexp }` as draft_work_form

app/forms/work_form.rb
```diff
@@ -9,7 +9,6 @@ class WorkForm < DraftWorkForm
   validates 'release', presence: true, inclusion: { in: %w[immediate embargo] }
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
   validates :attached_files, length: { minimum: 1, message: 'Please add at least one file.' }
-  validates :contact_email, presence: true, format: { with: Devise.email_regexp }
```
jermnelson, line 12: Should validate presence of at least one `contact_emails`, length validator

spec/services/request_generator_spec.rb
```diff
@@ -82,12 +82,12 @@
               {
                 value: 'test citation',
                 type: 'preferred citation'
-              },
-              {
-                value: 'io@io.io',
-                type: 'contact',
-                displayLabel: 'Contact'
               }
+              # {
```
jermnelson, line 86: Remove cruft

spec/services/request_generator_spec.rb
```diff
@@ -139,12 +139,12 @@
...
```
jermnelson, line 143: Remove cruft

app/components/works/title_component.html.erb
```diff
@@ -2,7 +2,7 @@
    <header>Title of deposit and contact information</header>
    <div class="mb-3 row">
      <%= form.label :title, 'Title of deposit', class: 'col-sm-2 col-form-label' %>
-     <div class="col-sm-10">
+     <div class="col-sm-9">
```
jcoyne, line 5: Why was this change made?

app/javascript/controllers/nested_form_controller.js
```diff
@@ -11,7 +11,7 @@ export default class extends Controller {
 
   removeAssociation(event) {
     event.preventDefault()
-    const item = event.target.closest(".inner-container")
+    const item = event.target.closest(".inner-container") || event.target.closest(".plain-container")
```
jcoyne, line 14: Can this class to use be passed in as a data- attribute?

app/services/description_generator.rb
```diff
@@ -112,13 +112,15 @@ def published_date
     )
   end
 
-  sig { returns(Cocina::Models::DescriptiveValue) }
+  sig { returns(T::Array[Cocina::Models::DescriptiveValue]) }
   def contact
```
jcoyne, line 116: Can you rename to `contacts`

db/migrate/20210113173329_create_contact_emails.rb
```diff
@@ -0,0 +1,12 @@
+class CreateContactEmails < ActiveRecord::Migration[6.0]
+  def change
+    create_table :contact_emails do |t|
+      t.string :email
+
+      t.timestamps
+    end
+    add_reference :contact_emails, :emailable, polymorphic: true
```
jcoyne, line 8: This can move up into the `create_table` as:
```ruby
t.references :emailable, polymorphic: true
```

spec/features/create_new_work_spec.rb
```diff
@@ -55,6 +56,10 @@
       fill_in 'Title of deposit', with: 'My Title'
       fill_in 'Contact email', with: user.email
 
+      # Adds second contact email
+      click_button '+ Add another email'
+      all('div.plain-container > div > input')[1].set(second_email)
```
jcoyne, line 61: You should try to avoid using selectors in capybara.  There should be a label or aria-label that allows us to use `fill_in`.

spec/forms/contact_emails_populator_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  it 'renders a contact email row' do
+    keep_fragment = ActionController::Parameters.new({ '_destroy': 'false', 'email': 'test@local.edu' })
+    expect(contact_email_populator.call(work_form,
+                                        fragment: keep_fragment)).to be_truthy
+  end
+
+  it 'removes a contact email row' do
+    remove_keep = ActionController::Parameters.new({ '_destroy': '1', 'email': 'test@local.edu' })
+    expect(contact_email_populator.call(work_form,
+                                        fragment: remove_keep).to_s).to eq 'Representable::Pipeline::Stop'
```
jcoyne, line 21: Is there a reason `.to_s` is necessary?

spec/requests/create_work_spec.rb
```diff
@@ -359,7 +379,7 @@
           expect(response).to have_http_status(:found)
           work = Work.last
           expect(work.title).to be_empty
-          expect(work.contact_email).to be_empty
+          expect(work.contact_emails.size).to eq 0
```
jcoyne, line 382: why is this change needed?

spec/requests/create_work_spec.rb
```diff
@@ -398,7 +417,7 @@
...
```
jcoyne, line 420: Why is this change needed?

spec/services/request_generator_spec.rb
```diff
@@ -296,12 +286,12 @@
...
```
jcoyne, line 290: remove?

jcoyne (changes requested):

app/components/works/title_component.html.erb
jermnelson: In order to keep the title input field and the substituent contact email fields the same size. Before the label div was `col-sm-2` and the input div was `col-sm-10`. When a new contact email is added, we needed a `col-sm-1` div for the delete button on the row otherwise the delete button would wrap to the next row because the total number of columns exceeded 12.

spec/requests/create_work_spec.rb
jermnelson: Changed to `be_empty`

spec/requests/create_work_spec.rb
jermnelson: Changed to `be_empty`

spec/forms/contact_emails_populator_spec.rb
jermnelson: Changed to test equality on class `Representable::Pipeline::Stop`

app/javascript/controllers/nested_form_controller.js
jermnelson: I did some experiments with using a data-attribute but part of the problem is we still need to select the containing div from the found item using the CSS.

mjgiarlo: @jcoyne howdy. @jermnelson and I looked at this and the Capybara selector comment you raised, when we paired earlier. We didn't find a lightweight way to arrive at deterministic selectors (e.g., email fields with predictable aria labels based on a simple incremented counter) for individual email fields. 

When discussing both of these comments, we realized that they both tie back to the nested-form stimulus controller, and it might be worth copying/pasting that controller to a new contact-email controller that has more appropriate behavior for `addAssociation()` and `removeAssociation()`. That said, if you have something else in mind, I think Jeremy is all ears---but he's also already invested a bunch of time in this PR, so we may be reaching the point when we ought to consider merging this PR and writing issues for the lingering bits we're not happy with. Thoughts?

spec/features/create_new_work_spec.rb
mjgiarlo: See comment above.

app/javascript/controllers/nested_form_controller.js
jcoyne: I'm saying pass in the selector as a data- attribute.

spec/features/create_new_work_spec.rb
jcoyne: Can we use something like `aria-label="<%= n.ordinalize %> contact"`?

mjgiarlo: @jcoyne can you pair with @jermnelson to help finish this off?

app/javascript/controllers/nested_form_controller.js
jermnelson: Updated to now use Stimulus data attribute instead of css class.

spec/features/create_new_work_spec.rb
jermnelson: Leaving for now as `TEMPLATE_RECORD` along with `new Date().valueOf()` is still used for setting the ID of input in the `nested_form_controller.js`.

jcoyne (approved):

PR #843, opened 2021-01-14
Add simple-DataTable sorting to collection works display

aaron-collier:
## Why was this change made?

This is quite straight forward:

Screen Shot 2021-01-13 at 5 55 54 PM

Manual review

N/A

### DIFF
--- a/app/components/collections/works_component.html.erb
+++ b/app/components/collections/works_component.html.erb
@@ -1,9 +1,9 @@
 <table class="table">
 <thead class="table-light">
   <tr>
-    <th><%= I18n.t 'collection.deposits' %></th>
+    <th width="18%"><%= I18n.t 'collection.deposits' %></th>
     <th><span class="visually-hidden"><%= I18n.t 'collection.actions' %></span></th>
-    <th><%= I18n.t 'collection.depositor' %></th>
+    <th width="10%"><%= I18n.t 'collection.depositor' %></th>
     <th><%= I18n.t 'collection.deposit_status' %></th>
     <th><%= I18n.t 'collection.last_modified' %></th>
     <th><%= I18n.t 'collection.number_of_files' %></th>
@@ -28,10 +28,19 @@
       <td><%= render Works::StateDisplayComponent.new(work: work) %></td>
       <td><%= I18n.l(work.updated_at.to_date, format: :abbr_month) %></td>
       <td><%= work.attached_files.count %></td>
-      <td><%= work.attached_files.sum(&:byte_size) %></td>
+      <td><%= number_to_human_size(work.attached_files.sum(&:byte_size)) %></td>
       <td><%= link_to work.purl, work.purl if work.purl %></td>
       <td><span class="fas fa-quote-left"></span> Cite</td>
     </tr>
   <% end %>
 </tbody>
 </table>
+
+<script src="https://cdn.jsdelivr.net/npm/simple-datatables@2.1.13"></script>
+<script>
+  new simpleDatatables.DataTable("table", {
+    columns: [
+        { select: 0, sort: "asc" },  // Sort the first column in ascending order
+        { select: [1, 5, 6, 7, 8], sortable: false } // Disable unsortable columns
+    ]});
+</script>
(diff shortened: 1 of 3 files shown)


app/components/collections/works_component.html.erb
```diff
@@ -1,9 +1,9 @@
 <table class="table">
 <thead class="table-light">
   <tr>
-    <th><%= I18n.t 'collection.deposits' %></th>
+    <th width="18%"><%= I18n.t 'collection.deposits' %></th>
```
aaron-collier, line 4: This was necessary to avoid overlapping the sort direction indicator.

app/components/collections/works_component.html.erb
```diff
@@ -28,10 +28,21 @@
...
 </tbody>
 </table>
+
+<script>
+  var table = new simpleDatatables.DataTable("table", {
+    columns: [
+        // Sort the first column in ascending order
+        { select: 0, sort: "asc" },
+
+        // Disable sorting on second, and 4th - 7th columns
```
jcoyne, line 45: I'm confused about this comment and how it relates to the next line.

app/views/layouts/application.html.erb
```diff
@@ -9,8 +9,10 @@
 
     <%= stylesheet_link_tag "https://www-media.stanford.edu/su-identity/css/su-identity.css" %>
     <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700" %>
+    <%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/simple-datatables@latest/dist/style.css" %>
     <%= stylesheet_pack_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
     <%= javascript_pack_tag 'application', 'data-turbolinks-track': 'reload' %>
+    <script src="https://cdn.jsdelivr.net/npm/simple-datatables@latest"></script>
```
jcoyne, line 15: Why not just put this in the component itself?

jcoyne (changes requested):

app/components/collections/works_component.html.erb
aaron-collier: `{ select: [...], sortable: false }` is the config to disable sorting on the columns in the select. Happy to drop the comment if it's confusing/misleading.

app/views/layouts/application.html.erb
aaron-collier: That makes sense, I wasn't sure if we wanted this available elsewhere, but it probably makes the most sense to only include it when we need it.

app/components/collections/works_component.html.erb
```diff
@@ -28,10 +28,21 @@
...
+      <td><%= number_to_human_size(work.attached_files.sum(&:byte_size)) %></td>
       <td><%= link_to work.purl, work.purl if work.purl %></td>
       <td><span class="fas fa-quote-left"></span> Cite</td>
     </tr>
   <% end %>
 </tbody>
 </table>
+
+<script>
+  var table = new simpleDatatables.DataTable("table", {
```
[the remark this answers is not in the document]
aaron-collier: Copied from the instructions. I can remove.

app/views/layouts/application.html.erb
```diff
@@ -9,6 +9,7 @@
 
     <%= stylesheet_link_tag "https://www-media.stanford.edu/su-identity/css/su-identity.css" %>
     <%= stylesheet_link_tag "https://fonts.googleapis.com/css?family=Source+Sans+Pro:300,400,700" %>
+    <%= stylesheet_link_tag "https://cdn.jsdelivr.net/npm/simple-datatables@latest/dist/style.css" %>
```
jcoyne, line 12: Do we want to pin to a specific version so that changes are managed by us or are you confident there will never be breaking changes with `@latest`?

aaron-collier: Pinning seems appropriate as the most recent tagged version is from June and doesn't change our expected behavior. Done.

jcoyne: I see you pinned the version of the JS but not of the CSS

aaron-collier: I clearly needed more coffee. Done.

jcoyne (approved):

PR #844, opened 2021-01-14
Add a notification when there is activity in a collection

### LINKED ISSUE #602: Email notifications: "Collection Manager: Anyone creates a draft, publishes, opens a new item"
https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#heading=h.w8tadvk8afo3

Example implementation: https://github.com/sul-dlss/happy-heron/pull/673

jcoyne:
## Why was this change made?
Fixes #602


app/models/work.rb
```diff
@@ -26,7 +26,7 @@ class Work < ApplicationRecord
     world: 'world'
   }
 
-  state_machine initial: :first_draft do
+  state_machine initial: :initial do
```
mjgiarlo, line 29: The state machine changes seem like a key part of this PR. Can you say more about how changing the initial state factors in here?

mjgiarlo, line 29: Can you regenerate the state machine diagrams as part of this PR?

Per https://github.com/sul-dlss/happy-heron/wiki/State-machine-diagrams, you can do that via `bin/rake state_machines:draw CLASS=Work,Collection FORMAT=svg`

jcoyne: We need to trigger an action when a draft is made, but there's no unique transition we can hook into without creating a new initial state.

jcoyne: I'm working on installing graphviz now...it's been running for a long time.

mjgiarlo: @jcoyne OK, thanks!

mjgiarlo: Fun. I can toss a commit at this branch if you'd prefer.

jcoyne: That would be fantastic.

mjgiarlo: @jcoyne done

mjgiarlo: @jcoyne last question: should we also make the initial state of the collection `initial` for symmetry?

mjgiarlo: @jcoyne [`3fd31d6` (#844)](https://github.com/sul-dlss/happy-heron/pull/844/commits/3fd31d60b107e0a9a20f72b6969da76e19b2729e)

mjgiarlo (approved):
LGTM, but also one lingering question: should we also make the initial state of the collection initial for symmetry?

app/models/work.rb
jcoyne: I've updated so that Collection has the same initial state as Work

mjgiarlo (approved):

PR #845, opened 2021-01-14
Allow empty collections to appear in All Collections

### LINKED ISSUE #776: All Collections does not show collections without any deposits
It should show all of the collections regardless of the number of deposits.

mjgiarlo:
Fixes #776

## Why was this change made?

This refactoring of StatBuilder allows all collections to render in the All Collections component within the dashboard. To my eyes, the class is now also easier to grok.

CI

None

### DIFF
--- a/app/services/stat_builder.rb
+++ b/app/services/stat_builder.rb
@@ -1,33 +1,24 @@
-# typed: true
+# typed: strict
 # frozen_string_literal: true
 
 # Runs queries and builds the data structure for the summary table.
 class StatBuilder
   extend T::Sig
 
-  sig { returns(T::Hash[Collection, Hash]) }
+  sig { returns(T::Hash[Collection, T::Hash[String, Integer]]) }
   def self.build_stats
-    collections_by_id = Collection.all.index_by(&:id)
+    works = Work.all.group(:collection_id, :state).count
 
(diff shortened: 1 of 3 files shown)


app/services/stat_builder.rb
```diff
@@ -1,33 +1,16 @@
...
-  end
-  private_class_method :by_state
-
-  sig { returns(T::Hash[Integer, Integer]) }
-  def self.total
-    Work.all.group(:collection_id).count.each_with_object({}) do |(id, count), obj|
-      obj[id] = count
+    Collection.all.index_with do |collection|
+      collection.works.group(:state).count.merge(
+        'total' => collection.works.count
```
jcoyne, line 12: This causes `2N+1` query where before we had 3 queries.  Can you find a solution where there are only 3 queries?

jcoyne (changes requested):

app/services/stat_builder.rb
mjgiarlo: @jcoyne You bet.

mjgiarlo: @jcoyne Got it down to 2 total queries. Pushing in a minute.

jcoyne (approved):

PR #851, opened 2021-01-14
add cancel button to buttons component

### LINKED ISSUE #794: Add "Discard draft" option to Collection Edit page
wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930437
design doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.z1sansuktydz

Update: Cancel button now in #857

### LINKED ISSUE #857: Add "Cancel" options to Collection/Work Edit page
wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930437
design doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.z1sansuktydz

The "Cancel" button

    cancels out of the data entry form without saving anything

    1) if the item had never been saved, then nothing is saved here either, ie. there is no item listed in tables on the dashboard or elsewhere and no trace of it should remain.

    2) if the item had previously been saved, then the item returns to the last saved state and none of the data that was entered into the form (if any) is saved.

    in either case, returns the user to the collection view page

peetucket:
## Why was this change made?

Fixes #857 - add back button to edit collection/work pages

Localhost browser

N/A

### DIFF
--- a/app/components/buttons_component.html.erb
+++ b/app/components/buttons_component.html.erb
@@ -3,6 +3,9 @@
     <%= delete_button %>
   </div>
   <div class="col-auto">
+    <%= cancel_button %>
+  </div>
+  <div class="col-auto">
     <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button' %>
     <%= form.submit submit_button_label, class: 'btn btn-primary' %>
   </div>
--- a/app/components/buttons_component.rb
+++ b/app/components/buttons_component.rb
@@ -23,6 +23,15 @@ class ButtonsComponent < ApplicationComponent
     end
   end
 
+  sig { returns(T.nilable(String)) }
+  def cancel_button
+    if model_type == 'Work'
+      render Works::CancelComponent.new(work: object.model)
+    else
+      render Collections::CancelComponent.new(collection: object.model)
+    end
+  end
+
   private
 
   delegate :object, to: :form
--- a/app/components/collections/cancel_component.html.erb
+++ b/app/components/collections/cancel_component.html.erb
@@ -1,0 +1,1 @@
+<%= link_to 'Cancel', dashboard_path %>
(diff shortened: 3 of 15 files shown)


app/components/buttons_component.html.erb
```diff
@@ -2,6 +2,9 @@
   <div class="col-auto">
     <%= delete_button %>
   </div>
+  <div class="col-auto">
+    <%= link_to 'Cancel', :back %>
```
jcoyne, line 6: What should this do if they've navigated directly to this page via an email? Then they won't have a back in their browser history to go to?

jcoyne (changes requested):

app/components/buttons_component.html.erb
peetucket: Hmm, yes, good point.  I did some research and Rails uses the referrer if one exists, and then tries to revert to javascript's history if not.  So if you open a new tab directly to the work/collection edit page, the cancel button takes you to the blank browser window.  Assuming you are logged in and the email link gets you a new tab.

jcoyne: Right, I don't think that's acceptable, especially since we're sending this link in an email. I think it needs an explicit route.

jcoyne (changes requested):

app/components/buttons_component.html.erb
peetucket: Ok, how about this approach, which sets the back button location based on model type:

Collection --> back to dashboard
Work --> back to collection details page listing the works

Thoughts?  Also, worth doing a test for?

jcoyne (approved):

PR #852, opened 2021-01-15
[DRAFT] Select current values when rendering the form

### LINKED ISSUE #771: "On this date" should automatically be selected when editing items that previously had embargo date entered.
I opened this item that previously had an embargo date selected. The embargo date is still there, but the radio button selected is for "Immediately." The "On this date" radio button is the one that should be selected here, as it should have been selected when the embargo date was added. 

See https://sul-h2-stage.stanford.edu/works/116/edit

mjgiarlo:
Fixes #771

## Why was this change made?

It appears the app currently lacks code to select the current value for the embargo date, the license, and the created date range. This PR fixes all three. I can now toggle back and forth with expected results.

CI (forthcoming, haven't written tests yet)

None

### DIFF
--- a/app/forms/embargo_date_params_filter.rb
+++ b/app/forms/embargo_date_params_filter.rb
@@ -9,7 +9,7 @@ class EmbargoDateParamsFilter
       next unless dfn[:embargo_date]
 
       name = dfn[:name]
-      date_attributes[name] = deserialize(params, name) if dfn[:assign_if].call(params)
+      date_attributes[name] = params['release'] == 'embargo' ? deserialize(params, name) : nil
     end
 
     params.merge(date_attributes)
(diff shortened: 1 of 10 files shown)


app/forms/embargo_date_params_filter.rb
```diff
@@ -9,7 +9,7 @@ def call(schema, params)
       next unless dfn[:embargo_date]
 
       name = dfn[:name]
-      date_attributes[name] = deserialize(params, name) if dfn[:assign_if].call(params)
+      date_attributes[name] = params['release'] == 'embargo' ? deserialize(params, name) : nil
```
mjgiarlo, line 12: This wipes out the embargo date for a work that used to have an embargo date and then flipped to immediate release.

PR #853, opened 2021-01-15
small CollectionChangeSet refactor

jmartin-sul:
## Why was this change made?

slightly clearer code, hopefully?  in particular, i was briefly confused by the fact that `CollectionChangeSet.from` returned a `PointInTime` (i expected a `CollectionChangeSet` based on the method name).

hopefully the type annotations, and the method and var renaming make things a bit more obvious.

type checker and unit tests

type annotations

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -39,10 +39,10 @@ class CollectionsController < ObjectsController
   def update
     collection = Collection.find(params[:id])
     authorize! collection
-    change_set = CollectionChangeSet.from(collection)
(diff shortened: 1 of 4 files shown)


jcoyne (approved):
Awww, I liked the old way....but I can see how this makes more sense.

jmartin-sul:
> Awww, I liked the old way....but I can see how this makes more sense.

😄 thanks for the approval!  yeah, the old way used fewer characters and i agree there was something aesthetically pleasing about the `.from` method.  but hoping the clarity is worth the trade-off.

PR #854, opened 2021-01-15
Extract WorkObserver from work

jcoyne:
## Why was this change made?

This contains business logic that is triggered when state transitions occur. Business logic should not be in the model class.

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -27,49 +27,20 @@ class Work < ApplicationRecord
   }
 
   state_machine initial: :new do
-    before_transition do |work, transition|
-      work.events.build(work.event_context.merge(event_type: transition.event))
-    end
+    before_transition WorkObserver.method(:before_transition)
 
-    after_transition on: :begin_deposit do |work, _transition|
-      DepositJob.perform_later(work)
-    end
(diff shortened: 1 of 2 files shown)


app/models/work.rb
```diff
@@ -27,49 +27,20 @@ class Work < ApplicationRecord
...
-      work.events.build(work.event_context.merge(event_type: transition.event))
-    end
+    before_transition WorkObserver.method(:before_transition)
 
-    after_transition on: :begin_deposit do |work, _transition|
-      DepositJob.perform_later(work)
-    end
+    after_transition WorkObserver.method(:after_transition)
+    after_transition on: :begin_deposit, do: WorkObserver.method(:after_begin_deposit)
+    after_transition on: :reject, do: WorkObserver.method(:after_rejected)
```
mjgiarlo, line 34: Trivial question: are you open to renaming `after_rejected` to `after_reject` for symmetry/consistency with the other method names?

mjgiarlo (approved):

app/models/work.rb
jcoyne: Sure, no strong feelings on this one.

PR #859, opened 2021-01-15
Add review access removed notification mailer

### LINKED ISSUE #600: Email notifications: "Reviewer: Removed from collection as a Reviewer"
https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#

aaron-collier:
## Why was this change made?

Fixes #600 by adding the notification for removal of reviewers,

Unit

### DIFF
--- a/spec/services/collection_observer_spec.rb
+++ b/spec/services/collection_observer_spec.rb
@@ -109,5 +109,19 @@ RSpec.describe CollectionObserver do
         )
       end
     end
+
+    context 'when reviewers are removed from a collection' do
+      let(:collection_after) { collection.dup.tap { |col| col.reviewers = [collection.reviewers.first] } }
+      let(:collection) do
+        create(:collection, :deposited, :with_reviewers)
+      end
+
+      it 'sends emails to those removed' do
+        expect { action }.to have_enqueued_job(ActionMailer::MailDeliveryJob).with(
+          'CollectionsMailer', 'review_access_removed_email', 'deliver_now',
+          { params: { user: collection.reviewers.last, collection: collection }, args: [] }
+        )
+      end
+    end
   end
 end
--- a/app/mailers/collections_mailer.rb
+++ b/app/mailers/collections_mailer.rb
@@ -30,6 +30,13 @@ class CollectionsMailer < ApplicationMailer
       'collection in the SDR')
   end
 
+  def review_access_removed_email
+    @user = params[:user]
+    @collection = params[:collection]
+    mail(to: @user.email, subject: "Your permissions have changed for the #{@collection.name} " \
+      'collection in the SDR')
+  end
+
   def collection_activity
     @user = params[:user]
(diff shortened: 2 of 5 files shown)


jcoyne:
Needs a rebase

db/migrate/20210115231123_add_email_reviewers_status_changed_to_collection.rb
```diff
@@ -0,0 +1,5 @@
+class AddEmailReviewersStatusChangedToCollection < ActiveRecord::Migration[6.0]
+  def change
+    add_column :collections, :email_reviewers_status_changed, :boolean, null: true
```
jcoyne, line 3: I'm not sure I understand why this was added.

jcoyne (changes requested):

db/migrate/20210115231123_add_email_reviewers_status_changed_to_collection.rb
aaron-collier: Following this: https://github.com/sul-dlss/happy-heron/blob/main/db/structure.sql#L159 ..

jcoyne: Screen Shot 2021-01-19 at 9 53 04 AM
That is there to drive this UI element.

aaron-collier: Won't I need to do something similar: https://github.com/sul-dlss/happy-heron/blob/main/app/services/collection_observer.rb#L29  ?

jcoyne: I don't think so.

spec/services/collection_observer_spec.rb
```diff
@@ -109,5 +109,19 @@
         )
       end
     end
+
+    context 'when reviewers are removed from a collection' do
+      let(:collection_after) { collection.dup.tap { |col| col.managers = [collection.reviewers.first] } }
```
jmartin-sul, line 114: should this be `col.reviewers = [collection.reviewers.first]` (instead of `col.managers = ...`)?  or am i confused?  (if i'm not off-base, i'm slightly surprised by the `review_access_removed_email` job getting queued, since i can't see from the rest of the code how that'd happen without a reviewer getting removed, hence my uncertainty)

jmartin-sul (commented):
one question, i might just be a bit confused, otherwise LGTM

spec/services/collection_observer_spec.rb
aaron-collier: yes, typo... I was working on both. Will fix, thanks for catching.

jmartin-sul (approved):

jcoyne (approved):

PR #864, opened 2021-01-15
Render work form creation & embargo date radio buttons based on current values

### LINKED ISSUE #771: "On this date" should automatically be selected when editing items that previously had embargo date entered.

mjgiarlo:
Fixes #771

## Why was this change made?

To fix a bug in deployed environments that could change values previously set by the user.

Also includes:
* Disentangle embargo and contributor validation in work form test 
  * And add a new factory for valid works for more descriptive factory names in specs

CI and in the browser locally and in QA

None

### DIFF
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -25,8 +25,10 @@ class DraftWorkForm < Reform::Form
   end)
   property :created_edtf, edtf: true, range: true
   property :published_edtf, edtf: true
-  property :release, virtual: true, default: 'immediate'
-  property :embargo_date, embargo_date: true, assign_if: ->(params) { params['release'] == 'embargo' }
+  property :release, virtual: true, prepopulator: (proc do |*|
+    self.release = embargo_date.present? ? 'embargo' : 'immediate'
+  end)
+  property :embargo_date, embargo_date: true
 
   validates :created_edtf, created_in_past: true
   validates :published_edtf, created_in_past: true
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -20,9 +20,11 @@ class DraftWorkForm < Reform::Form
   property :access
   property :license
   property :agree_to_terms
+  property :created_type, virtual: true, prepopulator: (proc do |*|
+    self.created_type = created_edtf.is_a?(EDTF::Interval) ? 'range' : 'single'
+  end)
   property :created_edtf, edtf: true, range: true
   property :published_edtf, edtf: true
-
   property :release, virtual: true, default: 'immediate'
   property :embargo_date, embargo_date: true, assign_if: ->(params) { params['release'] == 'embargo' }
 
--- a/app/forms/edtf.rb
+++ b/app/forms/edtf.rb
@@ -24,7 +24,6 @@ module Edtf
       property "#{prop_name}_range(4i)", virtual: true
       property "#{prop_name}_range(5i)", virtual: true
       property "#{prop_name}_range(6i)", virtual: true
-      property "#{prop_name}_type", virtual: true, default: 'single'
     end
   end
 
--- a/app/forms/embargo_date_params_filter.rb
+++ b/app/forms/embargo_date_params_filter.rb
@@ -9,7 +9,11 @@ class EmbargoDateParamsFilter
       next unless dfn[:embargo_date]
 
       name = dfn[:name]
-      date_attributes[name] = deserialize(params, name) if dfn[:assign_if].call(params)
+      date_attributes[name] = if params.slice(*release_params).values.include?('immediate') # rubocop:disable Performance/InefficientHashSearch
+                                nil # Do not attempt to deserialize date if release is immediate
+                              else
+                                deserialize(params, name)
+                              end
     end
 
     params.merge(date_attributes)
@@ -20,5 +24,12 @@ class EmbargoDateParamsFilter
     month = params.delete("#{date_attribute}(2i)").to_i
     day = params.delete("#{date_attribute}(3i)").to_i
     Date.new(year, month, day) if Date.valid_date?(year, month, day)
+  end
+
+  private
+
+  # Work and Collection models use differently named release params
+  def release_params
+    %w[release release_option]
   end
 end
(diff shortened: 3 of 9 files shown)


app/forms/edtf.rb
```diff
@@ -24,7 +24,6 @@ def create_range(prop_name)
       property "#{prop_name}_range(4i)", virtual: true
       property "#{prop_name}_range(5i)", virtual: true
       property "#{prop_name}_range(6i)", virtual: true
-      property "#{prop_name}_type", virtual: true, default: 'single'
```
mjgiarlo, line 27: Moved to `DraftWorkForm`.

app/forms/draft_work_form.rb
```diff
@@ -20,11 +20,15 @@ class DraftWorkForm < Reform::Form
   property :access
   property :license
   property :agree_to_terms
+  property :created_type, virtual: true, prepopulator: (proc do |*|
```
mjgiarlo, line 23: This is the pattern to set a virtual property based on a currently set value: using a reform prepopulator. The parens around the proc are to prevent a syntax error.

**NOTE**: these prepopulators do not automatically fire. They must be invoked in the form via `form.prepopulate!`. This is done in the controller, which is why testing this in specs that do not go through controller code must call it manually.

app/forms/embargo_date_params_filter.rb
```diff
@@ -9,7 +9,11 @@ def call(schema, params)
       next unless dfn[:embargo_date]
 
       name = dfn[:name]
-      date_attributes[name] = deserialize(params, name) if dfn[:assign_if].call(params)
+      date_attributes[name] = if params.slice(*release_params).values.include?('immediate') # rubocop:disable Performance/InefficientHashSearch
```
mjgiarlo, line 12: This code helps `nil` out previously set date values if the work switches from an embargo to immediate release. Prior to this change, the correct radio button was selected but a date rendered in the value associated with the unselected radio button. (That is, the "immediately" radio button would be clicked, but an embargo date was still seen beneath it.)

spec/components/works/embargo_component_spec.rb
```diff
@@ -9,15 +9,29 @@
   let(:work_form) { WorkForm.new(work) }
   let(:rendered) { render_inline(described_class.new(form: form)) }
 
+  before do
+    work_form.prepopulate!
```
mjgiarlo, line 13: See above.

spec/factories/works.rb
```diff
@@ -17,11 +17,12 @@
     state { 'first_draft' }
     collection
 
-    factory :valid_deposited_work do
-      deposited
-      with_keywords
-      with_contributors
-      with_attached_file
+    factory :valid_work do
```
mjgiarlo, line 20: Before this we had some specs creating a work factory with keywords, contributors, and an attached file but it wasn't obvious why THOSE three associations were used. It's because they're necessary for validation, so give this a name.

spec/forms/work_form_spec.rb
```diff
@@ -8,6 +8,10 @@
 
   let(:work) { build(:work) }
 
+  before do
+    form.prepopulate!
```
mjgiarlo, line 12: See above.

spec/forms/work_form_spec.rb
```diff
@@ -46,31 +50,23 @@
 
     it 'filters out name values for the wrong type' do
       form.validate(contributors: contributors)
-      expect(form.contributors.size).to eq 2
+      expect(form.contributors.size).to eq 3
```
mjgiarlo, line 53: There is now a third contributor because of the contributor prepopulator in the form. This is how the form works in deployed instances.

spec/forms/work_form_spec.rb
```diff
@@ -46,31 +50,23 @@
...
+      expect(form.contributors.third.last_name).to be_nil
     end
   end
 
   describe 'contributors validation' do
     let(:contributor_error) { { contributors: ['Please add at least one contributor.'] } }
+    let(:work) { build(:valid_work) }
 
-    context 'with no contributors' do
-      it 'does not validate' do
```
mjgiarlo, line 60: No, the contributor prepopulator prevents this from ever being the case, so remove the test as this is not a behavior in deployed instances.

spec/requests/create_work_spec.rb
```diff
@@ -347,7 +347,8 @@
             contact_email: '',
             abstract: '',
             license: License.license_list.first,
-            work_type: 'text'
+            work_type: 'text',
+            release: 'immediate'
```
mjgiarlo, line 351: Needed the changes in this file to make the tests pass and after spending more than half my day fighting reform after having already had a working PR... I didn't care enough to put more time into this PR.

spec/forms/work_form_spec.rb
jcoyne: Prepopulate only happens on show, not on update, right?

app/forms/embargo_date_params_filter.rb
```diff
@@ -21,4 +25,11 @@ def deserialize(params, date_attribute)
     day = params.delete("#{date_attribute}(3i)").to_i
     Date.new(year, month, day) if Date.valid_date?(year, month, day)
   end
+
+  private
+
+  # Work and Collection models use differently named release params
+  def release_params
+    %w[release release_option]
```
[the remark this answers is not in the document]
mjgiarlo: @jcoyne can we do this in a follow-on PR? I can write up an issue if so.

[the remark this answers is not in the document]
jcoyne: Yes, I wasn't suggesting do it here.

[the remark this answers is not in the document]
mjgiarlo: @jcoyne ok, thanks.

spec/forms/work_form_spec.rb
mjgiarlo: @jcoyne I mean, this test fails now due to the prepopulator change. We prepopulate when [rendering the new action](https://github.com/sul-dlss/happy-heron/blob/78f978ffcfbbbe50b505c8da65b3bbaa48b3050b/app/controllers/works_controller.rb#L19) and when [rendering the edit action](https://github.com/sul-dlss/happy-heron/blob/78f978ffcfbbbe50b505c8da65b3bbaa48b3050b/app/controllers/works_controller.rb#L41).

mjgiarlo: @jcoyne see the change I made above on line 12.

jcoyne (approved):

Issue #836, opened 2021-01-13
Impose restrictions on embargo date settings in an item based on collection settings


peetucket:
If the collection manager restricts the embargo date in a collection, this constraint needs to apply to items in the collection

See Design Doc for the various scenarios: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.n5g71obk0sxl

Has two impacts:

*  item edit/create page constrains the values available to the user based on collection values
* validate values on item save

peetucket:
Note: I believe we also need to translate the values set at the collection level (e.g. '6 months', '1 year') to actual dates that get set at the work level.  Equivalent to what happens in ETD here: https://github.com/sul-dlss/hydra_etd/blob/master/app/models/submission.rb#L110

PR #867, opened 2021-01-19
update embargo release options and reduce duplication

### LINKED ISSUE #818: Update Fixed release embargo options for collection settings
The issue was raised in the context of the H2 data migration meeting that the collection fixed release settings don't match the Hydrus settings and make mapping of this difficult. I've updated the Design Doc to reflect the Hydrus settings.

See Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.dh2mljowzvv

@astridu Could you double-check the changes I made in the Design Doc? Thanks.

The wireframes aren't updated, but here's what Hydrus does that we are trying to mimic here:

peetucket:
## Why was this change made?

Fix #818 - adjust the embargo release date options in the collection form to match the spec.  Also, I refactored the code to move the options to a single location (used in both the form and the validation of the values).

Unit tests and localhost browser

None

### DIFF
--- a/app/components/collections/form_component.html.erb
+++ b/app/components/collections/form_component.html.erb
@@ -77,12 +77,7 @@
           between date of deposit and
         <% end %>
         <%= form.select :release_duration,
-          options_for_select([['1 month from date of deposit', '1 month'],
-                              ['3 months from date of deposit', '3 month'],
-                              ['6 months from date of deposit', '6 month'],
-                              ['1 year from date of deposit', '1 year'],
-                              ['2 years from date of deposit', '2 years'],
-                              ['3 years from date of deposit', '3 years']]),
+          options_for_select(collection_form.class::EMBARGO_RELEASE_DURATION_OPTIONS),
                               {},
                               class: 'form-select release-duration' %>
       </div>
--- a/app/forms/draft_collection_form.rb
+++ b/app/forms/draft_collection_form.rb
@@ -6,6 +6,11 @@ class DraftCollectionForm < Reform::Form
   extend T::Sig
   feature EmbargoDate
 
+  EMBARGO_RELEASE_DURATION_OPTIONS = [['6 months from date of deposit', '6 month'],
+                                      ['1 year from date of deposit', '1 year'],
+                                      ['2 years from date of deposit', '2 years'],
+                                      ['3 years from date of deposit', '3 years']].freeze
+
   property :name
   property :description
   property :contact_email
@@ -39,9 +44,7 @@ class DraftCollectionForm < Reform::Form
   validate :reviewable_form
   validates :release_date, embargo_date: true
   validates :release_option, presence: true, inclusion: { in: %w[immediate delay depositor-selects] }
-  validates :release_duration, inclusion: {
-    in: ['1 month', '2 months', '6 months', '1 year', '2 years', '3 years']
-  }, allow_blank: true
+  validates :release_duration, inclusion: { in: EMBARGO_RELEASE_DURATION_OPTIONS.map(&:last) }, allow_blank: true
 
   def sync(*)
     update_depositors
--- a/spec/requests/create_collection_spec.rb
+++ b/spec/requests/create_collection_spec.rb
@@ -116,7 +116,7 @@ RSpec.describe 'Create a collection' do
             'release_date(1i)' => '2020',
             'release_date(2i)' => '7',
             'release_date(3i)' => '14',
-            'release_duration' => '1 month',
+            'release_duration' => '1 year',
             related_links_attributes: related_links
           }
         end
--- a/sorbet/rails-rbi/mailers/collections_mailer.rbi
+++ b/sorbet/rails-rbi/mailers/collections_mailer.rbi
@@ -3,6 +3,9 @@
 # Please rerun bundle exec rake rails_rbi:mailers to regenerate.
 class CollectionsMailer
   sig { returns(ActionMailer::MessageDelivery) }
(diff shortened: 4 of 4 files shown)


spec/requests/create_collection_spec.rb
```diff
@@ -116,7 +116,7 @@
             'release_date(1i)' => '2020',
             'release_date(2i)' => '7',
             'release_date(3i)' => '14',
-            'release_duration' => '1 month',
```
peetucket, line 119: This test uses a value which no longer allowed, I just changed it to an allowed value.  The actual value used is unimportant for the test itself, just needs to be balid.

app/components/collections/form_component.html.erb
```diff
@@ -77,12 +77,7 @@
           between date of deposit and
         <% end %>
         <%= form.select :release_duration,
-          options_for_select([['1 month from date of deposit', '1 month'],
-                              ['3 months from date of deposit', '3 month'],
-                              ['6 months from date of deposit', '6 month'],
-                              ['1 year from date of deposit', '1 year'],
-                              ['2 years from date of deposit', '2 years'],
-                              ['3 years from date of deposit', '3 years']]),
+          options_for_select(collection_form.class::EMBARGO_RELEASE_DURATION_OPTIONS),
```
mjgiarlo, line 80: If this doesn't need to vary, how might you feel about replacing `collection_form.class::` with `DraftCollectionForm::`?

app/forms/draft_collection_form.rb
```diff
@@ -6,6 +6,11 @@ class DraftCollectionForm < Reform::Form
   extend T::Sig
   feature EmbargoDate
 
+  EMBARGO_RELEASE_DURATION_OPTIONS = [['6 months from date of deposit', '6 month'],
+                                      ['1 year from date of deposit', '1 year'],
+                                      ['2 years from date of deposit', '2 years'],
+                                      ['3 years from date of deposit', '3 years']].freeze
```
mjgiarlo, line 12: I don't have strong feelings about this, but if you convert this to a hash (which `options_for_select` should be able to use just as well), you could change the `.map(&:last)` on line 47 to `.values`. Maybe clearer? Maybe not. What do you think?

app/forms/draft_collection_form.rb
```diff
@@ -6,6 +6,11 @@ class DraftCollectionForm < Reform::Form
   extend T::Sig
   feature EmbargoDate
 
+  EMBARGO_RELEASE_DURATION_OPTIONS = [['6 months from date of deposit', '6 month'],
```
mjgiarlo, line 9: Not sure if it matters, because apparently it used to vary, but should the value here be `6 month` or `6 months`?

spec/requests/create_collection_spec.rb
mjgiarlo: OK, no problem. And I'm not sure this PR really needs other testing, since it is a unification of existing code paths w/ some elimination of values.

app/components/collections/form_component.html.erb
peetucket: Yeah, that's fine.  The values currently never vary, in theory you can have a `DraftCollectionForm` or a `CollectionForm`.

app/forms/draft_collection_form.rb
peetucket: Yes, that's better.  I'll make it a hash.

app/forms/draft_collection_form.rb
peetucket: I just kept the current value from the old code.  Is there a specification somewhere showing that it should be '6 months' and not '6 month' ?

app/forms/draft_collection_form.rb
```diff
@@ -39,9 +44,7 @@ class DraftCollectionForm < Reform::Form
   validate :reviewable_form
   validates :release_date, embargo_date: true
   validates :release_option, presence: true, inclusion: { in: %w[immediate delay depositor-selects] }
-  validates :release_duration, inclusion: {
-    in: ['1 month', '2 months', '6 months', '1 year', '2 years', '3 years']
```
mjgiarlo, line 43: @peetucket this line is why I was inquiring re: `6 months` vs. `6 month`, but it might not make a different either way.

peetucket: Oh, hmmm... that is a mismatch to what was being set before in the options in the form itself.  I'll investigate.

peetucket: Here: https://github.com/sul-dlss/happy-heron/pull/867/files#diff-ec9c1247d60f9db33645bcaac1fbfb6f45fadb5841a990e6a9da52771222b0b5L82

mjgiarlo: @peetucket ok, fine by me! thanks

peetucket: Ok, I think the actual values are not currently used, but will need to be soon, when we translate to actual embargo date at the work level, captured in a comment I made to a different ticket here: https://github.com/sul-dlss/happy-heron/issues/836#issuecomment-763860711

However, in order to be consistent with hydra_etd, I'm going with "6 months" (see https://github.com/sul-dlss/hydra_etd/blob/master/app/models/submission.rb#L110 and https://github.com/sul-dlss/hydra_etd/blob/main/app/views/submit/lightbox/_lb_embargo.html.erb#L1)

mjgiarlo (approved):

PR #869, opened 2021-01-20
make collection form consistent with other attribute retrieval

peetucket:
## Why was this change made?

Make the collection form params consistent and get rid of reform indirection

Unit tests


mjgiarlo (approved):

mjgiarlo:
This is a pure refactoring and as such does not require any test changes.

PR #874, opened 2021-01-21
add edit links to collection details page

### LINKED ISSUE #695: Add pencil icons for editing to Collection details page
Each of the tables with editable information (everything but Collection Information and History) should have the pencil icons that open the form and take you to that spot in the form for editing.  We should also have a pencil icon at the top which just goes to the top of the edit collection page.

see see projects.invisionapp.com/share/EQXC9CLKCR2#/screens/426926002

This matches the current functionality on the work detail/edit pages

peetucket:
## Why was this change made?

Fixes #695, adding edit collection links on the collection detail pages

Some notes:

- I created a base class for the Collection show page components, so I could add a helper method to create the edit links in just one place, this mimics what is done for the Work show page (which uses a single wrapper component for the whole show page, unlike what is done for collections) ... see https://github.com/sul-dlss/happy-heron/blob/main/app/components/works/detail_component.rb#L37-L42 and https://github.com/sul-dlss/happy-heron/blob/main/app/views/works/show.html.erb#L4 and compare to the collection show page
- There is a link in the Terms part of the show page that is intended to link to the new section being added to the collection show page in 793

Localhost browser and unit tests

None


app/components/collections/detail_component.html.erb
```diff
@@ -1,7 +1,7 @@
 <table class="table table-sm mb-5">
   <thead class="table-light">
     <tr>
-      <th scope="col" class="table-title col-3">Details</th>
+      <th scope="col" class="table-title col-3">Details<%= edit_link('#', "Edit details") %></th>
```
jcoyne, line 4: Where is this edit_link method defined?

app/components/collections/detail_component.rb
```diff
@@ -3,15 +3,7 @@
...
-  class DetailComponent < ApplicationComponent
-    sig { params(collection: Collection).void }
-    def initialize(collection:)
-      @collection = collection
-    end
-
-    sig { returns(Collection) }
-    attr_reader :collection
-
+  class DetailComponent < Collections::ShowComponent
```
jcoyne, line 6: Why this inheritance?  This doesn't seem to me like a natural inheritance relationship.

app/components/collections/links_component.rb
```diff
@@ -3,15 +3,7 @@
...
-  class LinksComponent < ApplicationComponent
-    sig { params(collection: Collection).void }
-    def initialize(collection:)
-      @collection = collection
-    end
-
-    sig { returns(Collection) }
-    attr_reader :collection
-
+  class LinksComponent < Collections::ShowComponent
```
jcoyne, line 6: I don't understand why this is an inheritance.  This does not seem like natural inheritance to me.

app/components/collections/participants_component.rb
```diff
@@ -3,15 +3,7 @@
...
-  class ParticipantsComponent < ApplicationComponent
-    sig { params(collection: Collection).void }
-    def initialize(collection:)
-      @collection = collection
-    end
-
-    sig { returns(Collection) }
-    attr_reader :collection
-
+  class ParticipantsComponent < Collections::ShowComponent
```
jcoyne, line 6: Why inheritance here?

app/components/collections/release_component.rb
```diff
@@ -3,15 +3,7 @@
...
-  class ReleaseComponent < ApplicationComponent
-    sig { params(collection: Collection).void }
-    def initialize(collection:)
-      @collection = collection
-    end
-
-    sig { returns(Collection) }
-    attr_reader :collection
-
+  class ReleaseComponent < Collections::ShowComponent
```
jcoyne, line 6: why inheritance here?

app/components/collections/terms_of_use_component.rb
```diff
@@ -3,15 +3,7 @@
...
-  class TermsOfUseComponent < ApplicationComponent
-    sig { params(collection: Collection).void }
-    def initialize(collection:)
-      @collection = collection
-    end
-
-    sig { returns(Collection) }
-    attr_reader :collection
-
+  class TermsOfUseComponent < Collections::ShowComponent
```
jcoyne, line 6: why inheritance here?

app/components/collections/workflow_review_component.rb
```diff
@@ -3,15 +3,7 @@
...
-  class WorkflowReviewComponent < ApplicationComponent
-    sig { params(collection: Collection).void }
-    def initialize(collection:)
-      @collection = collection
-    end
-
-    sig { returns(Collection) }
-    attr_reader :collection
-
+  class WorkflowReviewComponent < Collections::ShowComponent
```
jcoyne, line 6: why inheritance here?

jcoyne (changes requested):
Inheritance should only be used when "X is a Y" is true.  The ShowComponent represents the whole page, so it's not true that the smaller components on the page are themselves specializations of the whole show page.

app/components/collections/detail_component.html.erb
mjgiarlo: https://github.com/sul-dlss/happy-heron/pull/874/files#diff-54534c85a1e46345b7a8c23b23c570bb737ed9ad39c06a725b86a7b08cd98d1cR16 

And I suspect that's why all the inheritance was added.

peetucket: Exactly correct.  Could use a mixin if we don't like the inheritance?  All of those components with inheritance are related to the collection detail page, which was the logic in doing it that way.

This method of defining an "edit_link" method is also used in the Work detail page, but in that case there is a one component for the whole page instead of breaking it out into sections like we do for the collection, so no need to for inheritance there to avoid duplication.

mjgiarlo: Maybe define `edit_link` as a helper method in `ApplicationController`, like `#user_with_groups`, and invoke it in components that need it via `helper.edit_link` ?

peetucket: yeah, we could do that, though then I would argue we call it something like `collection_edit_link` and do the same with the work to make it all consistent (by moving it from the work component to the application controller)

mjgiarlo: @peetucket dig that.

jcoyne: Perhaps make EditLinkComponent

peetucket: and same for the Work for consistency?  in this PR or maybe follow on?

peetucket:
> Inheritance should only be used when "X is a Y" is true. The ShowComponent represents the whole page, so it's not true that the smaller components on the page are themselves specializations of the whole show page.

Ok, reverted all the inheritance and now using a component for the collection edit links.

app/components/collections/header_component.html.erb
```diff
@@ -1,5 +1,6 @@
 <header class="title">
   <span class="header-text"><%= name %></span>
+  <%= render Collections::EditLinkComponent.new(collection: collection, anchor: '#', label: "Edit #{name}") %>
```
jcoyne, line 3: Won't this result in a link like: `http://example.com/path##` with two octothorpes?

app/components/collections/detail_component.html.erb
peetucket: Side note: this consistency with `Works` was completed in #891

app/components/collections/header_component.html.erb
peetucket: Yes, good catch, I missed this one - I fixed it now.

jcoyne (approved):

PR #884, opened 2021-01-21
Only allow access selection if collection is depositor-selects

### LINKED ISSUE #697: Impose restrictions on visibility settings in an item based on collection settings
If the collection manager sets the visibility settings for "world" or "Stanford" and not "depositor chooses", then the depositor should not be able to choose the other option. 

I was able to deposit an item with Stanford visibility (https://sul-h2-stage.stanford.edu/works/99) into a collection with world visibility (https://sul-h2-stage.stanford.edu/collections/82).

See Design Doc for the various scenarios: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.n5g71obk0sxl

Has two impacts:
- item edit/create page constrains the values available to the user based on collection values
- validate values on item save

aaron-collier:
## Why was this change made?

Fixes #697 by removing the selection list if the collection has an access right set other than `depositor-selects`.

When stanford:

Screen Shot 2021-01-21 at 3 35 35 PM

TBD

N/A

### DIFF
--- a/app/components/works/embargo_component.rb
+++ b/app/components/works/embargo_component.rb
@@ -30,5 +30,9 @@ module Works
     def access_from_collection
       'TBD' # TODO: by https://github.com/sul-dlss/happy-heron/pull/884
     end
+
+    def default_access
+      Collection.find(reform.collection_id).access
+    end
   end
 end
--- a/app/components/works/embargo_component.rb
+++ b/app/components/works/embargo_component.rb
@@ -18,7 +18,9 @@ module Works
 
     # The access level is specified by the collection
     def user_can_set_access?
-      true # TODO: by https://github.com/sul-dlss/happy-heron/pull/884
+      return true if access_from_collection == 'depositor-selects'
+
+      false
     end
 
     def availability_from_collection
@@ -28,17 +30,7 @@ module Works
     end
 
     def access_from_collection
-      'TBD' # TODO: by https://github.com/sul-dlss/happy-heron/pull/884
-    end
-
-    def collection_access_restriction
-      reform.model.collection.access
-    end
-
-    def depositor_selects_access?
-      return true if collection_access_restriction == 'depositor-selects'
-
-      false
+      collection.access
     end
   end
 end
--- a/app/components/works/embargo_component.rb
+++ b/app/components/works/embargo_component.rb
@@ -31,10 +31,14 @@ module Works
       'TBD' # TODO: by https://github.com/sul-dlss/happy-heron/pull/884
     end
 
-    def default_access
-      return unless reform.collection_id
+    def collection_access_restriction
+      reform.model.collection.access
+    end
 
-      Collection.find_by(id: reform.collection_id)&.access
+    def depositor_selects_access?
+      return true if collection_access_restriction == 'depositor-selects'
+
+      false
     end
   end
 end
--- a/app/components/works/embargo_component.rb
+++ b/app/components/works/embargo_component.rb
@@ -32,7 +32,9 @@ module Works
     end
 
     def default_access
-      Collection.find(reform.collection_id).access
+      return unless reform.collection_id
+
+      Collection.find_by(id: reform.collection_id)&.access
     end
   end
 end
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -39,6 +39,7 @@ class DraftWorkForm < Reform::Form
     # Choose between using the user provided citation and the auto-generated citation
     params['citation'] = params.delete('citation_auto') if params['default_citation'] == 'true'
     deserialize_embargo(params)
+    access_from_collection(params)
     super(params)
   end
 
@@ -58,6 +59,12 @@ class DraftWorkForm < Reform::Form
     end
   end
   # rubocop:enable Metrics/AbcSize
+
+  def access_from_collection(params)
+    return if model.collection.access == 'depositor-selects'
+
+    params['access'] = model.collection.access
+  end
 
   collection :contributors,
              populator: ContributorPopulator.new(:contributors, Contributor),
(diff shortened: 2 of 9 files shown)


app/components/works/embargo_component.rb
```diff
@@ -53,5 +53,11 @@ def embargo_month
     def embargo_day
       embargo_date&.day
     end
+
+    def default_access
+      return unless reform.collection_id
+
+      Collection.find_by(id: reform.collection_id)&.access
```
jcoyne, line 60:
```suggestion
      reform.model.collection.access
```

app/components/works/embargo_component.html.erb
```diff
@@ -40,7 +40,11 @@
   <div class="mb-3 row">
     <%= form.label :access, 'Who can access?', class: 'col-sm-2 col-form-label' %>
     <div class="col-sm-10">
-      <%= form.select :access, [['Stanford Community', 'stanford'], ['Everyone', 'world']], { disabled: "", prompt: "Select..." }, class: "form-select", required: true %>
+      <% if default_access == 'depositor-selects' %>
```
jcoyne, line 43: Can you move this conditional into the class?  So that we have `depositor_selects_access?`

app/components/works/embargo_component.rb
```diff
@@ -53,5 +53,11 @@ def embargo_month
     def embargo_day
       embargo_date&.day
     end
+
+    def default_access
```
jcoyne, line 57: can this be renamed to `collection_access_restriction`?

jcoyne (changes requested):

jcoyne (changes requested):
I'm not sure that this change ensures that a new item gets the access specified by the collection.  Can you add a test for that?

app/components/works/embargo_component.rb
```diff
@@ -53,5 +53,17 @@ def embargo_month
     def embargo_day
       embargo_date&.day
     end
+
+    def collection_access_restriction
+      return unless reform.collection_id
```
jcoyne, line 58: I'm not sure I understand this check. Every item should have a collection ID

aaron-collier: I believe when I first ran the specs there are instances where it is nil that threw errors. I will verify this morning.

app/components/works/embargo_component.html.erb
```diff
@@ -11,20 +11,7 @@
       "Deposit" at the bottom of this page. Or you may specify a specific date in the future.</p>
 
     <%= render Works::AvailableDateComponent.new(form: form) %>
-    <% if user_can_set_access? %>
```
jcoyne, line 14: I rather prefer to keep these conditionals here because then the AccessComponent doesn't need to know anything about availablilty.

jcoyne (changes requested):

app/components/works/embargo_component.html.erb
aaron-collier: I can see that. It felt like a lot of duplication, but happy to revert this change.

spec/components/works/embargo_component_spec.rb
```diff
@@ -13,6 +13,7 @@
 
   before do
     work_form.prepopulate!
+    collection.access = 'depositor-selects'
```
jcoyne, line 16: Can this move up to line 9 and use the `depositor_selects_access` trait?

spec/components/works/access_component_spec.rb
```diff
@@ -4,13 +4,30 @@
...
+  it 'renders the component' do
+    expect(rendered.to_html)
+      .to include('Select which audience you would like to have access to download')
+  end
+
+  context 'when collection access is depositor selects' do
+    let(:work) { build(:work) }
+
+    before do
+      work.collection.access = 'depositor-selects'
```
jcoyne, line 26: Can you define a new `let(:collection)` in this context that uses the :depositor_selects_access trait?

spec/components/works/form_component_spec.rb
```diff
@@ -12,6 +12,7 @@
 
   before do
     allow(controller).to receive(:allowed_to?).and_return(true)
+    work.collection.access = 'depositor-selects'
```
jcoyne, line 15: Can this move up to line 9 and use the `depositor_selects_access` trait?

jcoyne (changes requested):

app/forms/draft_work_form.rb
```diff
@@ -61,6 +62,12 @@ def deserialize_embargo(params)
   end
   # rubocop:enable Metrics/AbcSize
 
+  def access_from_collection(params)
+    return if model.collection.access == 'depositor-selects'
+
+    params['access'] = model.collection.access
```
jcoyne, line 68: Can you add a request spec that tests this case?

jcoyne (changes requested):

app/forms/draft_work_form.rb
aaron-collier: Let me know if the additions to the reqeust/create_new_work spec are not sufficient.

jcoyne:
@aaron-collier there's a conflict to resolve here.

app/forms/draft_work_form.rb
jcoyne: All of the test I see are doing the "depositor-selects" branch of this method.  I don't see a test for "the user submitted something that isn't allowed and we overwrote it".

spec/requests/create_work_spec.rb
```diff
@@ -143,7 +143,7 @@
         allow(DepositJob).to receive(:perform_later)
       end
 
-      context 'when a collection allows embargo and with everything' do
+      context 'when a collection allows embargo but sets access and with everything' do
```
jcoyne, line 146:
```suggestion
     context 'when a collection allows embargo but restricts access and with everything' do
 ```

jcoyne (approved):

PR #885, opened 2021-01-21
add breadcrumb navigation component

### LINKED ISSUE #835: Add breadcrumb navigation bar below header
See https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403 for an example (screenshot below):

Also see the design doc (https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#) for a description of what should show and when.  Note that we are using the word "Dashboard" instead of the home icon.

* Home icon --> goes to dashboard
* breadcrumbs after that (e.g. collection show page if on an item show/edit page for an item in the collection)

peetucket:
## Why was this change made?

Fixes #835 -- add breadcrumb navigation

(Note that we are using the word "Dashboard" instead of the home icon as described in a comment in #835, which is different than what is currently shown in the wireframes).

Also for previous reviewers to this PR, for simplicity and so we can get this PR merged sooner, I reverted the help link back to where it was in the top banner bar (i.e. it is no longer changed here as it was in earlier commits). I think we can have a separate ticket to move and style the help link over on the right side of the breadcrumbs if needed in future work.

Here is an example of how it currently looks:

Screen Shot 2021-01-25 at 1 45 03 PM

Todo:

- [x] change styling to add/remove underlines or other distinguishing of links from just display labels
- [x] reduce height of breadcrumb bar
- [x] fix separator so it is a caret as shown in design

Localhost browser and new unit tests added. 

None


app/components/bread_crumb_nav_component.html.erb
```diff
@@ -0,0 +1,20 @@
+<div>
```
mjgiarlo, line 1: Minor nit: can you rename files with `bread_crumb` to `breadcrumb` and classes with `BreadCrumb` to `Breadcrumb`?

app/components/bread_crumb_nav_component.rb
```diff
@@ -0,0 +1,19 @@
...
+  end
+
+  def breadcrumb_links
+    @breadcrumbs.map do |breadcrumb|
+      if breadcrumb[:link].empty?
+        breadcrumb[:title]
+      else
+        link_to breadcrumb[:title].presence || 'No title', breadcrumb[:link], class: 'breadcrumb-link'
+      end
+    end
```
mjgiarlo, line 17: LGTM

app/controllers/application_controller.rb
```diff
@@ -10,17 +10,34 @@ class ApplicationController < ActionController::Base
   # Smartly redirect user back to URL they requested before authenticating
   # From: https://github.com/heartcombo/devise/wiki/How-To:-Redirect-back-to-current-page-after-sign-in,-sign-out,-sign-up,-update
   before_action :store_user_location!, if: :storable_location?
+  before_action :root_breadcrumb
```
mjgiarlo, line 13: Does this wind up creating a dashboard breadcrumb even on the homepage?

app/controllers/application_controller.rb
```diff
@@ -10,17 +10,34 @@ class ApplicationController < ActionController::Base
...
 
   rescue_from ActionPolicy::Unauthorized, with: :deny_access
 
   authorize :user_with_groups
 
+  attr_reader :breadcrumbs
```
mjgiarlo, line 19: Hmm, can you say what this is adding?

app/controllers/collections_controller.rb
```diff
@@ -13,6 +13,8 @@ def new
 
     @form = CollectionForm.new(collection)
     @form.prepopulate!
+
+    add_breadcrumb 'Create', ''
```
mjgiarlo, line 17: I guess the empty string has to be passed because of l35 in the application controller?

app/components/bread_crumb_nav_component.html.erb
peetucket: yes, will do

app/controllers/application_controller.rb
peetucket: Yes, good point.  Let me sort that out -- this should probably move to the dashboard controller.

app/controllers/application_controller.rb
peetucket: Isn't this needed so we can access the breadcrumbs array in the application layout (to pass to the component for rendering)?  https://github.com/sul-dlss/happy-heron/pull/885/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3ebR45

app/controllers/collections_controller.rb
peetucket: yes, but this could be done more elegantly by perhaps passing in a hash to the `add_breadcrumb` method, allowing us to see more clearly what is being passed in and eliminating that empty value...i'll play with that, it may make the definition of the `add_breadcrumb` method more clear as well (right now just a copypasta from the gist you sent)

app/controllers/application_controller.rb
mjgiarlo: I don't think it is. I believe it's accessible by virtue of being an ivar in the controller. You open to commenting it out and verifying?

app/controllers/collections_controller.rb
mjgiarlo: That approach sounds better to me too, @peetucket

app/controllers/application_controller.rb
peetucket: indeed, it works, removing

app/controllers/application_controller.rb
peetucket: Had to leave in application controller to ensure the dashboard link is rendered on all other pages, but guarded against showing on the home page

app/controllers/collections_controller.rb
peetucket: Refactored and it looks a lot cleaner now.  Only deal is you add one breadcrumb at a time instead of being able to pass in an array, but I think this makes for cleaner method calls anyway.

app/controllers/collections_controller.rb
```diff
@@ -13,6 +13,8 @@ def new
 
     @form = CollectionForm.new(collection)
     @form.prepopulate!
+
+    add_breadcrumb title: 'Create', confirm: true
```
jcoyne, line 17: This should be "New collection" to be consistent with the design for items: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403

app/controllers/works_controller.rb
```diff
@@ -39,6 +42,10 @@ def edit
 
     @form = WorkForm.new(work)
     @form.prepopulate!
+
+    add_breadcrumb title: work.collection.name, link: collection_path(work.collection), confirm: true
+    add_breadcrumb title: work.title, link: work_path(work), confirm: true
+    add_breadcrumb title: 'Edit'
```
jcoyne, line 48: This is not in the design: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420954309

app/controllers/application_controller.rb
```diff
@@ -21,8 +22,27 @@ def user_with_groups
   end
   helper_method :user_with_groups
 
+  def add_root_breadcrumb
+    @breadcrumbs = []
```
jcoyne, line 26: I don't understand why we need to have these in the controller?  Can they just be part of the view component?

jcoyne (changes requested):

app/controllers/collections_controller.rb
peetucket: Note that the design doc actually says "Create Collection" in the breadcrumb nav for this page, but agreed consistency with the Work makes sense and will change to that (unless @astridu disagrees?).  See https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.1mct3tk7ytls

app/controllers/works_controller.rb
peetucket: I got this from the Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#  But you are right, it is not consistent with the Wireframe.  @astridu ?

app/controllers/application_controller.rb
peetucket: I'm not quite sure how we'd implement this functionality easily in the breadcrumb view component, since each controller action is adding different items to the breadcrumb nav before we get to the rendering (and the welcome controller is turning off breadcrumb nav completely)?

app/javascript/stylesheets/breadcrumb.scss
```diff
@@ -0,0 +1,8 @@
+.breadcrumb-link {
+    padding-left: 0.5rem;
+    padding-right: 0.5rem;
+    color: #000000;
```
jcoyne, line 4: Can you use `$body-color` here?

app/controllers/application_controller.rb
peetucket: Refactored to remove from controllers as discussed

app/javascript/stylesheets/breadcrumb.scss
peetucket: done

peetucket:
Note: For simplicity and so we can get this PR merged sooner, I reverted the help link back to where it was for this PR (i.e. it is no longer changed here as it was in earlier commits) and instead just added new styling for the breadcrumbs.  I think w can have a separate ticket to move and style the help link over on the right side of the breadcrumbs if needed in future work.

app/javascript/stylesheets/bootstrapVariables.scss
```diff
@@ -14,3 +14,5 @@ $modal-backdrop-opacity: 0.8;
 // Tables
 $table-border-color: $silver-sand;
 $table-group-separator-color: $silver-sand;
+
+$breadcrumb-divider: quote(">");
```
jcoyne, line 18: Is this used somewhere?

jcoyne: NM, i see this is a bootstrap variable.

jcoyne (approved):

PR #893, opened 2021-01-22
Switch back to local forms (and hotwire for help)

jcoyne:
## Why was this change made?

This seems to be the way the Rails ecosystem is now heading.


mjgiarlo (approved):
LGTM. Should we test in QA?

PR #894, opened 2021-01-22
add confirmation to cancel buttons on edit pages

### LINKED ISSUE #892: Cancel button on edit collection and work pages should show confirmation alert
Similar to what is shown when the user clicks "Discard".  This avoids users navigating away from an edit page and losing their changes by mistake.  Validated this request with Amy and Astrid.

peetucket:
## Why was this change made?

Fixes #892 - add a confirmation to the cancel button shown on edit pages to verify user will lose changes if clicked.  Already discussed with Astrid and Amy.

Localhost browser

None

### DIFF
--- a/app/components/collections/cancel_component.html.erb
+++ b/app/components/collections/cancel_component.html.erb
@@ -1,1 +1,2 @@
-<%= link_to 'Cancel', dashboard_path %>
+<%= link_to 'Cancel', dashboard_path,
+  data: { confirm: 'Are you sure you want to leave this page? Your changes will not be saved.' } %>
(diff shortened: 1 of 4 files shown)


mjgiarlo (commented):
Worth adding a quick feature spec or two to make sure this doesn't regress?

peetucket:
> Worth adding a quick feature spec or two to make sure this doesn't regress?

yes, that makes sense

peetucket:
Tests added for cancel button confirmation for works and collections

mjgiarlo (approved):

PR #913, opened 2021-01-27
Auto citation should not be enabled if they provided a citation

### LINKED ISSUE #660: User choice to have preferred citation off should stick
If I select to have the preferred citation off and save as draft or deposit an item, when I go in to edit or version that item, it always shows me the preferred citation as ON. This should always be whatever the last saved state was.

jcoyne:
## Why was this change made?
Fixes #660


app/javascript/controllers/auto_citation_controller.js
```diff
@@ -10,9 +10,13 @@ export default class extends Controller {
 
     this.updateDisplay()
 
-    this.displayDefault(this.switchTarget.checked)
+    // If the manualTarget matches the autoTarget, then display the auto.
+    const showAuto = this.manualTarget.value == this.autoTarget.value
```
mjgiarlo, line 14: Should `==` be `===` here?

mjgiarlo (approved):

app/javascript/controllers/auto_citation_controller.js
jcoyne: Does `[1] == true` ?

mjgiarlo: I am not a JavaScript interpreter. :laughing:

PR #934, opened 2021-01-27
Display collection license validation inline in the form

mjgiarlo:
Follows the approach used in https://github.com/sul-dlss/happy-heron/pull/928 and https://github.com/sul-dlss/happy-heron/pull/919

Also, fixes the build.

## Why was this change made?

Consistency and kinder UX. Note that the UI looks messed up in the screen shot due to #929 not having been merged yet.

license_validation

CI, local browser

None

### DIFF
--- a/app/components/collections/edit_license_component.rb
+++ b/app/components/collections/edit_license_component.rb
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: false
 # frozen_string_literal: true
 
 module Collections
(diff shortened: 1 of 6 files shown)


app/components/collections/edit_license_component.rb
```diff
@@ -1,4 +1,4 @@
-# typed: strict
+# typed: false
```
mjgiarlo, line 1: I fought sorbet for about ten minutes and gave up.

jcoyne (approved):

PR #967, opened 2021-01-29
[DRAFT] Add version description component

### LINKED ISSUE #454: Create "What's changing" UI element
https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420954309

aaron-collier:
## Why was this change made?

Fixes #454 

Screen Shot 2021-01-29 at 9 07 15 AM

Screen Shot 2021-01-29 at 9 07 09 AM

IN PROGRESS

### DIFF
--- a/app/components/works/version_component.rb
+++ b/app/components/works/version_component.rb
@@ -1,0 +1,13 @@
+# typed: true
+# frozen_string_literal: true
+
+module Works
+  # Renders a widget for setting the title and contact for the work.
+  class VersionComponent < ApplicationComponent
+    def initialize(form:)
+      @form = form
+    end
+
+    attr_reader :form
+  end
+end
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -106,6 +106,8 @@ class WorksController < ObjectsController
   sig { params(work: Work).void }
   def after_save(work)
     work.event_context = { user: current_user }
+    work.event_context[:description] = params[:work][:version_description] unless params[:work][:version_description].nil?
+
     work.update_metadata!
     if deposit_button_pushed?
       if work.collection.review_enabled?
@@ -127,7 +129,7 @@ class WorksController < ObjectsController
                      'created_range(1i)', 'created_range(2i)', 'created_range(3i)',
                      'created_range(4i)', 'created_range(5i)', 'created_range(6i)',
                      :abstract, :citation_auto, :citation, :default_citation,
-                     :access, :license,
+                     :access, :license, :version_description,
                      :release, 'embargo_date(1i)', 'embargo_date(2i)', 'embargo_date(3i)',
                      :agree_to_terms, subtype: [],
                                       attached_files_attributes: %i[_destroy id label hide file],
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -105,8 +105,10 @@ class WorksController < ObjectsController
 
   sig { params(work: Work).void }
   def after_save(work)
-    work.event_context = { user: current_user }
-    work.event_context[:description] = params[:work][:version_description] unless params[:work][:version_description].nil?
+    work.event_context = { 
+      user: current_user, 
+      description: params.dig(:work, :version_description) 
+    }.compact
 
     work.update_metadata!
     if deposit_button_pushed?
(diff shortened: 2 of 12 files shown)


app/javascript/stylesheets/work.scss
```diff
@@ -58,3 +58,11 @@
     margin-left: 10px;
   }
 }
+
+.work-editor {
+  .version {
+    background-color: $stanford-digital-pink;
+    border-color: $stanford-cardinal;
+    border-width: 3px;
+  }
```
mjgiarlo, line 67: I believe you should be able to move these lines into `app/javascript/stylesheets/workEditor.scss` which already has a `.work-editor` block in it.

app/javascript/stylesheets/palette.scss
```diff
@@ -1,9 +1,12 @@
 $alabaster: #f2f1eb;
 $silver-sand: #c0c0bf; // "30% black" in Stanford brand identity
 $stanford-cardinal: #8c1515;
+$stanford-cardinal-light: #B83A4B;
+$stanford-spirited-light: #F4795B;
```
mjgiarlo, line 5: Cruft now?

app/javascript/stylesheets/palette.scss
```diff
@@ -1,9 +1,12 @@
...
 $stanford-cool-grey: #53565a; // used in SDR header (not yet added)
 $stanford-digital-blue: #006cb8;
 $stanford-digital-green: #008566;
 $stanford-digital-red: #b1040e;
+$stanford-digital-pink: #F3E7E7;
```
mjgiarlo, line 10: This is TOTALLY a nitpick, but the other colors in `palette.scss` use lowercase hex digits and the new entry uses uppercase.

app/controllers/works_controller.rb
```diff
@@ -92,6 +92,8 @@ def work_form(work)
   sig { params(work: Work).void }
   def after_save(work)
     work.event_context = { user: current_user }
+    work.event_context[:description] = params[:work][:version_description] unless params[:work][:version_description].nil?
```
mjgiarlo, line 95: Would you consider:

```suggestion
work.event_context = { user: current_user, description: params.dig(:work, :version_description) }.compact
```

:question:

app/components/works/version_component.rb
```diff
@@ -0,0 +1,13 @@
...
+# frozen_string_literal: true
+
+module Works
+  # Renders a widget for setting the title and contact for the work.
+  class VersionComponent < ApplicationComponent
+    def initialize(form:)
+      @form = form
+    end
+
+    attr_reader :form
```
mjgiarlo, line 11: Should the component render no matter what state the work is in? Or should it only render in certain states? Might be a PO question. cc: @amyehodge @jcoyne

mjgiarlo (commented):
Looking good so far! @amyehodge do we want this for collections as well as works?

app/controllers/works_controller.rb
aaron-collier: I like.

amyehodge:
@mjgiarlo , that makes a lot of sense to me. It can provide useful information and frankly is a good reminder that you're versioning the collection. It is not in the spec at all. Should I make a new ticket for that?

app/components/works/version_component.rb
aaron-collier: That's a good question. I figured as this is the edit form it probably always needs to be there... but validation/etc?

mjgiarlo:
@amyehodge :speech_balloon: 
> @mjgiarlo , that makes a lot of sense to me. It can provide useful information and frankly is a good reminder that you're versioning the collection. It is not in the spec at all. Should I make a new ticket for that?

I'd defer to @aaron-collier on that.

app/javascript/stylesheets/palette.scss
mjgiarlo: I wonder if we want a different name, too, so as not to suggest that this is an official Stanford identity color. When googling that hex code, I see that some paint companies sell this as `Laughing Maiden` or `Ibis White`.

How about `$ibis-white: #f3e7e7; // light pinkish hue used in what's changing box` here?

app/components/works/version_component.rb
amyehodge: It needs to be there for VERSION drafts, but not for the initial drafts. In other words, it should always be there for drafts of previously accessioned content. 

A user may open this draft and save it multiple times, so 1) any entry should be saved when the draft is saved and continue to be editable until the deposit is made and 2) be required to make the deposit.

PR #974, opened 2021-01-30
truncate long titles in breadcrumbs

### LINKED ISSUE #953: Implement title length restrictions for breadcrumbs on all pages and In Progress section on dashboard
Very long titles are making a mess of the breadcrumb display and the In Progress section on the dashboard.

@astridu has recommended that we implement the same 150 character limit on the item titles in the breadcrumbs that we did for the tables, with ellipses at the end and mouseovers that display the entire title. 

And then the same treatment for the In Progress section on the dashboard, except that that will need to be 100 characters due to space constraints.

peetucket:
## Why was this change made?

Fixes #953 - truncate collection and work title lengths in breadcrumb navigation and in-progress section of dashboard

Unit test and localhost browser

### DIFF
--- a/app/components/breadcrumb_nav_component.html.erb
+++ b/app/components/breadcrumb_nav_component.html.erb
@@ -1,14 +1,17 @@
 <nav class="breadcrumbs">
   <nav id="breadcrumbs" aria-label="breadcrumb" class="container-xxl">
     <ol class="breadcrumb">
-      <% breadcrumbs.each do |breadcrumb| %>
+      <% breadcrumbs.each do |breadcrumb|
+        full_title = breadcrumb[:title].presence || 'No title'
+        truncated_title = truncate(full_title, length: 150, separator: ' ')
+        %>
         <li class="breadcrumb-item">
           <% if breadcrumb[:link].blank? %>
-            <span class="breadcrumb-link"><%= breadcrumb[:title] %></span>
+            <span title="<%=full_title%>" class="breadcrumb-link"><%= truncated_title %></span>
           <% else %>
-            <%= link_to breadcrumb[:title].presence || 'No title',
-                        breadcrumb[:link], class: 'breadcrumb-link',
-                        data: (breadcrumb[:confirm] ? { confirm: 'Are you sure you want to leave this page? Your changes will not be saved.' } : nil) %>
+            <%= link_to truncated_title,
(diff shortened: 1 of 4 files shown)


app/components/breadcrumb_nav_component.html.erb
```diff
@@ -1,14 +1,17 @@
 <nav class="breadcrumbs">
   <nav id="breadcrumbs" aria-label="breadcrumb" class="container-xxl">
     <ol class="breadcrumb">
-      <% breadcrumbs.each do |breadcrumb| %>
+      <% breadcrumbs.each do |breadcrumb|
+        full_title = breadcrumb[:title].presence || 'No title'
```
jcoyne, line 5: Can you move these to a method in the class file?

app/components/dashboard/in_progress_row_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-4">
-    <%= collection_name %> &gt; <%= edit_work_link %>
+    <span title="<%=collection_name%>"><%= truncate(collection_name, length: 100, separator: ' ') %></span> &gt; <%= edit_work_link %>
```
jcoyne, line 3: Please move this to a method in the class

jcoyne (changes requested):

app/components/breadcrumb_nav_component.html.erb
peetucket: good idea, done

app/components/dashboard/in_progress_row_component.html.erb
peetucket: done

app/components/breadcrumb_nav_component.rb
```diff
@@ -1,4 +1,4 @@
-# typed: true
+# typed: false
```
peetucket, line 1: this was switched because sorbet does not recognize `truncate` as being defined in activesupport (it complains its not defined in the component:

```
app/components/breadcrumb_nav_component.rb:18: Method truncate does not exist on BreadcrumbNavComponent https://srb.help/7003
    18 |    truncate(full_title(breadcrumb), length: 150, separator: ' ')
```

not sure how to deal with that

jcoyne (approved):

PR #976, opened 2021-02-01
Differentiate authors from other contributors

### LINKED ISSUE #545: Update contributors section to have two separate sections for ordered authors and unordered contributors
As per our discussion earlier today:

> We had a brief meeting today to discuss this change around order in the cocina models gem: https://github.com/sul-dlss/cocina-models/commit/1cef523e3c25a4817880435e95c2bb7de89f40a2#

> Concerns were raised by the developers that the new order property would be duplicative of the natural order of the list and would require additional code complexity everywhere this data is consumed.  It became clear that in our current system, where we persist MODS, we have no way of storing this explicit order. 

> It was decided that Amy and Astrid would work on changes to the UI that would help show the users that order is important for contributors that appear in the citation (mainly authors).  There will also be another section for non-cited contributors (e.g. funder, advisor, conference)

> The developers agreed that we would revert the cocina-models change above and continue to use the natural order for contributors, related items, and parts for now.  We will be able to revisit these decisions once we have moved off Fedora 3 and can persist more than what can map to MODS.

Astrid has updated the wireframes and the design doc (https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.d3q6bp4xt33j) to reflect the changes noted in the third paragraph.

jcoyne:
## Why was this change made?
The authors become part of the citation, but other contributors don't.

Fixes #545

Before:
Screen Shot 2021-02-01 at 12 43 47 PM

After:
Screen Shot 2021-02-01 at 12 42 12 PM


mjgiarlo (commented):
Can you include before and after screenshots?

jcoyne:
@mjgiarlo added

app/components/works/contributor_row_component.rb
```diff
@@ -4,14 +4,26 @@
 module Works
   # Renders a widget corresponding to a single contributor to the work.
   class ContributorRowComponent < ApplicationComponent
-    sig { params(form: ActionView::Helpers::FormBuilder).void }
-    def initialize(form:)
+    sig { params(form: ActionView::Helpers::FormBuilder, required: T::Boolean, citation: T::Boolean).void }
```
mjgiarlo, line 7: Is it worth considering splitting this component into two components: the current ContributorRowComponent and a new AuthorRowComponent? Or do you think that wouldn't get us much of value? I'm assuming the added complexity in this component is driven by using it in both contexts, but I may be off.

mjgiarlo (commented):
Should 'Author' still appear in the dropdown of the contributor row?

jcoyne:
@mjgiarlo That question was already answered here: https://github.com/sul-dlss/happy-heron/issues/545#issuecomment-768391033

app/components/works/contributor_row_component.rb
jcoyne: I don't fee this is too complex. It's effectively one boolean (as required and citation are always set the same), only more expressive.

app/components/works/authors_and_contributors_component.rb
```diff
@@ -0,0 +1,17 @@
+# typed: strict
+# frozen_string_literal: true
+
+module Works
+  # A widget for managing both the authors (ordered) and unordered contributors to the work.
```
mjgiarlo, line 5: Trivial tweak: can you move `(ordered)` in front of `authors`? Don't want to risk encouraging folks to infer that contributors are both unordered and ordered.

app/forms/contributor_populator.rb
```diff
@@ -10,7 +10,7 @@ def call(form, fragment:, **)
     item = existing_record(form: form, id: fragment['id'])
 
     if fragment['_destroy'] == '1'
-      form.contributors.delete(item)
+      form.send(field).delete(item)
```
mjgiarlo, line 13: No strong feelings about this, but since `ApplicationPopulator` already has `#value`, this could also be: `value(form).delete(item)` Alternatively, consider changing `send` to `public_send`?

app/forms/contributor_populator.rb
```diff
@@ -25,7 +25,7 @@ def call(form, fragment:, **)
       fragment['first_name'] = nil
       fragment['last_name'] = nil
     end
-    item || form.contributors.append(Contributor.new)
+    item || form.send(field).append(klass.new)
```
mjgiarlo, line 28: Same as line 13.

app/models/abstract_contributor.rb
```diff
@@ -0,0 +1,84 @@
+# typed: true
+# frozen_string_literal: true
+
+# This is the parent of Author and Contributors
+# It is necessary so that Rails STI will assign the proper type to each concrete model
+# and so that when you query Work#contributors you only get the non-author contributors.
```
mjgiarlo, line 6: Nice.

app/forms/contributor_populator.rb
jcoyne: Good eye. I have made those changes.

app/components/works/authors_and_contributors_component.rb
jcoyne: done

mjgiarlo (approved):

PR #979, opened 2021-02-01
fix test that fails periodically

### LINKED ISSUE #978: Fix test that fails periodically
The following test seems to fail periodically and randomly.  Investigate why.

`./spec/features/create_new_work_spec.rb:174:in block (4 levels) in <top (required)>`

```
Failure/Error: 
  expect(page).to have_content 'Keller, M. (2020, February). My Title. ' \
    "Stanford Digital Repository. Available at #{Work::LINK_TEXT}"

  expected to find text "Keller, M. (2020, February). My Title. Stanford Digital Repository. Available at :link will be inserted here automatically when available:" in "XXXXXXX.". (However, it was found 1 time including non-visible text.)
./spec/features/create_new_work_spec.rb:174:in `block (4 levels) in <top (required)>'
```

peetucket:
## Why was this change made?

Fixes #978 - an attempt to fix the flappy test

Test suite unflapped?

### DIFF
--- a/spec/features/create_new_work_spec.rb
+++ b/spec/features/create_new_work_spec.rb
@@ -153,6 +153,13 @@ RSpec.describe 'Create a new work in a deposited collection', js: true do
         page.attach_file(Rails.root.join('spec/fixtures/files/sul.svg')) do
           click_button('Choose files')
         end
+        sleep 1 # pause to ensure file upload completes before we proceed
+        # see https://github.com/sul-dlss/happy-heron/issues/978
+        # we were hoping the with block below would block until upload
+        # completes, but it did not
+        within '.dropzone-previews' do
+          expect(page).to have_content('sul.svg')
+        end
 
         fill_in 'Title of deposit', with: 'My Title'
         fill_in 'Contact email', with: user.email
--- a/sorbet/rbi/gems/ast.rbi
+++ b/sorbet/rbi/gems/ast.rbi
@@ -7,7 +7,7 @@
 #
 #   https://github.com/sorbet/sorbet-typed/new/master?filename=lib/ast/all/ast.rbi
 #
-# ast-2.4.1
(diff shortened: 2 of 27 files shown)


mjgiarlo (commented):
I'll give it a run-through.

spec/features/create_new_work_spec.rb
```diff
@@ -154,6 +154,8 @@
           click_button('Choose files')
         end
 
+        sleep 1 # pause to wait for file to upload
```
mjgiarlo, line 157: I wanted to test this locally but I can't get the dang tests to fail: instead of sleeping for a fixed, hard-coded value, I wonder if we can get Capybara to do this for us via:

```suggestion
        expect(page).to have_content('sul.svg')
```

Happy to pair too since apparently you *can* see this fail locally.

spec/features/create_new_work_spec.rb
```diff
@@ -171,6 +173,7 @@
         check 'I agree to the SDR Terms of Deposit'
         click_button 'Deposit'
 
+        expect(page).to have_current_path(work_path(Work.last)) # this will fail if validation fails
```
mjgiarlo, line 176: :+1:

spec/features/create_new_work_spec.rb
mjgiarlo: This didn't do the trick.

mjgiarlo:
@sul-dlss/infrastructure-team @peetucket and I sunk time into this and didn't come up with anything better than `sleep 1` and, given all the other work we want to do, felt the `sleep` was good enough for now. If anyone else is inclined to toss ideas at @peetucket or wants to take this on, I think that'd be welcome. Otherwise, I might recommend we move forward with this and iterate.

jcoyne (approved):

Issue #935, opened 2021-01-27
Implement progress bar on item deposit page


amyehodge:
Design doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.2j7a9yfzetyj
Wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403

amyehodge:
Waiting to put this in the ready column until details of the functionality of the progress bar on the dashboard are ironed out.

mjgiarlo:
Encompasses #993

PR #985, opened 2021-02-02
add terms to progress bars

### LINKED ISSUE #969: Add a new box at the end of the Progress Bar for "Terms"
This will make it clear to users when they haven't yet completed the required task of agreeing to the Terms of Deposit.

See https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/433047903

peetucket:
## Why was this change made?

Fixes #969 : 
- add "terms" to progress bar on both dashboard and edit work page.
- ensure terms progress field is updated dynamically when you check/uncheck on edit work page
- fix misnamed JS controller reference in other terms on edit work page
- use a different selector for the files that accurately reflects when files exist (previous selector always return length == 0, so that it never looked like files had been uploaded)

NOTE: when working on this, it appears multiple other fields in the terms progress bar on the edit work page are not being updated right now:
- author, description, release, license

Unit tests

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -7,7 +7,7 @@
                         data: { dropzone_target: 'input',
                                 edit_deposit_target: 'fileField',
                                 'progress-step': 'file',
-                                action: "change->progress#check" } %>
+                                action: "change->edit-deposit#check" } %>
     <div class="dz-message needsclick text-secondary">
       <p>Drop files here</p>
     </div>
--- a/app/components/works/agreement_component.html.erb
+++ b/app/components/works/agreement_component.html.erb
@@ -1,6 +1,9 @@
 <div class="row justify-content-end agreement">
   <div class="form-check col-auto">
-    <%= form.check_box :agree_to_terms, required: true, class: 'form-check-input' %>
+    <%= form.check_box :agree_to_terms, required: true, class: 'form-check-input',
+        data: { action: "change->edit-deposit#check",
+                        'progress-step': 'terms', edit_deposit_target: 'termsField' }
+     %>
     <%= form.label :agree_to_terms,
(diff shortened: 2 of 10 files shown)


app/components/works/add_files_component.html.erb
```diff
@@ -7,7 +7,7 @@
                         data: { dropzone_target: 'input',
                                 edit_deposit_target: 'fileField',
                                 'progress-step': 'file',
-                                action: "change->progress#check" } %>
+                                action: "change->edit-deposit#check" } %>
```
jcoyne, line 10: Nice catch!

app/components/works/agreement_component.html.erb
```diff
@@ -1,6 +1,9 @@
 <div class="row justify-content-end agreement">
   <div class="form-check col-auto">
-    <%= form.check_box :agree_to_terms, required: true, class: 'form-check-input' %>
+    <%= form.check_box :agree_to_terms, required: true, class: 'form-check-input',
+        data: { action: "change->auto-citation#updateDisplay change->edit-deposit#check",
```
jcoyne, line 4: I don't think `change->auto-citation#updateDisplay` is necessary, is it?  This doesn't contribute to the auto citation at all.

jcoyne (changes requested):

jcoyne:
@peetucket there is a ticket for completing the work progress: #935

app/components/works/agreement_component.html.erb
peetucket: oops, that was a mistake - i removed it, good catch

peetucket:
> @peetucket there is a ticket for completing the work progress: #935

ahh, cool, i won't worry about those other fields in this PR then

jcoyne (approved):

PR #986, opened 2021-02-02
Rename the contributors table to abstract_contributors

### LINKED ISSUE #982: [happy-heron/stage] ActiveRecord::SubclassNotFound: Invalid single-table inheritance type: Author is not a subclass of Contributor
## Backtrace

    line 123 of [PROJECT_ROOT]/app/components/works/detail_component.html.erb: call
    line 9 of [PROJECT_ROOT]/app/views/works/show.html.erb: _app_views_works_show_html_erb___4023963886571987361_73060

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/74777003)

jcoyne:
## Why was this change made?
Rails isn't as clever with STI as I had thought. It was complaining of Author records in the contributors table.
Fixes #982

### DIFF
--- a/app/models/abstract_contributor.rb
+++ b/app/models/abstract_contributor.rb
@@ -5,9 +5,6 @@
 # It is necessary so that Rails STI will assign the proper type to each concrete model
 # and so that when you query Work#contributors you only get the non-author contributors.
 class AbstractContributor < ApplicationRecord
-  self.abstract_class = true
-  self.table_name = 'contributors'
-
   extend T::Sig
   SEPARATOR = '|'
 
--- a/db/migrate/20210202044303_rename_contributors.rb
+++ b/db/migrate/20210202044303_rename_contributors.rb
@@ -1,0 +1,5 @@
+class RenameContributors < ActiveRecord::Migration[6.1]
+  def change
+    rename_table 'contributors', 'abstract_contributors'
+  end
+end
--- a/app/components/dashboard/deposit_progress_component.rb
+++ b/app/components/dashboard/deposit_progress_component.rb
@@ -24,7 +24,7 @@ module Dashboard
 
     sig { returns(T::Boolean) }
     def has_author?
-      work.contributors.any?
+      work.authors.any?
     end
 
     sig { returns(T::Boolean) }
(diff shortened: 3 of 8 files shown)


app/models/abstract_contributor.rb
```diff
@@ -5,8 +5,7 @@
 # It is necessary so that Rails STI will assign the proper type to each concrete model
 # and so that when you query Work#contributors you only get the non-author contributors.
 class AbstractContributor < ApplicationRecord
-  self.abstract_class = true
-  self.table_name = 'contributors'
+  # self.abstract_class = true
```
mjgiarlo, line 8: Cruft? Or do you mean to uncomment this again?

db/migrate/20210202044303_rename_contributors.rb
```diff
@@ -0,0 +1,5 @@
+class RenameContributors < ActiveRecord::Migration[6.1]
+  def change
+    rename_table 'contributors', 'abstract_contributors'
```
mjgiarlo, line 3: It seems the contributors are not so abstract anymore. Worth re-considering the name?

mjgiarlo (commented):
I have a couple small questions.

I also wonder what this suggests about our test suite. This seems like we had an app with something fundamentally broken in #976 and yet the test suite was :green_circle:.

db/migrate/20210202044303_rename_contributors.rb
jcoyne: They absolutely are abstract, we never instantiate on of them.

app/models/abstract_contributor.rb
jcoyne: Thanks, I've removed this cruft.

db/migrate/20210202044303_rename_contributors.rb
jcoyne: The `abstract_class = true` is really poorly named. That is one strategy for mapping the inheritance structure of the classes to tables, the pure STI strategy is another.

jcoyne:
@mjgiarlo I think this says more about the fragility of the edge-cases of ActiveRecord's ORM more than it does about our code or test suite.

mjgiarlo (approved):

PR #989, opened 2021-02-02
Reserve a PURL

### LINKED ISSUE #464: Reserve a PURL should register an item
The user should be able to reserve a PURL with a new button, which is on the dashboard alongside the "Deposit to this collection" button.  See https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/433047903

When clicked, the action is described here (see p.20 under heading "Secondary button: Reserve a PURL
"): https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.kgkkjfrle9fb 

Basically:
1. Ask user for title (UI to be mocked up by @astridu )
2. Use this title to create a new work
3. Register the object to obtain a druid
4. Save item in database, and then return to dashboard so user can continue to edit

jmartin-sul:
## Why was this change made?

fixes #464 

test suite, manual testing on local instance

n/a


jmartin-sul:
the modal:

Screen Shot 2021-02-02 at 12 56 26 AM

two small issues i noticed with the modal:
1. Even when i called `this.titleTarget.focus()` in the `set_collection` method in `purl_reservation_controller.js`, the text entry field wouldn't get focus.  Not sure if there was a different way I should set focus?
2. Pressing enter in the title entry text box in the modal doesn't submit the form and reserve the PURL, though clicking "Submit" does.

reserved PURLs and buttons for adding them to each collection:

Screen Shot 2021-02-02 at 12 56 08 AM

app/javascript/controllers/state_updates_controller.js
```diff
@@ -38,7 +38,9 @@ export default class extends Controller {
   renderUpdates(data) {
     for (const [attribute, value] of Object.entries(data)) {
       const target = this.targets.find(attribute)
-      target.innerHTML = value
+      if (target) {
+        target.innerHTML = value
+      }
```
jmartin-sul, line 43: without this check, i got a javascript error in the browser when i manually triggered deposit completion for testing.  this seemed like a harmless check, and didn't break any existing tests.

the errors i got without this fix were like:
Screen Shot 2021-02-02 at 12 22 34 AM
Screen Shot 2021-02-02 at 12 51 32 AM

jmartin-sul: the one in the browser manifested when i ran `./bin/rake complete_deposits` while i had the dashboard open in my browser.  the test suite failure happened in the new `spec/features/purl_reservation_spec.rb` without this change.

jmartin-sul:
One other stray observation:  I noticed that I couldn't edit the top level `work_type` once the work was created.  Not sure whether that's intentional, but that would seem to be a bit in tension with the lack of opportunity to pick a work type when reserving a PURL.

app/javascript/controllers/state_updates_controller.js
mjgiarlo: I think this error must be due to the page that is rendered not having an element with `data-state-updates-target="purl"`.

mjgiarlo: @jmartin-sul :speech_balloon: 
> * I noticed that I couldn't edit the top level `work_type` once the work was created.  Not sure whether that's intentional, but that would seem to be a bit in tension with the lack of opportunity to pick a work type when reserving a PURL.
> * The styling on the modal doesn't quite match the mockups, and in particular, I couldn't get the "(You can update this later.)" explanation under the header to display on its own line if I put it with the header in the same `modal-header` `div`.  Maybe there's a better way to structure that?
> * Where the ticket description and the linked google doc diverged, I went with the ticket description, since it seemed to have been updated more recently (the doc said to use a default title, but the ticket had a mockup with a title field; the doc said the status should be reserved, but the ticket just said to register the work).
> 

Thanks for writing this up! I'd like to get @amyehodge's takes on those.

mjgiarlo:
@jmartin-sul :speech_balloon: 
> * I noticed that I couldn't edit the top level `work_type` once the work was created.  Not sure whether that's intentional, but that would seem to be a bit in tension with the lack of opportunity to pick a work type when reserving a PURL.
> * The styling on the modal doesn't quite match the mockups, and in particular, I couldn't get the "(You can update this later.)" explanation under the header to display on its own line if I put it with the header in the same `modal-header` `div`.  Maybe there's a better way to structure that?
> * Where the ticket description and the linked google doc diverged, I went with the ticket description, since it seemed to have been updated more recently (the doc said to use a default title, but the ticket had a mockup with a title field; the doc said the status should be reserved, but the ticket just said to register the work).
> 

Thanks for writing this up! I'd like to get @amyehodge's takes on those.

amyehodge:
@jmartin-sul , @mjgiarlo 

- It is intentional that the top-level work type is not editable. In the future we hope to have some customizations in the deposit form depending on the work type chosen and allowing the user to edit the work type would make that very messy. 
- Sounds like we need to add a way for the user to select the work type when they are reserving the PURL, because they can't make this selection later, is that true?
- I do recall that having the title field option was the more recent decision, so that sounds good. 
- I'm not clear on the distinction between reserved and registered?

jcoyne:
> This is not better than STI, per se, but it may be a better solution in some cases, particularly when child models may have a very different set of characteristics from one another.

that would be the simplest course.

jmartin-sul:
>     * It is intentional that the top-level work type is not editable. In the future we hope to have some customizations in the deposit form depending on the work type chosen and allowing the user to edit the work type would make that very messy.
> 
>     * Sounds like we need to add a way for the user to select the work type when they are reserving the PURL, because they can't make this selection later, is that true?
> 

@amyehodge, right, once they reserve the PURL, the associated work is registered, and would be edited using the usual work edit form for things that've been deposited already.  Both the ticket and the google doc descriptions seemed to indicate that the work would get registered (and the PURL reserved) once the user successfully submitted the modal's form.

@astridu, how should I go about allowing work type selection from the PURL reservation modal?  Would you just want to use the existing top-level work type selection UI?  Like is used for regular deposit?   Or something else?  If something else, could you give a mockup or a description?  Thanks!

>     * I'm not clear on the distinction between reserved and registered?

Me either 🙂.  I believe they're effectively the same, since getting the PURL (AFAIK) requires registration.  From a depositor's perspective, I can see maybe differentiating in the UI... though I'm not sure whether it'd be more confusing for a user to see a PURL reservation as a deposited work like other deposited works, or as a differentiated state of its own.  But from a backend perspective, I don't think there's really a difference?

jmartin-sul:
> > This is not better than STI, per se, but it may be a better solution in some cases, particularly when child models may have a very different set of characteristics from one another.
> 
> that would be the simplest course.

@jcoyne, not sure I understand this comment in this context, but also thinking it might've actually been meant for #986?  But if I'm off-base there, could you elaborate on what you're suggesting for this PR?  Thanks!

jmartin-sul:
Ah, I see that @astridu and @jcoyne started discussing the work type selection question in the ticket, now that I look back at it again.  But the answer there suggests that the user "submit that information when they return to edit this deposit", which would appear to be in conflict with @amyehodge's statement that we intentionally disallow edits to the top level work type after registration.

Or, am I (or is the current H2 codebase) conflating PURL reservation/object registration/work creation?  Is there disentangling to be done there (either in my mental model, or in the codebase)?

astridu:
@jmartin-sul Let me talk to Amy in the morning and will get back to you.

amyehodge:
@jmartin-sul Apologies that this is going around in circles a bit. And apologies for the verbosity below. Trying to be as clear as possible.

- We do not want the user to be allowed to change the work type once it has been chosen.
- So, either 1) the user chooses the work type when they reserve a PURL and then when they come back later to edit the item, they skip the work type modal and go directly to the form, as they do when they edit all other items 
OR
- 2) the user does NOT choose a work type when reserving a PURL, but when they come back to edit the item, they must choose a work type first and then proceed to the edit form as they do when creating a new item via the normal work path.

Since option 2) preserves our interest in keeping this process as simple as possible for the user, which @astridu pointed out is a strong desire for this use case, that seems like the best approach.

So, to summarize the interaction here:

1. Use clicks on the "Reserve a PURL" button
2. The user is presented with a box for entering a title as @astridu mocked up above.
3. User enters title and clicks "Sumbit"
4. The druid is registered in the back end (literally shows as "v1 Registered" in Argo, I assume) and displayed on the dashboard (as status "Reserved") for the user so they can grab it.
5. When the user returns to edit this item and clicks on the pencil icon, they will first be presented with the work type modal and once options are chosen that meet the validations, then can continue to the edit form.

app/models/work_type.rb
```diff
@@ -90,9 +90,14 @@ def initialize(id:, label:, icon:, subtypes:, cocina_type:)
     @cocina_type = cocina_type
   end
 
+  sig { returns(WorkType) }
+  def self.purl_reservation_type
+    new(id: 'purl_reservation', label: 'PURL reservation', icon: '', subtypes: [], cocina_type: 'n/a')
```
jcoyne, line 95: very clever!

jmartin-sul: 😄

spec/features/purl_reservation_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      click_link "Choose Type and Edit #{title}"
+      find('label', text: 'Music').click
+      check 'Sound'
+      check 'Image'
+      click_button 'Continue'
+
+      expect(work_version.reload.work_type).to eq 'music'
+      expect(work_version.subtype.sort).to eq %w[Image Sound]
+      # TODO: the redirect to edit page behavior being tested here works IRL in FF, but test gets bounced to /dashboard
```
jmartin-sul, line 52: i'm happy to file a ticket for this and link it from the comment.  capybara seems to be encountering a problem i initially encountered when testing manually in firefox.  but as commented here and in `WorksController#update_type`, the solution suggested by the `redirect_to` documentation solved the problem in my browser, but not in the spec.  seems like something that'd be nice to include in this feature test.

it seems better to me to get this PR merged and fix this as follow-on work, but also happy to try and fix in this PR.

spec/jobs/deposit_job_spec.rb
```diff
@@ -47,6 +47,18 @@
       expect(SdrClient::Deposit::UploadFiles).to have_received(:upload)
       expect(DepositStatusJob).to have_received(:perform_later).with(object: work_version, job_id: 1234)
     end
+
+    context 'when the deposit is for a PURL reservation' do
+      let(:work_version) do
+        build(:work_version, :reserving_purl, work: work)
+      end
+
+      # TODO: failing because purl reservation placeholder type is not valid cocina type
```
jmartin-sul, line 56: not sure what the right thing to do here is.  add a placeholder type to the allowable cocina types?  allow null work type in H2?  something else i'm not thinking of?

also no opinion as to whether this PR should be blocked over this shortcoming.  probably, but i could also see some value in getting it merged since it's been open for a while.  it's not huge, but it's not small, and it's been through a couple significant rebases.

jmartin-sul (commented):
I think this is ready for review, though depending on opinions regarding the identified TODOs, it may not be ready to merge.  The feature doesn't quite work since it won't yet deposit successfully, though it's largely implemented, and AFAICT introduces no regressions.

jmartin-sul:
a couple new screenshots now that the user is forced to pick a work type before editing a PURL reservation.  the second is what pops up when clicking on the pencil icon for a PURL reservation (note the type selection form pointing to `WorksController#update_type` for this purpose).

a PURL that's been reserved, and a request to reserve a PURL where the associated SDR deposit/registration is still in progress:
Screen Shot 2021-02-19 at 11 02 29 PM

attempting to edit a work for a PURL reservation:
Screen Shot 2021-02-19 at 11 05 10 PM



one other implementation note:  after i realized i should probably add the PURL reservation states i added, i considered getting rid of the PURL reservation placeholder type, since i added that in lieu of new states when i thought the placeholder `work_type` would suffice.  but the placeholder type still seemed useful, both to keep the `NOT NULL` constraint on the DB column, and because it's an easy single thing to query.  but it is arguably just a de-normalization of the state info (since the intent is that the placeholder type will be in use IFF the state is `purl_requested`, `reserving_purl`, or `purl_reserved`).

jcoyne (approved):

spec/jobs/deposit_job_spec.rb
jmartin-sul: fixed by https://github.com/sul-dlss/happy-heron/pull/1223

PR #990, opened 2021-02-02
Require the email to be entered before the title is marked as complete

### LINKED ISSUE #957: "Title" section of progress bar on dashboard should consider entries for both title and email
The "Title" section of the progress bar should not be green unless BOTH the title and the email contact in this section have content in them. Right now it is only considering the title.

jcoyne:
## Why was this change made?

Fixes #957

### DIFF
--- a/app/components/dashboard/deposit_progress_component.rb
+++ b/app/components/dashboard/deposit_progress_component.rb
@@ -19,7 +19,7 @@ module Dashboard
 
     sig { returns(T::Boolean) }
     def has_title?
-      work.title.present?
+      work.title.present? && work.contact_email.present?
     end
 
     sig { returns(T::Boolean) }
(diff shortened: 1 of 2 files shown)


mjgiarlo (commented):
Do we also need changes to the edit_deposit_controller so that the client-side behavior is consistent with the server-side behavior change?

jcoyne:
@mjgiarlo ticket #957 was specifically about the dashboard.

mjgiarlo:
@jcoyne :speech_balloon: 
> @mjgiarlo ticket #957 was specifically about the dashboard.

That change needn't be in this PR. I can write up an issue.

jcoyne:
@mjgiarlo I think that would be covered in #935

mjgiarlo (approved):

PR #991, opened 2021-02-02
link contact us text in file upload box to modal

### LINKED ISSUE #937: "Contact us" text in file upload section should open the contact form modal

peetucket:
## Why was this change made?

Fixes #937 - link contact us text to modal

Localhost browser

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -1,7 +1,11 @@
 <section data-controller="dropzone" data-dropzone-max-file-size="2000" data-dropzone-max-files="1000" id="file">
   <header>Add your files</header>
   <p>All file types are accepted. If you have a deposit that is over 10GB, or
-    have any trouble with adding your files, contact us.</p>
+    have any trouble with adding your files,
+    <%= link_to('contact us.', '#contactUsModal', data: {
+                              bs_toggle: 'modal',
+                              bs_target: '#contactUsModal' }) %>
+  </p>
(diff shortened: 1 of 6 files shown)


app/components/works/add_files_component.html.erb
```diff
@@ -1,7 +1,13 @@
 <section data-controller="dropzone" data-dropzone-max-file-size="2000" data-dropzone-max-files="1000" id="file">
   <header>Add your files</header>
   <p>All file types are accepted. If you have a deposit that is over 10GB, or
-    have any trouble with adding your files, contact us.</p>
+    have any trouble with adding your files,
+    <%= link_to('contact us.', '#contactUsModal', data: {
+                              bs_toggle: 'modal',
+                              bs_target: '#contactUsModal',
+                              controller: 'contact-us'
```
jcoyne, line 8: Why do we wrap the link with a stimulus controller?

jcoyne (changes requested):

app/components/works/add_files_component.html.erb
peetucket: this was copypasta from the link in the header, which calls stimulus like this:

https://github.com/sul-dlss/happy-heron/blob/main/app/views/layouts/application.html.erb#L42-L48

This seems to just run the connect action to ensure any links to `#help` show the modal?  I removed the reference to the controller in this link under the principle it doesn't need to run connect twice and the link still works.

jcoyne (approved):

PR #1011, opened 2021-02-04
only show edit icons when user is allowed to edit works or collections

### LINKED ISSUE #943: Item titles in the Progress Bar on the dashboard should link to the Item view page, not the edit form
Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.vhwtx3wnoqek
Wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/433047903

peetucket:
## Why was this change made?

- Users should only see edit icons/links if they are allowed to edit a collection or work.
- Refactor one spot where we create a edit work link to use the component instead
- Update specs
- Also fixes #943 (sorry @jcoyne didn't see you had this in progress), work text in progress bar should link to work show page, not work edit page

Unit tests and localhost browser

### DIFF
--- a/app/components/dashboard/in_progress_row_component.html.erb
+++ b/app/components/dashboard/in_progress_row_component.html.erb
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-4">
-    <%= show_collection_link %> &gt; <%= work_link %> <%= edit_work_link %>
+    <%= show_collection_link %> &gt; <%= work_link %> <%= render Works::EditLinkComponent.new(work: work, anchor: '', label: "Edit #{title}") %>
   </div>
   <div class="col-5">
     <%= render Dashboard::DepositProgressComponent.new(work: work) %>
--- a/app/components/dashboard/in_progress_row_component.html.erb
+++ b/app/components/dashboard/in_progress_row_component.html.erb
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-4">
-    <%= show_collection_link %> &gt; <%= work_link %> <%= render Works::EditLinkComponent.new(work: work, anchor: '', label: "Edit #{title}") %>
+    <%= show_collection_link %> &gt; <%= work_link %> <%= helpers.turbo_frame_tag dom_id(work, :edit), src: edit_button_work_path(work), target: '_top' %>
   </div>
   <div class="col-5">
     <%= render Dashboard::DepositProgressComponent.new(work: work) %>
--- a/app/components/dashboard/in_progress_row_component.rb
+++ b/app/components/dashboard/in_progress_row_component.rb
@@ -16,18 +16,12 @@ module Dashboard
       Dashboard::CollectionHeaderComponent.new(collection: work.collection).name
     end
 
-    def edit_work_link
-      link_to edit_work_path(work), aria: { label: "Edit #{title}" } do
-        tag.span class: 'fas fa-pencil-alt'
-      end
-    end
-
     def title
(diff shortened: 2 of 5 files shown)


app/components/dashboard/in_progress_row_component.rb
```diff
@@ -16,18 +16,12 @@ def collection_name
...
-      end
-    end
-
     def title
       @title ||= Works::DetailComponent.new(work: work).title
     end
 
     def work_link
-      link_to truncate(title, length: 100, separator: ' '), edit_work_path(work), title: title
+      link_to truncate(title, length: 100, separator: ' '), work_path(work), title: title
```
peetucket, line 24: this is the fix to #943

spec/features/edit_draft_work_spec.rb
```diff
@@ -20,7 +20,7 @@
       visit dashboard_path
 
       within('#deposits-in-progress') do
-        click_link work.title
+        click_link "Edit #{work.title}"
       end
```
peetucket, line 24: tests updated as a result of the change for #943 (title link no longer goes to edit page)

spec/features/edit_draft_work_spec.rb
```diff
@@ -59,7 +59,7 @@
       visit dashboard_path
 
       within('#deposits-in-progress') do
-        click_link work.title
+        click_link "Edit #{work.title}"
```
peetucket, line 62: tests updated as a result of the change for #943 (title link no longer goes to edit page)

spec/features/edit_draft_work_spec.rb
```diff
@@ -73,7 +73,7 @@
...
```
peetucket, line 76: tests updated as a result of the change for #943 (title link no longer goes to edit page)

peetucket:
@jcoyne  thoughts on why all the test failures?  it looks like the hotwire changes may be triggering re-rendering of html that includes the components i updated here, but that when this happens, the user context is lost, which breaks the ability to check current_user abilities.... i verified it happens in the browser too

app/components/works/edit_link_component.rb
```diff
@@ -11,6 +11,11 @@ def initialize(work:, anchor:, label:)
       @label = label
     end
 
+    sig { returns(T::Boolean) }
+    def render?
+      work.can_update_metadata? && !work.pending_approval?
```
jcoyne, line 16: will this prevent a reviewer from making changes?

jcoyne (changes requested):

app/components/works/edit_link_component.rb
peetucket: oh, yes ... i'll ditch the `!work.pending_approval?` check

app/components/dashboard/in_progress_row_component.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="row">
   <div class="col-4">
-    <%= show_collection_link %> &gt; <%= work_link %> <%= edit_work_link %>
+    <%= show_collection_link %> &gt; <%= work_link %> <%= helpers.turbo_frame_tag dom_id(work, :edit), src: edit_button_work_path(work), target: '_top' %>
```
peetucket, line 3: this is an edit link on the dashboard, switching to use turbo like other edit work links on the dashboard (so that only edit links on the work detail page use the edit link component)

jcoyne (approved):

PR #1029, opened 2021-02-06
add collections in progress section for collection managers

### LINKED ISSUE #946: Create section for managers on dashboard with Collections in progress.
This will look similar to the Deposits in progress but is for collections and will display only for collection managers.

Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.1lwd85ry26fj
Wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/434978163

peetucket:
## Why was this change made?

Fixes #946 - add an in progress section to the top of the dashboard, but for collections (not works) that are in a draft state and for which the user is defined as a collection manager

To do:
- [x] NO ~~do admins get to see all collections in draft state?  best way to add this to the query if so?~~

Todo

- [x] create a new spec for `in_progress_collection_component`
- [x] verify in a feature spec?

### DIFF
--- a/app/components/dashboard/in_progress_collection_component.rb
+++ b/app/components/dashboard/in_progress_collection_component.rb
@@ -1,0 +1,22 @@
+# typed: true
+# frozen_string_literal: true
+
+module Dashboard
+  # Renders a list of collections in progress
+  class InProgressCollectionComponent < ApplicationComponent
+    sig { params(presenter: DashboardPresenter).void }
+    def initialize(presenter:)
+      @presenter = presenter
+    end
+
+    attr_reader :presenter
+
+    delegate :collection_managers_in_progress, to: :presenter
+    delegate :user_with_groups, to: :helpers
+
+    sig { returns(T::Boolean) }
+    def render?
+      user_with_groups.administrator? || collection_managers_in_progress.size.positive?
+    end
+  end
+end
--- a/app/components/dashboard/in_progress_collection_component.rb
+++ b/app/components/dashboard/in_progress_collection_component.rb
@@ -16,7 +16,7 @@ module Dashboard
 
     sig { returns(T::Boolean) }
     def render?
-      user_with_groups.administrator? || collection_managers_in_progress.size.positive?
+      user_with_groups.administrator? || collection_managers_in_progress.present?
     end
   end
 end
--- a/app/components/dashboard/in_progress_collection_row_component.html.erb
+++ b/app/components/dashboard/in_progress_collection_row_component.html.erb
@@ -1,4 +1,4 @@
-<div style="margin-bottom: 0.5rem;" class="row">
+<div class="mb-2" class="row">
   <div class="col-4">
     <%= show_collection_link %>
   </div>
--- a/app/components/dashboard/in_progress_collection_row_component.html.erb
+++ b/app/components/dashboard/in_progress_collection_row_component.html.erb
@@ -1,0 +1,11 @@
+<div style="margin-bottom: 0.5rem;" class="row">
+  <div class="col-4">
+    <%= show_collection_link %>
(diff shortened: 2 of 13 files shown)


app/components/dashboard/in_progress_collection_row_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<div style="margin-bottom: 0.5rem;" class="row">
```
peetucket, line 1: adding some extra margin to these rows, else the complete draft buttons in each row will stack up too tightly

app/controllers/dashboards_controller.rb
```diff
@@ -14,9 +15,12 @@ def show
                      .joins(collection: :reviewed_by)
                      .where('reviewers.user_id' => current_user),
       in_progress: Work.with_state(:first_draft, :version_draft, :rejected)
-                       .where(depositor: current_user)
+                       .where(depositor: current_user),
+      collection_managers_in_progress: Collection.with_state(:first_draft, :version_draft)
+                                                 .joins(:managers).where("managers.user_id = #{current_user.id}")
```
peetucket, line 20: is there a better way to scope the query for collections where the current user is marked as the collection manager? or does this look ok?

app/components/dashboard/in_progress_collection_row_component.html.erb
jcoyne: can you use the `mb-*` bootstrap classes for this?

app/components/dashboard/in_progress_collection_component.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    attr_reader :presenter
+
+    delegate :collection_managers_in_progress, to: :presenter
+    delegate :user_with_groups, to: :helpers
+
+    sig { returns(T::Boolean) }
+    def render?
+      user_with_groups.administrator? || collection_managers_in_progress.size.positive?
```
jcoyne, line 19:
```suggestion
      user_with_groups.administrator? || collection_managers_in_progress.present?
```

app/controllers/dashboards_controller.rb
jcoyne: this is fine, but you should use bind variables. See https://blog.saeloun.com/2019/10/28/bind-parameters-in-activerecord-sql-queries.html

jcoyne: Alternatively `current_user.manages_collections.with_state(:first_draft, :version_draft)`

spec/features/delete_draft_collection_spec.rb
```diff
@@ -16,7 +16,9 @@
       visit dashboard_path
 
       accept_confirm do
-        click_link "Delete #{collection.name}"
+        within '#your-collections' do
```
jcoyne, line 19: Would `within_section 'Your collections'` work here?

jcoyne (changes requested):

app/controllers/dashboards_controller.rb
peetucket: thanks, i am familiar with but forgot about bind variables for this case, but that habtm association format looks nice

spec/features/delete_draft_collection_spec.rb
peetucket: It doesn't work, perhaps because the "+ Create a new collection" button is in the <section><header> tag too

jcoyne: Yeah, I was afraid of that.

jcoyne (approved):

PR #1043, opened 2021-02-08
Add a field for license_option

### LINKED ISSUE #958: On collection edit form, selecting option "Depositor selects license" should be saved when drafts are saved, even if the default license is not selected.
When editing a collection (in this case I'm on https://sul-h2-stage.stanford.edu/collections/105/edit#), if I choose the option in the Terms of use and license section that says "Depositor selects license" and then do *not* choose a default (which is required) 
 

and then click "Save as draft", the draft is saved, but the license option I chose isn't. When I open this back up again it has the other option selected:

If I have previously chosen a default license and saved the draft and then back in and edit this to keep this option selected but remove the default I've chosen and revert it to "Select..." then the same thing as above happens.

Desired behavior is for the user's radio button selection to stick when saving drafts. For both license options, a license must be selected before this can be deposited.

jcoyne:
## Why was this change made?
So we can store the radio button choice separate from whether a license was chosen.
Fixes #958

### DIFF
--- a/db/migrate/20210208201246_add_license_option_to_collection.rb
+++ b/db/migrate/20210208201246_add_license_option_to_collection.rb
@@ -1,0 +1,10 @@
+class AddLicenseOptionToCollection < ActiveRecord::Migration[6.1]
+  def change
+    add_column :collections, :license_option, :string
+    Collection.all.each do |col|
+      col.update(license_option: col.required_license.nil? ? 'depositor-selects' : 'required')
+    end
+    change_column_null :collections, :license_option, false
+    change_column_default :collections, :license_option, from: nil, to: 'required'
+  end
+end
--- a/app/components/collections/terms_of_use_component.html.erb
+++ b/app/components/collections/terms_of_use_component.html.erb
@@ -12,15 +12,15 @@
       <th>Terms of use</th>
       <td><%= I18n.t(:terms_of_use) %></td>
     </tr>
-    <% if required_license.present? %>
-      <tr>
-        <th>Required license</th>
-        <td><%= required_license %></td>
-      </tr>
-    <% elsif default_license.present? %>
+    <% if user_can_set_license? %>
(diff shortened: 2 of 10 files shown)


db/migrate/20210208201246_add_license_option_to_collection.rb
```diff
@@ -0,0 +1,10 @@
+class AddLicenseOptionToCollection < ActiveRecord::Migration[6.1]
+  def change
+    add_column :collections, :license_option, :string
+    Collection.all.each do |col|
+      col.update(license_option: col.user_can_set_license? ? 'depositor-selects' : 'required')
```
mjgiarlo, line 5: I'm a little confused. At this point in the migration, won't all collections wind up getting 'required' since `user_can_set_license?` checks against the value of `license_option`? Which was just added in the prior line? I expected this line to use logic similar to https://github.com/sul-dlss/happy-heron/pull/1043/files#diff-ca0f3357f3c92b4b6b2f37ffd7dee610174fbe9a9a1d8c4a9354c985c73ca7a1L28 but maybe I am more confused than I realize :laughing:

mjgiarlo (commented):
Question about DB migration. Otherwise looks ready to go.

db/migrate/20210208201246_add_license_option_to_collection.rb
jcoyne: Oh, yeah, I made this before i made that other change.

jcoyne: Fixed it.

mjgiarlo: @jcoyne :beers:

mjgiarlo (approved):

mjgiarlo:
@jcoyne feel free to merge once :green_heart:

PR #1045, opened 2021-02-09
allow collections to be saved with no contact emails

### LINKED ISSUE #1044: Cannot create a draft collection with an invalid or missing email address
You should be able to create a draft collection without submitting any data, but new validation code prevents this from happening with no user interface clues.

To reproduce:

1. Create a new collection.
2. Fill in only the title.  Click "Save as Draft".
3. It doesn't seem to do anything (but if you look at the rails console, you will see it returned a 422 and re-rendered the page but no messages).  
4. Click "Save as Draft" again and get an exception.

I believe this is because the collection object is not returning as valid from the `DraftCollectionForm` causing the 422 from here: https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/collections_controller.rb#L31-L35

The `DraftCollectionForm` needs to allow the entering of no email addresses to be valid (at the draft level) so it can be saved: https://github.com/sul-dlss/happy-heron/blob/main/app/forms/draft_collection_form.rb#L41-L47

peetucket:
## Why was this change made?

Fixes #1044 

- need to remove all validation from the `DraftCollectionForm` to have it work properly
- need to remove model level validation for `ContactEmail`, else you cannot save a draft collection with invalid emails (since it needs to persist to the database, and the model level validation will block it)
- add validation for contact emails to the `CollectionForm` to ensure it gets validated when you try and deposit
- added tests to verify you can save draft collections with missing or bogus contact emails
- updated and added tests 

**Notes**:
Seems to be some fallout from #893 and #842 .... you cannot save a draft collection when leaving the contact email blank or when submitting an invalid contact email.  The problem seems to stem from the fact that we do this validation on the `DraftCollectionForm` but our code no longer supports re-rendering the form once saved as draft (which we try to do now if validation fails).  This means we cannot currently do any validation when saving as draft, because any failed validation will result in an exception, but we currently do validation at the draft level for contact emails.

Even though this PR removes all contact email validation from the `DraftCollectionForm` and moves it to the `CollectionForm`, you will still have a problem if you add validation to the draft form later for any field, as our code currently does not support any validation display for an invalid draft save.

Tests and localhost browser

### DIFF
--- a/app/forms/collection_form.rb
+++ b/app/forms/collection_form.rb
@@ -5,6 +5,14 @@
 class CollectionForm < DraftCollectionForm
   extend T::Sig
 
+  collection :contact_emails, populator: ContactEmailsPopulator.new(:contact_emails, ContactEmail),
+                              prepopulator: ->(*) { contact_emails << ContactEmail.new if contact_emails.blank? } do
+    property :id
+    property :email
+    property :_destroy, virtual: true
+    validates :email, format: { with: Devise.email_regexp }
+  end
+
   validates :name, :description, :manager_sunets, :access, presence: true
   validates :contact_emails, length: { minimum: 1, message: 'Please add at least contact email.' }
   validates_with CollectionLicenseValidator
--- a/app/forms/draft_collection_form.rb
+++ b/app/forms/draft_collection_form.rb
@@ -41,7 +41,6 @@ class DraftCollectionForm < Reform::Form
     property :id
     property :email
     property :_destroy, virtual: true
-    validates :email, format: { with: Devise.email_regexp }, allow_blank: true
   end
 
   collection :related_links, populator: RelatedLinksPopulator.new(:related_links, RelatedLink),
--- a/app/forms/draft_collection_form.rb
+++ b/app/forms/draft_collection_form.rb
@@ -41,7 +41,7 @@ class DraftCollectionForm < Reform::Form
     property :id
     property :email
     property :_destroy, virtual: true
-    validates :email, format: { with: Devise.email_regexp }
+    validates :email, format: { with: Devise.email_regexp }, allow_blank: true
   end
 
   collection :related_links, populator: RelatedLinksPopulator.new(:related_links, RelatedLink),
--- a/app/models/contact_email.rb
+++ b/app/models/contact_email.rb
@@ -4,6 +4,4 @@
 # Models an email in the database
 class ContactEmail < ApplicationRecord
   belongs_to :emailable, polymorphic: true
-
-  validates :email, format: { with: Devise.email_regexp }, allow_blank: true
 end
--- a/spec/models/contact_email_spec.rb
+++ b/spec/models/contact_email_spec.rb
@@ -12,11 +12,6 @@ RSpec.describe ContactEmail, type: :model do
     expect(contact_email.email).to be_present
   end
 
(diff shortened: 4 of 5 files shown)


peetucket:
What you get if you to try to render the new form after saving as a draft with failed validation:

Screen Shot 2021-02-08 at 4 39 29 PM

app/components/collections/form_component.rb
```diff
@@ -8,6 +8,8 @@ class FormComponent < ApplicationComponent
 
     delegate :release_duration, to: :collection_form
 
+    alias draft_collection_path collection_path
```
peetucket, line 11: matches what we do in the work form

app/models/contact_email.rb
```diff
@@ -4,6 +4,4 @@
 # Models an email in the database
 class ContactEmail < ApplicationRecord
   belongs_to :emailable, polymorphic: true
-
-  validates :email, format: { with: Devise.email_regexp }, allow_blank: true
```
peetucket, line 8: need to remove model level validation for `ContactEmail`, else you cannot save a draft collection with invalid emails (since it needs to persist to the database, and the model level validation will block it)

spec/models/contact_email_spec.rb
```diff
@@ -12,11 +12,6 @@
     expect(contact_email.email).to be_present
   end
 
-  it 'has a valid email' do
```
peetucket, line 15: no more model level validation for emails

app/forms/draft_collection_form.rb
```diff
@@ -41,7 +41,6 @@ class DraftCollectionForm < Reform::Form
     property :id
     property :email
     property :_destroy, virtual: true
-    validates :email, format: { with: Devise.email_regexp }
```
peetucket, line 44: do not validate on the draft collection form

app/forms/collection_form.rb
```diff
@@ -5,6 +5,14 @@
 class CollectionForm < DraftCollectionForm
   extend T::Sig
 
+  collection :contact_emails, populator: ContactEmailsPopulator.new(:contact_emails, ContactEmail),
+                              prepopulator: ->(*) { contact_emails << ContactEmail.new if contact_emails.blank? } do
+    property :id
+    property :email
+    property :_destroy, virtual: true
+    validates :email, format: { with: Devise.email_regexp }
```
peetucket, line 13: validate contact email on the collection form, note: the regex will NOT allow blanks

spec/requests/create_collection_spec.rb
```diff
@@ -212,10 +220,32 @@
...
+            expect(response).to redirect_to(collection_path(collection))
+          end
+        end
+
+        context 'with an invalid contact email' do
+          let(:draft_collection_params) do
+            {
+              name: '',
+              description: '',
+              contact_emails_attributes: invalid_contact_email,
```
jcoyne, line 236: Since this let is only used in this place, can you move it into this context?

jcoyne (changes requested):

spec/requests/create_collection_spec.rb
peetucket: sure, actually just ditched the let since the value was small and fits inline

jcoyne (approved):

PR #1050, opened 2021-02-10
Version works so we can rollback to an old state

### LINKED ISSUE #858: Add the ability to revert/discard an already saved work in the database to the previous state it had when last deposited
Users need to be able to revert a version draft of a work, which would set it back to the last deposited state, thereby removing any changes they made in the version.

Since some changes take effect immediately, and since the work is immediately saved in the database when the user clicks the save draft button, this may require some versioning of database rows and other changes.

Note: collection reversion now in #1055

jcoyne:
## Why was this change made?
This allows us to revert changes to an old version. Fixes #858

Diagram of the new model:
https://docs.google.com/drawings/d/1LJLWj-psVTTAtDpjb6Jpse2LtqTk8a-XhvkPfUB-V4w


jcoyne:
Superseded by #1101

PR #1051, opened 2021-02-10
Adds Approximate check-boxes for Creation Date  

### LINKED ISSUE #938: Add approximate date check boxes to the creation date options
Design doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.uqrdv5w5a8ip
wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/428088403

jermnelson:
## Why was this change made?
Fixes #938 

Screen Shot 2021-02-10 at 11 23 10 AM

New component tests

n/a


app/components/works/dates_component.html.erb
```diff
@@ -65,6 +70,10 @@
               id: 'work_created_range_start_day',
               class: "form-control" %>
         </div>
+        <div>
+         <%= check_box_tag 'work[created_range(approx0)]', 1, created_range_start_approximate? %>
```
peetucket, line 74: just curious if you tried using a value of `true` instead of `1`, which maybe could then let you do `if uncertain` instead then `if uncertain == '1'` below, which seems a bit clearer

jermnelson: Thanks @peetucket, lasted commit now uses `true` instead of 1.

peetucket:
LGTM

peetucket (approved):

PR #1057, opened 2021-02-11
fix tests

peetucket:
## Why was this change made?

Tests are broken on main ... this fixes them

The tests!

### DIFF
--- a/spec/requests/dashboards_spec.rb
+++ b/spec/requests/dashboards_spec.rb
@@ -51,7 +51,7 @@ RSpec.describe 'Dashboard requests' do
 
   context 'when user is a collection manager and there is a collection in progress' do
     let(:collection) do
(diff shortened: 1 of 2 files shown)


spec/requests/dashboards_spec.rb
```diff
@@ -51,7 +51,7 @@
 
   context 'when user is a collection manager and there is a collection in progress' do
     let(:collection) do
-      create(:collection, :first_draft, creator: user, managers: [collection_manager], name: 'Happy little collection')
+      create(:collection, :first_draft, creator: user, managed_by: [collection_manager], name: 'Happy collection')
```
peetucket, line 54: collection is no longer little, and rubocop no longer complains that the line is too long

mjgiarlo (approved):

PR #1067, opened 2021-02-11
Use captions to describe the table rather than table headers

### LINKED ISSUE #1018: Accessibility issue on View deposit page - header and cells associated
image

jcoyne:
## Why was this change made?
Fixes #1018


mjgiarlo (commented):
Does this change the look & feel? If so, can you include at least one before/after screenshot to show how it changes? Otherwise LGTM.

jcoyne:
It looks no different to sighted users, it's very different on a screen reader.

mjgiarlo (approved):

PR #1072, opened 2021-02-12
Introduce subtype requirements for mixed material-type works

### LINKED ISSUE #609: Require two subtype selections when a depositor chooses "Mixed Materials" as the primary work type and require a subtype for "Music"
"Mixed Materials" should contain a mixture of materials, and we want to know more than that it's "mixed materials" so we should require them to choose two things.

Other main types do not require subtype selections, except for "Other" which requires an entry in the free text box.

Also, Music must have a subtype. See Arcadia's note below. For Music, this subtype must be from a select list of the type and subtype terms.

Make sure this is the behavior in:

- [x] Work type modal
- [x] Work form

mjgiarlo:
Fixes #609

## Why was this change made?

This commit adds the desired behavior to the existing mixed material work type, which is unique in that it requires two or more primary subtypes to be selected.

Includes:

* Make work type stimulus controller able to render a subheader for mixed material-type works
* Change subtype validator so it can enforce mixed material-type works' unique constraint (see above)

### Screenshot: Work form 

form

### Screenshot: Work type modal

modal

CI

None

### DIFF
--- a/spec/requests/create_work_spec.rb
+++ b/spec/requests/create_work_spec.rb
@@ -100,6 +100,25 @@ RSpec.describe 'Create a new work' do
         end
       end
 
+      context 'with a valid work_type and not enough valid required subtypes' do
+        it 'redirects to dashboard with an informative flash message' do
+          get "/collections/#{collection.id}/works/new?work_type=mixed+material&subtype%5B%5D=Data"
+          expect(response).to redirect_to(dashboard_path)
+          follow_redirect!
+          expect(response).to have_http_status(:ok)
+          expect(response.body).to include 'Invalid subtype value for work_type &#39;mixed material&#39;: Data'
+        end
+      end
+
+      context 'with a valid work_type and enough valid required subtypes' do
+        it 'renders the form' do
+          get "/collections/#{collection.id}/works/new?work_type=mixed+material&subtype%5B%5D=Data&subtype%5B%5D=Text"
+          expect(response).to have_http_status(:ok)
+          expect(response.body).to include 'Data'
+          expect(response.body).to include 'Text'
+        end
+      end
+
       context 'with a work_type that is missing a required user-supplied subtype' do
         it 'redirects to dashboard with an informative flash message' do
           get "/collections/#{collection.id}/works/new?work_type=other"
--- a/app/components/works/subtypes_component.html.erb
+++ b/app/components/works/subtypes_component.html.erb
@@ -14,7 +14,10 @@
   <% else %>
     <% if music_type? %>
       <h6>Select at least one term below:</h6>
+    <% elsif mixed_material_type? %>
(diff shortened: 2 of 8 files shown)


spec/requests/create_work_spec.rb
```diff
@@ -100,6 +100,25 @@
         end
       end
 
+      context 'with a valid work_type and not enough valid required subtypes' do
+        it 'redirects to dashboard with an informative flash message' do
+          get "/collections/#{collection.id}/works/new?work_type=mixed+material&subtype%5B%5D=Data"
+          expect(response).to redirect_to(dashboard_path)
+          follow_redirect!
+          expect(response).to have_http_status(:ok)
+          expect(response.body).to include 'Invalid subtype value for work_type &#39;mixed material&#39;: Data'
```
jcoyne, line 109: Is this something that is useful to a user? It seems like they would be stuck at this point. I would be hesitant to create a test that says "the application must behave this way", unless this is the way the PO has requested it to behave.

mjgiarlo: Maybe not, but this batch of tests also serves as coverage of the WorkSubtypeValidator which is not directly tested in its own set of unit tests.

jcoyne: Can this move to a unit test on WorkSubtypeValidator then?  I want to avoid this being something that confuses us in the future.

mjgiarlo: @jcoyne I'm happy to put up a small, new, follow-up PR that changes how we test the subtype validator, sure. Here's what I'm thinking re: scope---let me know if you agree or have any tweaks:

We should have a new work subtype validator spec that adds unit tests for the validator (to test all the validation-specific behaviors), and the validation-related tests in this requests spec can be reducde to one or two cases that test that the work subtype validator is wired up to the request without needing to test every validation behavior.

What do you think?

jcoyne: I'd rather not have any tests with regard to how the request should behave if it's invalid without talking to @astridu and @amyehodge about what their expectations are in this case.  I think if we have this path, the user should have a way out of this dead end, or we should just let it throw an error.

jcoyne: At a minimum drop a comment on this test that says "We have no requirement for this UX, but we want to show that it only accepts valid inputs"

mjgiarlo: @jcoyne OK. We already have this behavior in the app. We can strip it out, of course. I'm reluctant to initiate making changes to that behavior this late in the work cycle, and would like to focus on test strategy. Instead of dropping a comment on this test, how about I do what I proposed above (with your delta of removing all request specs related to validation since not supported by design or P.O. requirements) in a follow-up PR?

PR #1080, opened 2021-02-12
Allow works to have a small number of legacy licenses

### LINKED ISSUE #1047: Support a list of licenses that can’t be chosen but can be displayed
This is necessary for migrating content from Hydrus that has licenses that we are not supporting in H2.

When items with unsupported licenses are migrated, the depositor should see the license they previously chosen, e.g. a CC 3.0 license, displayed on the item show page. When the user versions this item, they will be required to choose one of the valid H2 license options in order to be able to deposit the new version.

Content in draft that has a license assigned will be required to choose a new license, unless that license was set by the collection manager already.

mjgiarlo:
Fixes #1047

## Why was this change made?

This change is intended to help migrate Hydrus 1 data into Hydrus 2, and to display migrated works with their current (CC 3.0) licenses. It allows a small set of displayable licenses to show both in the work show page and on the work form. Before a new version of such a work may be deposited again, a new H2 license must be selected.

Includes:

* Remove license validation from work model
  * We are already validating licenses in the work form so this is not necessary and is counter to our approach to validation where we favor form validation over modal validation. This also allows us to migrate *any* license from Hydrus 1.0.

### Screenshot: Show legacy license in show page

license_show

### Screenshot: Display legacy license in form

selected_license

### Screencast: Show dropdown behavior in form

Peek 2021-02-12 14-58

CI, local browser

None

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,6 +61,9 @@ RSpec/ScatteredSetup:
 
 #### Turning on New Rules
 
+Gemspec/DateAssignment: # (new in 1.10)
+  Enabled: true
+
 Layout/BeginEndAlignment: # (new in 0.91)
   Enabled: true
 Layout/EmptyLinesAroundAttributeAccessor: # (new in 0.83)
@@ -245,6 +248,8 @@ Style/ExponentialNotation: # (new in 0.82)
 Style/GlobalStdStream: # (new in 0.89)
   Enabled: true
 Style/HashAsLastArrayItem: # (new in 0.88)
+  Enabled: true
+Style/HashConversion: # (new in 1.10)
   Enabled: true
 Style/HashEachMethods: # (new in 0.80)
   Enabled: true
--- a/app/components/works/detail_component.rb
+++ b/app/components/works/detail_component.rb
@@ -62,7 +62,7 @@ module Works
 
     sig { returns(String) }
     def license
-      License::ID_LABEL_HASH[work.license]
+      License.label_for(work.license)
     end
 
     sig { returns(String) }
--- a/app/components/works/license_component.html.erb
+++ b/app/components/works/license_component.html.erb
@@ -5,7 +5,7 @@
       <%= form.label :license, class: 'col-sm-2 col-form-label' %>
       <div class="col-sm-5">
         <%= form.select :license,
-                        grouped_options_for_select(License.grouped_options, license),
+                        grouped_options_for_select(License.grouped_options(license), license),
                         {}, class: "form-select" %>
       </div>
       <div class="col-sm-5">
--- a/app/models/license.rb
+++ b/app/models/license.rb
@@ -6,8 +6,8 @@
 class License
   extend T::Sig
 
-  # valid values
-  ID_LABEL_HASH = {
+  # valid, selectable licenses
+  SELECTABLE_LICENSES = {
     'CC0-1.0' => 'CC0-1.0',
     'CC-BY-4.0' => 'CC-BY-4.0 Attribution International',
     'CC-BY-SA-4.0' => 'CC-BY-SA-4.0 Attribution-Share Alike International',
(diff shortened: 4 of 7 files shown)


app/components/works/detail_component.rb
```diff
@@ -62,7 +62,7 @@ def access
 
     sig { returns(String) }
     def license
-      License::ID_LABEL_HASH[work.license]
+      License.label_for(work.license)
```
mjgiarlo, line 65: This is an internal refactoring to use the public interface of the License model, preferring message-sending to using constants. Has no effect outside the component.

app/components/works/license_component.html.erb
```diff
@@ -5,7 +5,7 @@
       <%= form.label :license, class: 'col-sm-2 col-form-label' %>
       <div class="col-sm-5">
         <%= form.select :license,
-                        grouped_options_for_select(License.grouped_options, license),
+                        grouped_options_for_select(License.grouped_options(license), license),
```
mjgiarlo, line 8: Pass the license to the method so it knows whether to show the merely displayable licenses as well as the selectable ones. (Don't show them unless there's a legacy license.)

app/models/license.rb
```diff
@@ -6,8 +6,8 @@
 class License
   extend T::Sig
 
-  # valid values
-  ID_LABEL_HASH = {
+  # valid, selectable licenses
+  SELECTABLE_LICENSES = {
```
mjgiarlo, line 10: Rename to clarify distinction between selectable licenses and displayable ones.

app/models/work.rb
```diff
@@ -18,9 +18,6 @@ class Work < ApplicationRecord
   has_many :events, as: :eventable, dependent: :destroy
 
   validates :state, presence: true
-  validates :license, presence: true, inclusion: { in: License.license_list }
-  validates :subtype, work_subtype: true
-  validates :work_type, presence: true, work_type: true
```
mjgiarlo, line 23: We already have these validations in the form, which is more consistent with our validation philosophy. No need to have them here.

peetucket:
LGTM, not sure if you want another approver before merging.

peetucket (approved):

app/models/license.rb
```diff
@@ -33,23 +33,44 @@ class License
...
-    GROUPINGS.map do |group|
+  sig { params(license: T.nilable(String)).returns(T::Array[T::Array[T.any(String, T::Array[String])]]) }
+  def self.grouped_options(license = nil)
+    options = GROUPINGS.map do |group|
       [group.fetch(:label), group.fetch(:options).map { |license_id| [label_for(license_id), license_id] }]
     end
+    return options unless license.in?(DISPLAYABLE_LICENSES)
+
+    # If the current license is merely displayable, add the displayable licenses to the dropdown as disabled options
+    options.append(
```
jcoyne, line 61: I don't think we want to add this here, this would be in the selectable list. We want the user to select a new valid license if they make a change.

app/models/work.rb
jcoyne: I think it's a good idea to have them in both places. We want to ensure the licences in the model is SELECTABLE_LICENSES or DISPLAYABLE_LICESES, but we want those in the form to only be the SELECTABLE_LICENSES.  We want to make sure we aren't importing any invalid data (which bypasses the form).

jcoyne (changes requested):

app/models/license.rb
mjgiarlo: @jcoyne It shouldn't because of the check on line 58. This only appends displayable licenses if one of them is already selected.

jcoyne: 👍

app/models/work.rb
mjgiarlo: @jcoyne I'm open to making this change and it's easy enough. This leaves me fuzzy on the following question: how do we decide where a validation ought to live? What's the guideline?

I hear you regarding not wanting bogus licenses in the database. If that's true, doesn't that imply we want a bunch more of our validations down inthe models? (Less important sidebar: And if they move to the models, would we even need them in the forms anymore?) It'd be great to get a shared sense of this, so I'm copying @sul-dlss/infrastructure-team

peetucket: I am interested in this question too, having previously moved validations around between forms and models.  Our use of draft vs regular forms to deal with validations needed sometimes but not other times just adds to the complexity (and maybe the original driver behind moving most if not all validation outside the model).

jcoyne: I think you want to validate things you put in the database (for migrations, etc).  You also want to validate some of the same things on the form (and possibly some more things).  I don't think there is a simple rule of thumb for this. Validations are complex and context dependent.

mjgiarlo: Yeah, this is rather muddy for me, still. I'm happy to embrace the muddiness and rejigger this, though.

mjgiarlo: Done.

mjgiarlo:
@jcoyne this is ready for another look.

.rubocop.yml
```diff
@@ -246,6 +249,8 @@ Style/GlobalStdStream: # (new in 0.89)
   Enabled: true
 Style/HashAsLastArrayItem: # (new in 0.88)
   Enabled: true
+Style/HashConversion: # (new in 1.10)
+  Enabled: true
```
mjgiarlo, line 253: Rubocop changed while I was working on this PR, so I included the new rules here.

app/models/work.rb
```diff
@@ -18,9 +18,7 @@ class Work < ApplicationRecord
   has_many :events, as: :eventable, dependent: :destroy
 
   validates :state, presence: true
-  validates :license, presence: true, inclusion: { in: License.license_list }
-  validates :subtype, work_subtype: true
```
jcoyne, line 22: I'm not sure why these disappeared in this PR . can they be restored?   Maybe remove in a different PR?

jcoyne (changes requested):

app/models/work.rb
mjgiarlo: Ah, these disappeared because of the conversation we had re: "validation in forms or models" we had on take 1 of the PR. So, I will restore these and push.

mjgiarlo: @jcoyne Fixed. One more look?

jcoyne (approved):

PR #1081, opened 2021-02-12
Consider subtypes when showing that description is complete in the da…

### LINKED ISSUE #971: "Description" section of progress bar on dashboard should consider entries for work subtypes when they are required
If the item type is "Music", "Mixed Materials", or "Other", the "Description" section of the Progress Bar should not turn green until this section, along with the abstract and keywords, has the necessary entry.

The necessary entries are:
- for Music, at least one of the required list of subtypes selected (see list in https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.xatek3rhlho, this is a subset of the full list of terms)
- for Mixed Materials, at least two terms must be selected
- for Other, the box must have an entry

jcoyne:
…shboard progress

## Why was this change made?
Fixes #971


app/components/dashboard/deposit_progress_component.rb
```diff
@@ -6,30 +6,31 @@ module Dashboard
   class DepositProgressComponent < ApplicationComponent
     sig { params(work: Work).void }
     def initialize(work:)
-      @work = work
+      @form = T.let(WorkForm.new(work), WorkForm)
+      @form.valid?
```
mjgiarlo, line 10: I was a bit surprised to see `#initialize` end with a question-mark method. What does this do? Populate the errors array? If needed, can you drop a comment to that effect?

app/models/work_type.rb
```diff
@@ -9,6 +9,9 @@ class InvalidType < StandardError; end
 
   MINIMUM_REQUIRED_MUSIC_SUBTYPES = 1
   MINIMUM_REQUIRED_MIXED_MATERIAL_SUBTYPES = 2
+  MIXED_MATERIAL = 'mixed material'
+  MUSIC = 'music'
+  OTHER = 'other'
```
mjgiarlo, line 14: :+1:

mjgiarlo (approved):
Would like a comment added. Approving pending consideration of that change.

PR #1089, opened 2021-02-16
Add tooltips to the select worktype mockup

### LINKED ISSUE #1030: Add tooltips to the Work Type modal
These are the little blue circles with the white "i" in them.

Note that the "which collection will you deposit to" part that is in the tooltips doc does not exist in the work type modal (just the first use modal), so you don't need to worry about that one here.

See Tooltips doc for text: https://docs.google.com/document/d/1gjvqYXmwMpMGWTdZynoHtxDgQVTt9fGn7QGVE4uhXXQ/edit#heading=h.v2a0c7kuvnie

See wireframes for locations: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/431445056

jcoyne:
Depends on #1088 

## Why was this change made?
Fixes #1030

### DIFF
--- a/app/components/popover_component.html.erb
+++ b/app/components/popover_component.html.erb
@@ -1,0 +1,3 @@
+<a tabindex="0" data-controller="popover" data-bs-placement="top" data-bs-content="<%= text %>">
+  <span class="fas fa-info-circle"></span><span class="visually-hidden">more information</span>
+</a>
--- a/app/components/popover_component.rb
+++ b/app/components/popover_component.rb
@@ -1,0 +1,13 @@
+# typed: false
+# frozen_string_literal: true
+
+# Draws a bootstrap popover icon.
+class PopoverComponent < ApplicationComponent
+  def initialize(key:)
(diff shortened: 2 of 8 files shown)


app/components/popover_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<a tabindex="0" data-controller="popover" data-bs-placement="top" data-bs-content="<%= text %>">
```
jermnelson, line 1: Could you wrap this anchor with `sup` to shrink the size and make it superscript?

jermnelson (changes requested):
Looks good; the only comment I have is that icon is large (component could be wrapped with `sup` or changed with css) and not blue as per the design.

app/components/popover_component.html.erb
jcoyne: I have changes that adjust the size and color in #1090, however the designs do not call for superscript.

jermnelson (approved):

PR #1138, opened 2021-02-18
Automatic citation now generates year based on current year, embargo date, or publication year

### LINKED ISSUE #655: Dates need to be included in the default citation according to the description in the design doc.
These are the rules for including dates in the default citation:

"Rules for the date:

- If there is a publication date entered, use  that date, but only the year
- If there is no publication date but there is an embargo date, use that date, but the year only
- Otherwise use the current year"

When there is a publication date, it is getting included, but the entire date is there with month and day and this should just be the year.

When there is an embargo date, the embargo date is not being included. Again, this should be year only.

When there is no embargo date or publication date, the year of deposit should be included. Again this should be year only.

jermnelson:
## Why was this change made?
Fixes #655 

New feature test of automatic citation 

n/a

### DIFF
--- a/app/javascript/controllers/auto_citation_controller.js
+++ b/app/javascript/controllers/auto_citation_controller.js
@@ -3,7 +3,7 @@ import { Controller } from "stimulus";
 export default class extends Controller {
   static targets = ["titleField", "manual", "auto", "switch",
     "contributorFirst", "contributorLast", "contributorRole", "contributorOrg",
-    "year", "month", "day", "embargoYear"];
+    "year", "embargoYear"];
 
   connect() {
     this.purl = this.data.get("purl") || ":link will be inserted here automatically when available:" // Use a real purl on a persisted item or a placeholder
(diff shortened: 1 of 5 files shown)


app/javascript/controllers/auto_citation_controller.js
```diff
@@ -73,33 +73,17 @@ export default class extends Controller {
...
-  get month() {
-    return this.monthTarget.value
-  }
-
-  get day() {
-    return this.dayTarget.value
-  }
-
   get date() {
     const date = new Date();
```
jcoyne, line 83: Do we still need the `Date` code since we're only dealing in years now?

jcoyne (changes requested):

app/javascript/controllers/auto_citation_controller.js
jermnelson: I think so since we want to return the current year if there is not an `embargoYear` or `publication_year`.

jcoyne (approved):

PR #1144, opened 2021-02-19
Add the what's changing element element for version description to works

### LINKED ISSUE #454: Create "What's changing" UI element

aaron-collier:
## Why was this change made?

Fixes #454 

This adds the required "What's changing?" UI component to capture version description when editing a work. If the work is currently in a deposited state, then a new version will be created on edit and the version description component should be blank. However, if the user is continuing to edit a version_draft, the version description should be populated with the existing value.

Screen Shot 2021-02-19 at 11 23 33 AM

Deposited state:

Screen Shot 2021-02-19 at 11 23 42 AM

Version_draft state:

Screen Shot 2021-02-19 at 11 23 54 AM

Unit and updated feature tests

N/A


app/components/works/form_component.rb
```diff
@@ -33,5 +33,12 @@ def work
     def work_version
       work_form.model.fetch(:work_version)
     end
+
+    sig { returns(T::Boolean) }
+    def version_description_required?
+      return true if %w[deposited version_draft].include? work_version.state
```
mjgiarlo, line 39: I have two comments about this method. 

First, I believe `[]#include?` is already a method that returns a boolean, so I think you could consider removing the explicit `true`/`false` here and make this method a one-liner: `%w[states go here].include?(work_version.state)`. What would you think about that approach?

mjgiarlo: Second, can you say a bit more about why only `deposited` and `version_draft` are included here? Presumably we don't want this component to render with `new` or `initial_draft` states because requiring a description for the first version is not a requirement we have for H2. And with `depositing` and `pending_approval`, the item cannot be edited by the depositor, so no need for the box to render. What about `rejected`, though?

mjgiarlo: Third, instead of a conditional rendering of `Works::VersionDescriptionComponent` within the `Works::FormComponent` HTML, what would you think about removing the conditional and moving this logic to `Works::VersionDescriptionComponent#render?`?

app/javascript/stylesheets/palette.scss
```diff
@@ -1,4 +1,5 @@
 $alabaster: #f2f1eb;
+$ibis-white: #f3e7e7; // light pinkish hue used in what's changing box
```
mjgiarlo, line 2: This color looks close to what we're using for the background of the currently checked work type in the work type modal. I believe that is set to `rgba($primary, 0.1)`: https://github.com/sul-dlss/happy-heron/blob/main/app/javascript/stylesheets/toggle.scss#L7 I wonder if we should standardize on one or the other or if we're OK with this varying. Cc: @jcoyne

app/models/work_version.rb
```diff
@@ -123,4 +123,11 @@ def published_edtf
   def created_edtf
     EDTF.parse(super)
   end
+
+  sig { returns(T.nilable(String)) }
+  def description
+    return nil unless %w[version_draft].include? state
```
mjgiarlo, line 129: Can you say more about what this method is doing?

aaron-collier: Yes, I can add a comment if that would help. Essentially, we do not want the description field to be populated when editing a work that isn't in a draft state. If we're opening an object in a draft state, whatever was entered in the what's changing box as part of the earlier edit, we do want to populate that. But if one is starting a new version, we want the field blank.

app/components/works/form_component.rb
aaron-collier: Good suggestions, will apply shortly.

mjgiarlo: Thank you!

app/models/work_version.rb
mjgiarlo: OK, gotcha. What would you think about nulling it out when new versions are instantiated? Like here: https://github.com/sul-dlss/happy-heron/blob/3e679fd9a5299035c1a90866f52897665af85b95/app/controllers/works_controller.rb#L139-L145 with something like

```ruby
work_version.description = nil
```

mjgiarlo: Let me know if I'm grokking. I may be turned around about this. Happy to pair this afternoon too if you have time. I'm free all PM.

aaron-collier: I don't think that will work as, from my understanding, the new version is not initiated the form is edited and save or save as draft is clicked... at least that's what I was able to see in my local testing. Is that incorrect?

mjgiarlo: I'm not sure I'm following. :pear: ?

app/components/works/version_description_component.rb
```diff
@@ -0,0 +1,22 @@
+# typed: true
+# frozen_string_literal: true
+
+module Works
+  # Renders a widget for enetering a version description
```
mjgiarlo, line 5: Typo: `s/enetering/entering/`

app/javascript/stylesheets/palette.scss
mjgiarlo: @aaron-collier Yeah, don't know how much I care about this so :man_shrugging:

mjgiarlo (approved):
Other than the typo, LGTM!

mjgiarlo:
Are the commits meaningful to ongoing development or more a by-product of development? Will help determine whether to squash-merge or merge.

app/components/works/version_description_component.rb
aaron-collier: I have to make sure and have at least one typo to keep you on your toes.

Fixed.

aaron-collier:
@mjgiarlo byproduct, feel free to squash.

app/components/works/version_description_component.rb
mjgiarlo: Thank you for giving me something to do!

Issue #1110, opened 2021-02-17
Split collection edit page into two pages, one for info that gets versioned and one for settings


amyehodge:
Design Doc
Collection Details (content that is versioned): https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.m80ipne8rlzk
Collection Settings (content that is not versioned): https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.pbwvx7aodi0w

Wireframes
Collection Details (content that is versioned): https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/421463079
Collection Settings (content that is not versioned): https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/444874100

amyehodge:
This work should be done in conjunction with #794 Add "Discard draft" option to Collection Edit page

PR #1158, opened 2021-02-20
Version collections

jcoyne:
## Why was this change made?

Supports #1110


app/forms/draft_collection_form.rb
```diff
@@ -103,16 +110,30 @@ def field_to_users(field)
...
+      collection.update(head: model.fetch(:collection_version))
+    end
+  end
+
+  delegate :persisted?, :to_param, to: :collection
+
+  # def self.model_name
+  #   byebug
+  #   super
+  # end
```
mjgiarlo, line 138: Cruft?

app/forms/draft_collection_form.rb
```diff
@@ -5,45 +5,48 @@
 class DraftCollectionForm < Reform::Form
   extend T::Sig
   feature EmbargoDate
+  include Composition
+  model 'collection'
```
mjgiarlo, line 9: I know these were in the work version PR as well, but can you drop a comment here about what's going on?

app/forms/draft_collection_form.rb
```diff
@@ -103,16 +110,30 @@ def field_to_users(field)
...
   end
+
+  def save_model
+    Work.transaction do
+      super
+      collection.update(head: model.fetch(:collection_version))
+    end
+  end
+
+  delegate :persisted?, :to_param, to: :collection
```
mjgiarlo, line 133: Can you say more about why these are needed?

app/models/concerns/aggregate_associations.rb
```diff
@@ -0,0 +1,16 @@
+# typed: false
+# frozen_string_literal: true
+
+# Provides a helper method for associations
+module AggregateAssociations
+  extend ActiveSupport::Concern
+
+  class_methods do
+    # Which assocations are has_many?
```
mjgiarlo, line 9: Typo: `s/assocations/associations/`

app/models/concerns/aggregate_associations.rb
```diff
@@ -0,0 +1,16 @@
+# typed: false
+# frozen_string_literal: true
+
+# Provides a helper method for associations
```
mjgiarlo, line 4: Would you consider expanding the scope of this comment to say more about how/why we need it?

app/services/new_collection_version_parameter_filter.rb
```diff
@@ -0,0 +1,25 @@
+# typed: true
+# frozen_string_literal: true
+
+# Filters the passed in parameters so they create new objects rather than point at objects for a past version
+# TODO: there is lots of duplicate code here from NewVersionParameterFilter
```
mjgiarlo, line 5: :ok_hand:

db/migrate/20210219142356_create_collection_versions.rb
```diff
@@ -0,0 +1,33 @@
...
+      version = CollectionVersion.create(state: collection.state,
+                                         name: collection.name,
+                                         description: collection.description,
+                                         collection: collection,
+                                         version: collection.version)
+      collection.update(head: version)
+      RelatedLink.where(linkable_type: 'Collection', linkable_id: 5)
+        .update_all(linkable_type: 'CollectionVersion', linkable_id: version.id)
+      ContactEmail.where(emailable_type: 'Collection', emailable_id: 5)
+        .update_all(emailable_type: 'CollectionVersion', emailable_id: version.id)
```
mjgiarlo, line 24: What is the `5` here?

app/forms/draft_collection_form.rb
jcoyne: yes

db/migrate/20210219142356_create_collection_versions.rb
jcoyne: 5 stands for `collection.id` (fixed this)

mjgiarlo:
Can you include updated state diagrams in this PR too and update the wiki page for that? If you'd rather handle in a follow-up PR, that'd be OK, too, and I'm happy to help. LMK.

app/controllers/collections_controller.rb
```diff
@@ -11,52 +11,71 @@ def new
...
+      after_save(collection: collection, collection_version: collection_version)
     else
       render :new, status: :unprocessable_entity
     end
   end
 
+  def edit
+    collection = Collection.find(params[:id])
+    collection_version = collection.head
+    authorize! collection_version
```
jermnelson, line 36: When running locally with `REMOTE_USER=auser@stanford.edu ROLES=dlss:hydrus-app-administrators bin/rails server` current_user is failing the `authorize!` call and redirecting to the dashboard when trying to edit a Collection. Is there another role I should add?

jcoyne: I fixed this by tweaking the CollectionVersionPolicy to checking for the object to be `updatable?` rather than `can_update_metadata?`

db/structure.sql
```diff
@@ -203,13 +203,46 @@ ALTER SEQUENCE public.attached_files_id_seq OWNED BY public.attached_files.id;
...
+ALTER SEQUENCE public.collection_versions_id_seq OWNED BY public.collection_versions.id;
+
+
+--
+-- Name: collections; Type: TABLE; Schema: public; Owner: -
+--
+
+CREATE TABLE public.collections (
+    id bigint NOT NULL,
     release_option character varying,
```
jermnelson, line 246: Is there a reason why properties like `release_option`, `release_duration`, `access`, `required_license`, are in the collection model verses in the collection_version model?

jermnelson (approved):
LGTM; just have a question regarding what properties are in collection verses the collection_version.

db/structure.sql
jcoyne: Yes, because those are now just settings and they don't go in the repository. They are not versioned.

PR #1171, opened 2021-02-23
Switch to lograge for creating machine parsable logs

jcoyne:
## Why was this change made?

So we can search the logs in cloudwatch.


Gemfile
```diff
@@ -73,3 +73,5 @@ gem 'view_component', '~> 2.18'
 gem 'webpacker', '~> 5.0'
 gem 'whenever'
 gem 'zipline', '~> 1.3'
+
+gem 'lograge', '~> 0.11.2'
```
mjgiarlo, line 77: Not a huge deal, and I realize you prolly added this via `bundle add`, but would you consider popping this above so the Gemfile is in alpha order again?

mjgiarlo (approved):

aaron-collier (approved):

PR #1180, opened 2021-02-24
Embargo Validator now check's Collection Duration

### LINKED ISSUE #923: Display correct error message when embargo date exceeds collection manager imposed end date
I set a collection to only allow embargoes up to a year out. When I tried to deposit an item with an embargo farther out than that, an error message was displayed at the top of the page with inaccurate info indicating the limit was 3 years. The error message needs to match the limitation that was set.

jermnelson:
## Why was this change made?
Fixes #923 

New unit tests

n/a

### DIFF
--- a/app/validators/embargo_date_validator.rb
+++ b/app/validators/embargo_date_validator.rb
@@ -6,7 +6,7 @@ class EmbargoDateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.nil?
 
-    collection = Collection.where(id: record.model[:work][:collection_id]).first
+    collection = record.model.fetch(:work).collection
     return if collection.blank? || collection.release_duration.blank?
 
     valid_embargo_value(record, attribute, value) if valid_embargo_collection(collection, record, attribute, value)
--- a/app/validators/embargo_date_validator.rb
+++ b/app/validators/embargo_date_validator.rb
@@ -6,6 +6,27 @@ class EmbargoDateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.nil?
 
+    collection = Collection.where(id: record.model[:work][:collection_id]).first
+    return if collection.blank? || collection.release_duration.blank?
+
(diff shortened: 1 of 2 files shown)


app/validators/embargo_date_validator.rb
```diff
@@ -6,6 +6,27 @@ class EmbargoDateValidator < ActiveModel::EachValidator
   def validate_each(record, attribute, value)
     return if value.nil?
 
+    collection = Collection.where(id: record.model[:work][:collection_id]).first
```
jermnelson, line 9: Not sure if this is the best way to retrieve the Collection associated with the work version.

app/validators/embargo_date_validator.rb
```diff
@@ -6,6 +6,27 @@ class EmbargoDateValidator < ActiveModel::EachValidator
...
+    return if collection.blank? || collection.release_duration.blank?
+
+    valid_embargo_value(record, attribute, value) if valid_embargo_collection(collection, record, attribute, value)
```
jermnelson, line 12: Only calls the method if the collection release duration fails preventing duplicate errors if the collection's release duration is `3 years`

app/validators/embargo_date_validator.rb
```diff
@@ -6,6 +6,27 @@ class EmbargoDateValidator < ActiveModel::EachValidator
...
+  end
+
+  private
+
+  def valid_embargo_collection(collection, record, attribute, value)
+    year_match = collection.release_duration.match(/(\d+)/)
```
jermnelson, line 18: Since the `collection.release_duration` is no longer a date but a string, this regex seemed to be the best option but open to suggestions on a better method of extract the number of years.

app/validators/embargo_date_validator.rb
jcoyne: You ought to be able to do `collection = record.model.fetch(:work).collection`

jermnelson: Thanks! Now updated.

jcoyne (approved):

Issue #1191, opened 2021-02-24
Reserve a PURL function results in errors


amyehodge:
When trying to reserve a PURL on an H2-created collection, I entered a title and clicked Submit, and it loaded the work edit page (which I don’t understand) and reported this error:

![Screen Shot 2021-02-24 at 2.20.17 PM.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/48e89385-a350-431c-80a2-66a61a628d09)

Then when I tried to reserve a PURL on a migrated collection, it didn't open the form (this seems right) and appeared to reload the dashboard, but I didn’t see the reserved item listed in the collection. And I appeared to have triggered this in HB:

line 53 of [PROJECT_ROOT]/app/forms/draft_work_form.rb: deserialize_embargo
line 41 of [PROJECT_ROOT]/app/forms/draft_work_form.rb: deserialize!
line 35 of [PROJECT_ROOT]/app/controllers/works_controller.rb: create

https://app.honeybadger.io/projects/77112/faults/76941540

jmartin-sul:
In trying to reproduce this error, I ran into something different:  the deposit failed (because an invalid cocina model was generated, per @jcoyne on slack as well as [this comment](https://github.com/sul-dlss/happy-heron/pull/989#discussion_r579609141) on the original PR).  Then, because no state transitions existed that allowed `update_metadata` before deposit was complete (which would get the `WorkVersion` to `purl_reserved`), I was unable to select a type for my wedged PURL reservation.

* test object on stage:  https://sul-h2-stage.stanford.edu/works/9365
* HB error for missing state transition:  https://app.honeybadger.io/projects/77112/faults/76952510 (`StateMachines::InvalidTransition: Cannot transition state via :update_metadata from :reserving_purl (Reason(s): State cannot transition via "update metadata")`).
* HB for invalid cocina model (the placeholder work type is invalid):  https://app.honeybadger.io/projects/77112/faults/76946844
```
[...]

PR #1202, opened 2021-02-25
state differentiation to allow update_metadata while purl res background/async work happens

jmartin-sul:
## Why was this change made?

address [this issue](https://github.com/sul-dlss/happy-heron/issues/1191#issuecomment-785688674) discovered while trying to reproduce #1191.

existing test suite for regressions.  will also deploy to stage in the morning to test whether it resolves the problem encountered there.  will also add unit tests to cover new state transitions.

the state machine diagram


jmartin-sul:
superseded by #1223 and its yet-to-be-implemented companion PR (the latter will more directly address what this PR does, by preventing editing of the work before the `purl_reserved` state is hit, saving us the added complexity of all the new state transitions of this PR).

PR #1214, opened 2021-02-25
Update sneakers to the head of the master branch

jcoyne:
## Why was this change made?

See if this allows systemd to stop the process cleanly.


ndushay:
can this be closed and branch deleted?  Please do it if so.

PR #1218, opened 2021-02-25
Update resource type and genre headings

arcadiafalcone:
## Why was this change made?
Map all the resource types

Specs need to be updated first

n/a


peetucket:
Closed in favor of #1251 which has the relevant commits picked out of this and squashed down to one

PR #1223, opened 2021-02-26
fix Cocina deposit creation for PURL reservation

jmartin-sul:
- `WorkType.purl_reservation_type.cocina_type` becomes a valid cocina type (`Cocina::Models::Vocab.object` instead of `'n/a'`)
- fix `reserving_purl` trait for `work_version` FactoryBot fixture to have the `reserving_purl` state (instead of copypasta `purl_reserved` state)
- unskip `deposit_job_spec.rb` test now that the above two changes allow it to pass

## Why was this change made?

This is part of the fix for #1191 (generate valid a valid Cocina model so that initial SDR resource gets created and a PURL is reserved).

The other part of fix for #1191 will be preventing editing/type selection for the work until the PURL is reserved.  That will prevent having to add a bunch of state transitions (as tried in #1202).  The simpler approach of preventing editing in the initial purl reservation states has the upside of keeping the state machine simpler (at the expense of briefly disallowing editing, though PURL reservation should finish quickly).

unit tests, tested PURL reservation using this branch on stage.

n/a

### DIFF
--- a/app/models/work_type.rb
+++ b/app/models/work_type.rb
@@ -92,7 +92,8 @@ class WorkType
 
   sig { returns(WorkType) }
   def self.purl_reservation_type
-    new(id: 'purl_reservation', label: 'PURL reservation', icon: '', subtypes: [], cocina_type: 'n/a')
+    new(id: 'purl_reservation', label: 'PURL reservation', icon: '', subtypes: [],
+        cocina_type: Cocina::Models::Vocab.object)
   end
 
   sig { params(id: T.nilable(String)).returns(WorkType) }
--- a/spec/factories/work_versions.rb
+++ b/spec/factories/work_versions.rb
@@ -125,6 +125,6 @@ FactoryBot.define do
     abstract { '' }
     citation { nil }
     license { 'none' }
-    state { 'purl_reserved' }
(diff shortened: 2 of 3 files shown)


jmartin-sul:
Per [this comment](https://github.com/sul-dlss/happy-heron/issues/1191#issuecomment-786360516), it seems like the `deposit_complete!` event isn't being fired as expected by the `DepositStatusJob` on the PURL reservation works.  I'm not sure if the issue is me re-starting the sneakers workers incorrectly (I attached to the existing screen session and ran this `rake` task from the [README](https://github.com/sul-dlss/happy-heron#rabbitmq-queue-workers), as well as the one above it).  Or if it's something else.

At any rate, this branch seems like an improvement over the current state of `main`, since it at least seems valid Cocina is now being generated.  Taking out of draft for review.

Manually calling `deposit_complete!` from rails console on the works stuck in the `reserving_purl` state gets the state to transition to the `purl_reserved` state.  And the app change in this PR is one line that doesn't mess with state definitions or worker code.  So I'm not sure how this change would cause issues for `DepositStatusJob`.

jcoyne (approved):

PR #1227, opened 2021-02-26
Use capistrano to manage sneakers

### LINKED ISSUE #1185: Run sneakers in systemd and capistrano
Currently we have to log in to the box after every deploy, kill the old screen session, start a new screen session and run:
```
RAILS_ENV=production WORKERS=AssignPidJob,DepositStatusJob bin/rake sneakers:run
```

It would be great if this could be handled by systemd and we could stop/start it with capistrano.

jcoyne:
## Why was this change made?

Fixes #1185

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,13 +128,25 @@ H2 is deployed via Capistrano to servers running the Passenger server in standal
 ### Setup RabbitMQ
 You must set up the durable rabbitmq queues that bind to the exchange where workflow messages are published.
 
-```
+```sh
 RAILS_ENV=production bin/rake rabbitmq:setup
 ```
 This is going to create queues for this application that bind to some topics.
 
 ### RabbitMQ queue workers
-RAILS_ENV=production WORKERS=AssignPidJob,DepositStatusJob bin/rake sneakers:run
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -128,13 +128,25 @@ H2 is deployed via Capistrano to servers running the Passenger server in standal
...
+In a development environment you can start sneakers this way:
+```sh
+WORKERS=AssignPidJob,DepositStatusJob bin/rake sneakers:run
+```
+
+but on the production machines we use systemd to do the same:
+```sh
+sudo /usr/bin/systemctl start sneakers
+sudo /usr/bin/systemctl stop sneakers
+sudo /usr/bin/systemctl status sneakers
```
mjgiarlo, line 146: Worth changing these lines to show how to invoke them via cap? Or adding that hereabouts?

jcoyne: I don't think we need to, we don't do that for any other cap invoked service.  `cap deploy` was what I was thinking of.

mjgiarlo: @jcoyne K.

mjgiarlo (approved):

PR #1236, opened 2021-02-27
slightly stricter updatability checks

### LINKED ISSUE #1191: Reserve a PURL function results in errors

### LINKED ISSUE #1200: [happy-heron/stage] StateMachines::InvalidTransition: Cannot transition state via :update_metadata from :reserving_purl (Reason(s): State cannot transition via "update metadata")
## Backtrace

    line 148 of [PROJECT_ROOT]/app/models/work_version.rb: block in choose_type_for_purl_reservation
    line 146 of [PROJECT_ROOT]/app/models/work_version.rb: choose_type_for_purl_reservation
    line 58 of [PROJECT_ROOT]/app/controllers/works_controller.rb: update_type

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/76952510)

jmartin-sul:
## Why was this change made?

closes #1191 (should prevent the state transition error i ran into)
closes #1200

* ~~`WorkVersion#updatable?` -- can't update something in the initial stages of PURL reservation, because the `update_metadata!` from `purl_reserved` brings the work version to draft.  If picking type and the first two PURL reservation states (`purl_requested`, `reserving_purl`) could happen in any order, it'd make the possible [states and transitions balloon](https://github.com/sul-dlss/happy-heron/pull/1202#issuecomment-786371113), to little benefit (since in most cases PURL reservation should happen quickly, and should block type selection and metadata update for only a few seconds).  See also https://github.com/sul-dlss/happy-heron/pull/1202 and https://github.com/sul-dlss/happy-heron/issues/1191~~ (as @jcoyne observed, this change was unnecessary, because the state defs already enforced it)
* consistency for `WorkVersionPolicy#update?` -- can't update something if it's not in an updatable state (even for admins/managers/reviewers -- no one's allowed to make illegal state transitions). `Collections::EditLinkComponent` and `Works::EditLinkComponent` already decline to render if the work or collection isn't updatable.
* `work_version_policy_spec.rb` -- a bit of additional testing, a bit of `context` grouping

unit tests, local dev instance (notice no edit link for `Reserving PURL`, but still get an edit link for the one that's reserved):
Screen Shot 2021-03-02 at 4 45 18 PM

n/a

### DIFF
--- a/app/policies/work_version_policy.rb
+++ b/app/policies/work_version_policy.rb
@@ -28,7 +28,7 @@ class WorkVersionPolicy < ApplicationPolicy
   end
 
   # Can edit a work iff:
-  #   The work is in a state where it can be updated (e.g. not depositing)
+  #   The work is in a state where it can be updated (e.g. not depositing, not an in-progress purl reservation)
   #   AND if any one of the following is true:
   #     1. The user is an administrator
   #     2. The user is the depositor of the work and it is not currently pending approval (review workflow)
@@ -36,11 +36,13 @@ class WorkVersionPolicy < ApplicationPolicy
   #     4. The user is a reviewer of the collection the work is in
   sig { returns(T::Boolean) }
   def update?
+    return false unless record.updatable?
+
     return true if administrator? ||
                    collection.managed_by.include?(user) ||
                    collection.reviewed_by.include?(user)
 
-    depositor? && record.updatable? && !record.pending_approval?
+    depositor? && !record.pending_approval?
   end
   # Can show a work iff any one of the following is true:
   #   1. The user is an administrator
--- a/spec/factories/work_versions.rb
+++ b/spec/factories/work_versions.rb
@@ -119,6 +119,15 @@ FactoryBot.define do
     state { 'purl_reserved' }
   end
 
+  trait :purl_requested do
+    work_type { WorkType.purl_reservation_type.id }
+    subtype { [] }
+    abstract { '' }
+    citation { nil }
+    license { 'none' }
+    state { 'purl_requested' }
+  end
+
   trait :reserving_purl do
     work_type { WorkType.purl_reservation_type.id }
     subtype { [] }
--- a/spec/policies/work_version_policy_spec.rb
+++ b/spec/policies/work_version_policy_spec.rb
@@ -63,6 +63,24 @@ RSpec.describe WorkVersionPolicy do
       let(:record) { build_stubbed :work_version, :depositing, work: work }
     end
 
+    failed 'when user is an admin and status is depositing' do
+      let(:groups) { [Settings.authorization_workgroup_names.administrators] }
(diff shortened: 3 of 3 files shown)


app/models/work_version.rb
```diff
@@ -89,6 +89,8 @@ class WorkTypeUpdateError < RuntimeError; end
   end
 
   def updatable?
+    return false if purl_requested? || reserving_purl?
```
jcoyne, line 92: Don't the existing cases already preclude one this method from returning true if the work is in on of those states? Can you write a test?

app/policies/work_version_policy.rb
```diff
@@ -28,19 +28,21 @@ def create?
...
-  #   The work is in a state where it can be updated (e.g. not depositing)
+  #   The work is in a state where it can be updated (e.g. not depositing, not an in-progress purl reservation)
   #   AND if any one of the following is true:
   #     1. The user is an administrator
   #     2. The user is the depositor of the work and it is not currently pending approval (review workflow)
   #     3. The user is a manager of the collection the work is in
   #     4. The user is a reviewer of the collection the work is in
   sig { returns(T::Boolean) }
   def update?
+    return false unless record.updatable?
```
jcoyne, line 39: This would break our existing requirement that admins can edit things that are in an in-review state.

jcoyne (changes requested):

app/models/work_version.rb
jmartin-sul: good point, will drop this guard clause

app/policies/work_version_policy.rb
jmartin-sul: for posterity:  discussed on slack w/ @jcoyne.  Determined that this would still allow admins to edit things that are in-review (since the state machine allows `update_metadata!` on `pending_approval` work versions).  This _does_ break the assumption that admins can edit anything at any time, by disallowing them from editing things in the `depositing` state, but I confirmed with @amyehodge that this was acceptable (like `purl_requested` and `reserving_purl`, `depositing` should be a pretty fleeting state when all is working normally).

jcoyne (approved):

PR #1246, opened 2021-03-03
use a default name for users if none is found

### LINKED ISSUE #1220: For new users, if name can't be found because they haven't logged in yet, substitute generic text
Currently, the email notifications require a user to have logged into the system for us to record the name to use in emails. For new users who are added to collections in any capacity, this log in will not typically have happened yet and the email salutation reads as "Dear ," with no address.

If the system does not find a name in the logs to use for the email salutation, please insert "New SDR User" so that it reads "Dear New SDR User,"

peetucket:
## Why was this change made?

Fixes #1220 - show a default name for users (this will work in both emails and anywhere else user names are used, which would otherwise result in blank showing)

Added a new test

### DIFF
--- a/app/mailers/collections_mailer.rb
+++ b/app/mailers/collections_mailer.rb
@@ -4,14 +4,14 @@
 # Sends email notifications about collections
 class CollectionsMailer < ApplicationMailer
   def invitation_to_deposit_email
-    @user = params[:user]
+    @user = UserPresenter.new(user: params[:user])
     @collection_version = params[:collection_version]
     @collection = @collection_version.collection
     mail(to: @user.email, subject: "Invitation to deposit to the #{@collection_version.name} collection in the SDR")
   end
 
   def deposit_access_removed_email
-    @user = params[:user]
+    @user = UserPresenter.new(user: params[:user])
     @collection_version = params[:collection_version]
     @collection = @collection_version.collection
     mail(to: @user.email, subject: "Your Depositor permissions for the #{@collection_version.name} " \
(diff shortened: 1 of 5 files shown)


app/models/user.rb
```diff
@@ -28,6 +28,11 @@ class User < ApplicationRecord
 
   devise :remote_user_authenticatable
 
+  sig { returns(String) }
+  def name
+    self[:name] || 'New SDR User' # return either the name in the database or a default if none exists (i.e. new users)
```
jcoyne, line 33: Can this be moved into the email, as that is the only place we're expecting this value?

jcoyne (changes requested):

app/models/user.rb
peetucket: Yeah, I considered this, but there are 4 mailers, some with many views, and this change would affect all of them.  It also seems that if we ever show the user's name in the top nav somewhere (as shown in the mockups) this would be desirable there too (@amyehodge ?).  Putting it on the model seemed a like nice clean way to handle it across the board.

Another way if we really do want to isolate to only the emails may be to put a `name` method in `ApplicationMailer` with similar logic and then use that across the mailers instead of using the attribute in the user model.

jcoyne: Another way is to make a `UserPresenter` (or `UserDecorator`) object that takes the `User` model as a parameter.  This can have:
```
def name
  model.name || 'New SDR User'
end
```
This means we don't need to have this logic in lots of different places and also keeps it out of the persistence layer.

jcoyne (approved):

jcoyne:
@peetucket this looks good, but needs a rebase.  Feel free to merge after rebasing if the tests pass.

peetucket:
Rebased, tests passing, merging.

PR #1251, opened 2021-03-03
Update resource type and genre mappings

peetucket:
## Why was this change made?

Replacing #1218 - updated resource and genre type mappings

Also
- remove genre duplicates which causes duped genres which came from both the top level type and the subtypes
- update tests to reflect new data

Some updated tests

### DIFF
--- a/app/services/types_generator.rb
+++ b/app/services/types_generator.rb
@@ -79,7 +79,7 @@ class TypesGenerator
     return [] if work_type == 'Other'
 
     # add the top level resource type mapping (i.e. top level work type with no subtype)
-    resource_types = [types_to_resource_types.dig(work_type, 'type')]
+    resource_types = Array(types_to_resource_types.dig(work_type, 'type'))
 
     # uniq and compact the list of resource types, since multiple subtypes can map
     # to the same resource type but we only need them mapped once
--- a/app/services/types_generator.rb
+++ b/app/services/types_generator.rb
@@ -64,7 +64,7 @@ class TypesGenerator
     # add the top level genre mapping (i.e. top level work type with no subtype)
     type_genres = types_to_genres.dig(work_type, 'type') || []
 
-    all_genres = type_genres + subtype_genres
+    all_genres = (type_genres + subtype_genres).uniq
     all_genres.map { |genre| Cocina::Models::DescriptiveValue.new(genre) }
   end
 
--- a/app/services/types_generator.rb
+++ b/app/services/types_generator.rb
@@ -64,6 +64,9 @@ class TypesGenerator
     # add the top level genre mapping (i.e. top level work type with no subtype)
     type_genres = types_to_genres.dig(work_type, 'type') || []
 
+    # NOTE: we should not add duplicate genres if the same one is coming from both the top level mapping and then the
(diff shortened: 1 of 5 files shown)


app/services/types_generator.rb
```diff
@@ -64,7 +64,7 @@ def build_genres
     # add the top level genre mapping (i.e. top level work type with no subtype)
     type_genres = types_to_genres.dig(work_type, 'type') || []
 
-    all_genres = type_genres + subtype_genres
+    all_genres = (type_genres + subtype_genres).uniq
```
peetucket, line 67: Per Arcadia in https://github.com/sul-dlss/happy-heron/issues/1254#issuecomment-790935330, we should not have duplicate genres if the same one is coming from both the top level mapping and then sub-type.  We will try and avoid duplicating in the .yml mappings, but this ensures it.

spec/services/types_generator_spec.rb
```diff
@@ -187,15 +234,15 @@
   end
 
   describe 'cocina mapping' do
-    non_other_work_types = WorkType.all.reject { |work_type| work_type.id == 'other' }
```
peetucket, line 190: we now have a mapping for 'Other', so it should be included in the tests

spec/services/types_generator_spec.rb
```diff
@@ -208,31 +255,6 @@
       end
     end
 
-    describe 'for subtypes' do
```
peetucket, line 211: this block of tests is no longer valid (we do not need to have a one-to-one mapping from work type to all of the subtypes as many will just come from the top level mapping, which is additive)

spec/services/types_generator_spec.rb
```diff
@@ -250,7 +272,12 @@
           .sort
           .uniq
       end
-      let(:known_genreless) { ['Policy brief', 'Speaker notes'] }
+      # these represent subtypes that will get the genre from the parent type
+      let(:known_genreless) do
+        ['Policy brief', 'Speaker notes', '3D model', 'Book chapter', 'Broadcast',
+         'Conference session', 'Other spoken word', 'Presentation recording',
+         'Presentation slides', 'Text']
+      end
```
peetucket, line 280: We do not have to have a genre mapping for all subtypes, as it will get the genre from the top level mapping.  We only need a subtype genre configured if it is additive to the main type.

spec/services/types_generator_spec.rb
```diff
@@ -187,15 +234,15 @@
   end
 
   describe 'cocina mapping' do
-    non_other_work_types = WorkType.all.reject { |work_type| work_type.id == 'other' }
+    work_types = WorkType.all
```
peetucket, line 237: we now have a mapping for 'Other', so it should be included in the tests

app/services/types_generator.rb
jcoyne: Would you mind adding that to a comment in the code here?

jcoyne: Would you mind adding that to a comment in the code here?

jcoyne (changes requested):

app/services/types_generator.rb
peetucket: Sure thing, added.

jcoyne (approved):

PR #1257, opened 2021-03-05
add new form for editing first draft of collections

### LINKED ISSUE #1250: User should be presented with the single collection edit form with all details and settings until the first version of the collection has completed deposit
If a user starts a collection and saves a draft and then returns to edit the details, as long as the title, description, and email validate, they can deposit the collection, even if they have not properly completed settings. @peetucket and I created multiple collections this way that then did not have the required license set, resulting in bad data that prevented us from creating items in the collection.

To be discussed at stand-up on Mar 4 to see if this approach will work.

Note: When done, please add back the "Save as Draft" button on first draft collections that was removed in #1253

peetucket:
## Why was this change made?

Fixes #1250 - create a new form for editing first drafts of collections that shows details and settings on the same page (just like when create a new collection).

Localhost browser
Existing tests
New tests


app/components/collections/create/buttons_component.html.erb
```diff
@@ -14,9 +14,7 @@
   </div>
   <div class="col-auto">
     <%= cancel_button %>
-    <%# The Save As Draft button is temporarily hidden on collection creation (first draft) ... see https://github.com/sul-dlss/happy-heron/issues/1250%>
-    <%# see also spec/components/collections/create/buttons_component_spec.rb which changes a test for this %>
-    <%#= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button', data: { action: 'unsaved-changes#allowFormSubmission' } %>
+    <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button', data: { action: 'unsaved-changes#allowFormSubmission' } %>
```
peetucket, line 17: add back the save as draft button to collection create form

app/components/collections/edit_link_component.rb
```diff
@@ -11,6 +11,15 @@ def initialize(collection_version:, anchor:, label:)
...
 
+    sig { returns(String) }
+    def link_path
+      if @collection_version.first_draft?
+        edit_first_draft_collection_path(collection, anchor: anchor)
+      else
+        edit_collection_path(collection, anchor: anchor)
+      end
+    end
+
```
peetucket, line 22: this approach to ensuring we link to the correct edit collection page is sprinkled in other parts too but is not comprehensive and prone to missing a spot --- i also added a redirect in the collections controller to send you to the new route if needed, and I think that might make more sense that trying to catch all of the links everywhere

spec/components/collections/create/buttons_component_spec.rb
```diff
@@ -19,14 +19,7 @@
...
-  # remove the following one when the button is added back
-  # see https://github.com/sul-dlss/happy-heron/issues/1250 and components/collections/create/button_compontent.html.erb
-  xit 'renders the save draft button' do
+  it 'renders the save draft button' do
     expect(rendered.css('input[value="Save as draft"]')).to be_present
   end
-
-  it 'temporary: it does not renders the save draft button' do
-    expect(rendered.css('input[value="Save as draft"]')).not_to be_present
-  end
```
peetucket, line 31: put the test back

app/policies/collection_policy.rb
```diff
@@ -16,6 +16,11 @@ def update?
     administrator? || manages_collection?(record)
   end
 
+  sig { returns(T::Boolean) }
+  def edit_first_draft?
+    update? && record.collection_versions.first.first_draft?
+  end
```
peetucket, line 22: we are only allowed to edit a first draft of a collection if its actually a first draft -- this ensures we don't end up on that page otherwise

app/components/dashboard/in_progress_collection_row_component.html.erb
```diff
@@ -3,7 +3,7 @@
     <%= show_collection_link %>
   </div>
   <div class="col-3">
-    <%= link_to 'Complete draft', edit_collection_path(collection), class: "mb-2 btn btn-primary" %>
+    <%= link_to 'Complete draft', edit_collection_version_path(collection.head), class: "mb-2 btn btn-primary" %>
```
peetucket, line 6: This links to the page that actually has the "Deposit" button (the details part of the from) and not the settings part of the form.  This should reduce user confusion when they click the "Complete draft" button from the dashboard.

spec/components/dashboard/in_progress_collection_component_spec.rb
```diff
@@ -28,9 +28,9 @@
 
   context 'when presenter has one or more in progress collections' do
     before do
-      create(:collection_version)
-      create(:collection_version)
-      create(:collection_version)
+      create(:collection_version_with_collection)
+      create(:collection_version_with_collection)
+      create(:collection_version_with_collection)
```
peetucket, line 33: The dashboard now has links to the collection_version edit page and so the tests needs a fully formed collection.

app/policies/collection_policy.rb
```diff
@@ -16,6 +17,11 @@ def update?
     administrator? || manages_collection?(record)
   end
 
+  sig { returns(T::Boolean) }
+  def update_first_draft?
+    update? && record.head.first_draft?
+  end
```
peetucket, line 23: we are only allowed to edit a first draft of a collection if its actually a first draft -- this ensures we don't end up on the wrong edit page

app/controllers/collection_versions_controller.rb
```diff
@@ -31,6 +31,9 @@ def edit
     collection_version = CollectionVersion.find(params[:id])
     authorize! collection_version
 
+    # if we end up on the edit page for a first draft (non-deposited collection), we need a separate form
```
jcoyne, line 34: I'm not sure I understand "we need a separate form". Can you clarify?

app/controllers/collection_versions_controller.rb
```diff
@@ -31,6 +31,9 @@ def edit
...
+    redirect_to edit_first_draft_collection_path(collection_version.collection.id) if collection_version.first_draft?
```
jcoyne, line 35:
```suggestion
    redirect_to edit_first_draft_collection_path(collection_version.collection) if collection_version.first_draft?
```

app/controllers/collections_controller.rb
```diff
@@ -41,19 +41,57 @@ def create
   # rubocop:enable Metrics/MethodLength
   # rubocop:enable Metrics/AbcSize
 
+  # this is a special route meant to edit a first draft (non deposited) version of a collection
+  # so that we can show the complete form instead of the split details/settings form
+  def edit_first_draft
```
jcoyne, line 46: Can you move these to a separate controller (e.g. FirstDraftCollectionController) or can we move the "Collection settings" related stuff to a different controller?  Having more than one form per controller seems more complicated than it ought to be

app/policies/collection_policy.rb
```diff
@@ -16,6 +17,11 @@ def update?
     administrator? || manages_collection?(record)
   end
 
+  sig { returns(T::Boolean) }
+  def update_first_draft?
```
jcoyne, line 21: I don't understand why we need a separate policy for this.  Isn't what `update?` is doing sufficient?

app/policies/collection_policy.rb
jcoyne: But we already have separate checks for that and redirect the user. So why do we need this?

config/routes.rb
```diff
@@ -18,6 +18,8 @@
 
   resources :collections, only: %i[new create show edit update destroy] do
     member do
+      get :edit_first_draft
```
jcoyne, line 21: Please make these two methods on their own controller.  We are saying "a first draft collection is so different that it ought to be treated as its own kind of resource".

app/components/dashboard/in_progress_collection_row_component.html.erb
jcoyne: This PR is really complicated, so can you pull this work out into it's own PR?

jcoyne (changes requested):

app/components/dashboard/in_progress_collection_row_component.html.erb
peetucket: ok, moved to #1267

app/controllers/collection_versions_controller.rb
peetucket: basically a form with all of the fields instead of being split into details and settings, like we have for a new collection... i can rephrase the comment

app/policies/collection_policy.rb
peetucket: this policy ensures you don't end up on the page that is only designed for editing first drafts if you aren't actually editing a first draft -- there aren't any links that will take you there, but you could craft one in the browser -- this just guards against that, so its not critical

app/controllers/collections_controller.rb
peetucket: i can create a new controller for these two new actions like you suggest

app/policies/collection_policy.rb
peetucket: the redirects go the other way (ensuring you go to the first draft actions when needed), not ensuring you don't go to the first draft action when you shouldn't ... but that could be handled in the same way with a redirect for consistency

config/routes.rb
peetucket: Done

app/policies/collection_policy.rb
peetucket: I just removed the policy changes -- the collection policy and redirects should be sufficient.

app/policies/collection_policy.rb
peetucket: Used redirects in the new controller and removed the policy changes.

app/controllers/first_draft_collections_controller.rb
```diff
@@ -0,0 +1,67 @@
...
+# frozen_string_literal: true
+
+# Handles updates for first drafts of collections.
+# When updating a first draft of a collection, you need to show a form that includes both
+# the settings and the details all in one page.  For later versions, the form is split in two.
+class FirstDraftCollectionsController < ObjectsController
+  before_action :authenticate_user!
+  before_action :ensure_sdr_updatable
+
+  def edit
```
jcoyne, line 11: Should we move `new` and `create` here as well since they're also working on the `CreateCollectionForm`?

jcoyne (changes requested):

app/controllers/first_draft_collections_controller.rb
peetucket: Yeah, I thought about that too.  Let me give it a try and see how many changes it actually invokes.

peetucket:
Ok, added a commit that moves the collection creation to the new first draft controller, which also involved moving components and some other pieces

peetucket:
Note: I don't think this drop in test coverage is correct.  Code got moved around but is still exercised in the tests, none of which were removed.

jcoyne (approved):

Issue #1248, opened 2021-03-03
Reserve a PURL


amyehodge:
This is still not working.

jmartin-sul:
I've done some testing and am a bit stumped at the moment.  I wonder if there's some flakiness with communication between  SDR (specifically... dor-services-app?  sdr-api?) and H2?  Or some flakiness in H2 job execution?

I tried to reserve a PURL at 10:28 am pacific today, and that seemed to get stuck in the "Reserving PURL" state (still stuck as of this comment, see screenshot).  But subsequent attempts to reserve a PURL all seemed to succeed.

Because of the timing (10:28 am), I suspected this sdr-api error reported by Honeybadger, but I think it's a red herring:  https://app.honeybadger.io/projects/67994/faults/64639563/01EZWNHS132M1NPG4NRWVQD9G1?page=0#notice-params

```
Dor::Services::Client::UnexpectedResponse: Bad Request: 400 ({"errors":[{"status":"400","title":"Bad Request","detail":"Unable to find collection: 'Unable to find 'druid:mp757fq0674' in fedora. See logger for details.'"}]})
```

I suspected maybe a race condition, where the collection I'd just created couldn't be referenced yet, but the collection in the error message is `mp757fq0674`, and the collection for the stuck purl res is `kr811mj1951`.  The couple of errors I see in dor-indexing-app from around the same time (10:40 am) seem to be about a data error (attempted reindexing hitting an object without a title).

The other thing I noticed was that a PURL reservation I successfully completed got stuck on "Deposit in progress" after I edited metadata and tried to deposit.  I reserved that PURL as fast as I could in a newly created collection, trying to reproduce the stuck reservation by creating the race condition I'd initially thought might be the culprit, but I was able to choose a type and edit metadata without a problem.

The last thing I noticed is that the PURL link worked for all my test reservations, _except_ for the stuck reservation.  I can visit Argo for all of the druids.  The stuck one only has `registrationWF`, whereas the rest have that plus `preservationIngestWF` and `accessionWF`.
<img width="1411" alt="Screen Shot 2021-03-03 at 3 47 44 PM" src="https://user-images.githubusercontent.com/7741604/109894688-ca616b80-7c42-11eb-985f-5a45c02ebe8b.png">

No other obvious errors jumped out at me from Honeybadger.

After typing all that out, the only somewhat concrete hunch I have is something something stale worker code?  Since there was a fix to the accessioning flag in sdr-client? 🤷

[...]

PR #1261, opened 2021-03-06
Fix PURL reservation state transition handling

jmartin-sul:
## Why was this change made?

an attempt at progress on #1248 

* unit tests
* testing on stage

n/a


jmartin-sul:
this doesn't seem to fully fix things, but it does seem like an improvement based on my testing of this on stage this past friday.

- Paired with @jcoyne Friday morning.  Determined that work versions in `reserving_purl` were no longer getting moved to `purl_reserved` because that transition was on the `deposit_complete!` event (prior to this PR).  There was no longer a deposit to complete (and so no call from RabbitMQ to H2's `DepositStatusJob`, which calls `deposit_complete!`).  The lack of deposit was because of the switch to `accessioning`=false in the initial SDR object creation when reserving a PURL (see the fri 2/26 fix to sdr-client, and the first PURL res PR https://github.com/sul-dlss/sdr-client/pull/148 and https://github.com/sul-dlss/happy-heron/pull/989/files#diff-ced6e105bac360d3d0e85f61aa63495a21ba9c6018742345a388f6b61a1317f7R37)
  - This PR fixes H2 to have a specific `pid_assigned!` event that does the `reserving_purl` to `purl_reserved` state transition, called from `AssignPidJob` in H2, which got things unstuck from "Reserving PURL" (based on my testing on -stage on Friday).
- With a PURL that reserved successfully, I was able to pick a type for it and edit metadata.  But then it gets stuck on `depositing`.  see https://purl-stage.stanford.edu/mc527yk6759
  - But sometimes it still seems the purl reservation sets the `accesion` flag to true for SDR object creation, or at least that's what it looks like from the DSA logs, and `accessionWF` exists on some objects if you look at them in Argo.  e.g. https://purl-stage.stanford.edu/wh665zf3298
  - I didn't see any stale sneakers workers when I looked at the `ps` listing on H2 stage -- there were 4 processes, and all were started after my last deployment
  - Regular deposit still seems to work fine, thankfully, see https://purl-stage.stanford.edu/th908kt1622
  - I paired with @peetucket Friday afternoon, we couldn't figure out where things were getting dropped (re: PURL reservation works getting stuck on "Depositing" when trying to pick a type and flesh out metadata).
- No obviously relevant honeybadger alerts, though it is weird that `AssignPidJob` got called on states that already had a druid (or maybe it's not, but it does seem that `sdr.objects.created` event gets broadcast from DSA more than just for the druid assignment -- e.g. https://app.honeybadger.io/projects/77112/faults/77571207).
- I did see an unexpected error where `deposit_complete!` gets called on `purl_reserved` work, see https://app.honeybadger.io/projects/77112/faults/77580929

app/models/work_version.rb
```diff
@@ -64,7 +64,11 @@ class WorkTypeUpdateError < RuntimeError; end
 
     event :deposit_complete do
       transition depositing: :deposited
+    end
+
+    event :pid_assigned do
       transition reserving_purl: :purl_reserved
+      transition %i[first_draft depositing pending_approval purl_reserved] => same
```
jmartin-sul, line 71: Originally, I had expected only needing to include `first_draft` and `depositing`, but I guess there's an `sdr.objects.created`/`h2.druid_assigned` message in other situations too?  I added `pending_approval` and `purl_reserved` based (respectively iirc) on running unit tests and doing manual testing on stage.

spec/features/purl_reservation_spec.rb
```diff
@@ -32,10 +32,14 @@
...
-      work_version.add_purl_to_citation
-      work_version.deposit_complete!
+      # IRL, dor-services-app would send a message that RabbitMQ would route to h2.druid_assigned,
+      # for processing by AssignPidJob, so we'll just fake that by running the job manually
+      source_id = "hydrus:object-#{work_version.work.id}"
+      identification = instance_double(Cocina::Models::Identification, sourceId: source_id)
+      model = instance_double(Cocina::Models::DRO, identification: identification, externalIdentifier: druid)
+      assign_pid_job = AssignPidJob.new
+      allow(assign_pid_job).to receive(:build_cocina_model_from_json_str).and_return(model)
+      assign_pid_job.work('{}') # don't need to fake JSON for fully valid Cocina model, just mock resulting DRO
```
jmartin-sul, line 42: i initially started out by faking the json and calling `AssignPidJob` in a more realistic manner, but it seemed like that'd grow cumbersome, and might lead to brittle tests.  happy to revisit if that does in fact seem like a better approach than the mocking done here.

jmartin-sul:
this PR does not fully get PURL reservation working, but it does seem like it improves things without introducing regression, so i think this is ready for review (and possibly merging if it's not on the wrong track)

jmartin-sul:
hmm, might've figured out what this is missing, going back to draft for more tweaking/testing...

jmartin-sul:
> hmm, might've figured out what this is missing, going back to draft for more tweaking/testing...

my hunch didn't pan out.  marking ready for review again...

app/models/work_version.rb
jcoyne: I'm a bit confused as to how first_draft. pending_approval or purl_reserved would ever make this transition.  Can you explain more?

jcoyne (changes requested):

app/models/work_version.rb
jmartin-sul: iirc, i added all three because either the unit tests tried to make that that transition, or because the app tried to make that transition when i tested on -stage on friday.  will try to reproduce and give a bit more detail as to what triggered this transition on those states...

jmartin-sul: (i was also surprised, esp on the `purl_reserved` one)

jcoyne: I think the unit test should be adjusted so that they are in the correct state rather than change the code to try and accomodate objects that aren't correct.

jmartin-sul: fixed the test issue, but deploying to -qa to see if i can reproduce the transition issue i stumbled on when testing on -stage...

jmartin-sul: full disclosure: i tested just now with this commit:  0329901ce6cb9ea3b71e3883da03fcd8ef3b4695

but that just adds some debug logging to the changes in this branch, in an attempt to figure out why a PURL res isn't turned into a real deposit after a user selects a type (and edits metadata and hits the "Deposit" button).

https://github.com/sul-dlss/happy-heron/compare/b5d6de56c686c8e8fea3f834a6fdd2d86cab5b30...0329901ce6cb9ea3b71e3883da03fcd8ef3b4695

jmartin-sul: paired with @jcoyne this morning, and after walking through the code, we realized that `pid_assigned!` was getting called twice on a `WorkVersion` in `purl_reserved`, because `DepositJob` was being run twice for PURL reservations (because both the `reserve_purl` and `begin_deposit` events were being invoked for PURL reservation, and both of those events have an `after_transition` hook that invokes `WorkObserver#after_begin_deposit`).

this should be fixed in the third commit on this PR, and now `depositing` and `reserving_purl` should be the only "from" states for the `pid_assigned` event/transition.

jmartin-sul:
@jcoyne ready for review yet again 🙂 

based on my testing on QA, this does not yet fix the issue with not being able to turn a PURL res into a full fledged deposit.  but hopefully this is forward progress, and the third commit should address the state transition concern raised earlier in this PR's review.

jcoyne (approved):

PR #1417, opened 2021-05-11
Upgrade to webpack5

jcoyne:
## Why was this change made?
This avoids the warning error we're seeing about "loose" configs.
See https://github.com/rails/webpacker/issues/3008

Also avoids CVE-2021-23368


Gemfile
```diff
@@ -73,6 +73,6 @@ gem 'sorbet-runtime'
 gem 'state_machines-activerecord'
 gem 'turbo-rails', '~> 0.5.9'
 gem 'view_component', '~> 2.18'
-gem 'webpacker', '~> 5.0'
+gem 'webpacker', '6.0.0.beta7'
```
jermnelson, line 76: Is the intention to pin webpacker until the version `6.0.0` is no longer in beta?

jermnelson (approved):
Approved with an informational question.

Gemfile
jcoyne: Yes, that's correct.

PR #1423, opened 2021-05-12
Update the keyword entry to be a single text box with "add another" button

### LINKED ISSUE #1377: Change keywords to single entry boxes similar to other elements in the UI
After discussing with @astridu and @hannahfrost, we think changing this entry field so that the user enters one keyword at a time and then clicks on an "add another keyword" box to add another will be much clearer to the users and will be consistent with the rest of the UI where multiple entries can be made. It also eliminates an inconsistency in the UI where this field required use of the enter key, while hitting that key anywhere else on the page results in saving the page and returning to the view page.

@jcoyne and @mjgiarlo, how do you think this will work if we want to implement the FAST vocabulary lookup in the future? We prefer not to do something that will make that harder later. Also, let me know if you see any complications for the validations with this method. Thanks.

aaron-collier:
## Why was this change made?

Fixes #1377

1. Single entry:

Screen Shot 2021-05-13 at 10 35 45 AM

2. When another keyword is added:

Screen Shot 2021-05-13 at 10 35 55 AM

3. Error message if blank:

Screen Shot 2021-05-13 at 10 36 11 AM

4. Entering multiple values:

Screen Shot 2021-05-13 at 10 36 34 AM

5. Showing the values saves as draft:

Screen Shot 2021-05-13 at 10 36 42 AM

Updated feature tests

### DIFF
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -1,4 +1,4 @@
-<section id="description" data-controller="nested-form" data-nested-form-select-value=".inner-container">
+<section id="description">
   <header>Describe your deposit</header>
   <p>Enter a summary statement about the deposit (600 words max) to help others
     discover your deposits in SearchWorks and on the internet. Add at least one
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -21,7 +21,7 @@
       <%= render PopoverComponent.new key: 'work.keywords' %>
     </div>
     <div class="col-sm-10">
-      <%= render Works::KeywordsComponent.new(form: form) %>
+      <%= render Works::KeywordsComponent.new(form: form, key: 'work.keywords') %>
     </div>
   </div>
 
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -21,7 +21,7 @@
       <%= render PopoverComponent.new key: 'work.keywords' %>
     </div>
     <div class="col-sm-10">
-      <%= render Works::KeywordsComponent.new(form: form, key: 'work.keywords') %>
+      <%= render Works::KeywordsComponent.new(form: form) %>
     </div>
   </div>
 
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -1,4 +1,4 @@
-<section id="description">
+<section id="description" data-controller="nested-form" data-nested-form-select-value=".inner-container">
   <header>Describe your deposit</header>
   <p>Enter a summary statement about the deposit (600 words max) to help others
     discover your deposits in SearchWorks and on the internet. Add at least one
--- a/app/components/works/keywords_row_component.rb
+++ b/app/components/works/keywords_row_component.rb
@@ -4,9 +4,8 @@
 module Works
   # Allows the user to search for keywords or provide freetext keywords
   class KeywordsRowComponent < ApplicationComponent
-    def initialize(form:, key:)
+    def initialize(form:)
       @form = form
-      @key = key
     end
 
     attr_reader :form, :key
--- a/app/components/works/keywords_row_component.rb
+++ b/app/components/works/keywords_row_component.rb
@@ -3,7 +3,7 @@
 
 module Works
   # Allows the user to search for keywords or provide freetext keywords
-  class KeywordsComponent < ApplicationComponent
+  class KeywordsRowComponent < ApplicationComponent
     def initialize(form:, key:)
       @form = form
       @key = key
@@ -21,10 +21,6 @@ module Works
 
     def errors
       form.object.errors.where(:keywords)
-    end
-
-    def row(keyword_form)
-      render Works::KeywordsRowComponent.new(form: keyword_form, key: key)
     end
   end
 end
--- a/app/components/works/keywords_row_component.rb
+++ b/app/components/works/keywords_row_component.rb
@@ -8,7 +8,7 @@ module Works
       @form = form
     end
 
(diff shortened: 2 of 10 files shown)


aaron-collier:
@jcoyne I'm a bit stumped on why the keywords aren't saving. I think it must be something around stimulus data-targets as I see this when running tests:

```
%o Error invoking action "change->keywords#add" Error: Missing target element "keywords.addItem"
    at extended.get (http://127.0.0.1:51254/packs-test/js/application-4df07eaf82da9e3da629.js:20313:15)
    at extended.add (http://127.0.0.1:51254/packs-test/js/application-4df07eaf82da9e3da629.js:1493:12)
    at Binding.invokeWithEvent (http://127.0.0.1:51254/packs-test/js/application-4df07eaf82da9e3da629.js:18689:19)
    at Binding.handleEvent (http://127.0.0.1:51254/packs-test/js/application-4df07eaf82da9e3da629.js:18662:12)
    at EventListener.handleEvent (http://127.0.0.1:51254/packs-test/js/application-4df07eaf82da9e3da629.js:19632:17) Object
      deposits and renders work show page (FAILED - 2)
  HTML screenshot: /Users/amcollie/github/sul-dlss/happy-heron/tmp/capybara/screenshot_2021-05-12-15-22-52.709.html
  Image screenshot: /Users/amcollie/github/sul-dlss/happy-heron/tmp/capybara/screenshot_2021-05-12-15-22-52.709.png
```

I know this may be blocking work, and would like to get it wrapped up. Can I ping you in the morning for pairing? Or if you see anything obvious, feel free to comment/suggest changes. Thanks.

app/components/works/keywords_row_component.rb
```diff
@@ -0,0 +1,27 @@
+# typed: false
+# frozen_string_literal: true
+
+module Works
+  # Allows the user to search for keywords or provide freetext keywords
+  class KeywordsRowComponent < ApplicationComponent
+    def initialize(form:, key:)
+      @form = form
+      @key = key
```
jcoyne, line 9: Is this parameter necessary?

jcoyne (changes requested):

app/components/works/keywords_row_component.rb
aaron-collier: No - doesn't seem to be. I've dropped the key parameters in this PR.

aaron-collier:
@jcoyne I suspect the coverage drop here is from the small bits of removed code.

app/components/works/keywords_row_component.rb
```diff
@@ -0,0 +1,26 @@
...
+# frozen_string_literal: true
+
+module Works
+  # Allows the user to search for keywords or provide freetext keywords
+  class KeywordsRowComponent < ApplicationComponent
+    def initialize(form:)
+      @form = form
+    end
+
+    attr_reader :form, :key
```
jcoyne, line 11: Remove `:key`?

app/components/works/description_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<section id="description">
+<section id="description" data-controller="nested-form" data-nested-form-select-value=".inner-container">
```
jcoyne, line 1: I'm pretty sure this `data-nested-form-select-value=".inner-container"` isn't doing anything and can be removed.

jcoyne: Why are you changing this component at all?

spec/components/works/keywords_row_component_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  let(:keyword_form) do
+    work_form.prepopulate!
+    work_form.keywords.first
+  end
+  let(:form_builder) do
+    ActionView::Helpers::FormBuilder.new('work', keyword_form, controller.view_context, {})
+  end
+
+  it 'renders the component' do
+    expect(rendered.to_html).to include('Keyword')
```
jcoyne, line 21: How about a test for `.plain-container` since that is required for the script to work?

spec/components/works/keywords_component_spec.rb
```diff
@@ -11,8 +11,6 @@
   let(:rendered) { render_inline(described_class.new(form: form)) }
 
   it 'renders the component' do
-    expect(rendered.css('.keywords').to_html)
-      .to be_present
-    expect(rendered.css('.keywords-container.is-invalid')).not_to be_present
+    expect(rendered.to_html).to include('Keyword')
```
jcoyne, line 14: You can add a test for the "Add another keyword" button here.

jcoyne (changes requested):

app/components/works/keywords_row_component.rb
aaron-collier: Cruft, will remove.

app/components/works/description_component.html.erb
aaron-collier: Cruft. I think this was when I was trying to figure out why the changes weren't taking effect. Will remove.

spec/components/works/keywords_row_component_spec.rb
aaron-collier: Happy to add this, could you provide a short example of what you are thinking? I'm not seeing this tested on contact_email (following that pattern, but I can add it there too if we wish)

jcoyne: `expect(rendered.css('.plain-container')).to be_present`

aaron-collier: Perfect, thanks. Done.

jcoyne (approved):

PR #1424, opened 2021-05-12
Removes expired embargo dates from Work Form

### LINKED ISSUE #1352: Allow users to version objects released from embargo
I attempted to version https://sul-h2-stage.stanford.edu/works/13246 as discussed with @jcoyne and @mjgiarlo earlier today in sprint planning. This item was embargoed and then released the next day on schedule. Three versions are recorded in Argo but only two in H2.

I opened this items to version it -- added a version comment and added a new author, then clicked "Deposit." The page seemed to reload without my changes saved. I reviewed the page and made the same changes as above. I noted that the email field was empty -- I don't know if it was initially or not. I added an email with the other info and tried again to deposit and I got the same sort of page reload and my changes were again not saved -- no version comment, no new author, and this time no email address. When I check on the status, it is showing as deposited instead of being in a draft state.

Also, every time I try to version it I receive the email that there is new activity on this item.

jermnelson:
## Why was this change made?
Maybe part of the solution for #1352 

New unit test

n/a

### DIFF
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -29,7 +29,12 @@ class DraftWorkForm < Reform::Form
   property :release, virtual: true, prepopulator: (proc do |*|
     self.release = embargo_date.present? ? 'embargo' : 'immediate'
   end)
-  property :embargo_date, embargo_date: true, on: :work_version
+  property :embargo_date, embargo_date: true, on: :work_version, prepopulator: (proc do |*|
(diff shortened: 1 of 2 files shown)


jcoyne (approved):
This seems like a good change, however I'm not convinced this is the complete fix for #1352.  Would you mind changing the description to just referencing #1352 (not fixing) and then merge and deploy to stage and test before we call it fixed?  I'm concerned we also have to deal with version mismatches as dor-services-app will have incremented a version to remove the embargo.

PR #1444, opened 2021-05-13
Update required field labels and remove optional text for works

### LINKED ISSUE #1391: Add asterisks for required fields on work edit form
See wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420954309_v_Item_1_Edit

This ticket is in response to user testing where several users had trouble understanding which fields were required and which weren't.

Changes include:
1. Update text under work title on pages to say "*REQUIRED FIELDS"
2. Adding in asterisks next to section headers and field labels as indicated in the wireframes.
3. Removal of "(optional)" text

aaron-collier:
## Why was this change made?

Fixes #1391 

Screen Shot 2021-05-13 at 3 48 14 PM

- Adds '*' to required fields
- Update text to "* REQUIRED FIELDS*
- Removes (optional) where unnecessary.

Updated tests when looking for text

### DIFF
--- a/app/components/works/contributor_row_component.rb
+++ b/app/components/works/contributor_row_component.rb
@@ -4,11 +4,17 @@
 module Works
   # Renders a widget corresponding to a single contributor to the work.
   class ContributorRowComponent < ApplicationComponent
-    sig { params(form: ActionView::Helpers::FormBuilder, required: T::Boolean, citation: T::Boolean).void }
-    def initialize(form:, required: false, citation: false)
+    sig do
+      params(form: ActionView::Helpers::FormBuilder,
+             required: T::Boolean,
+             citation: T::Boolean,
+             optional: T::Boolean).void
+    end
+    def initialize(form:, required: false, citation: false, optional: false)
       @form = form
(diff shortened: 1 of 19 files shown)


app/components/works/contributor_row_component.rb
```diff
@@ -4,11 +4,17 @@
...
+      params(form: ActionView::Helpers::FormBuilder,
+             required: T::Boolean,
+             citation: T::Boolean,
+             optional: T::Boolean).void
+    end
+    def initialize(form:, required: false, citation: false, optional: false)
       @form = form
       @required = required
       @citation = citation
+      @optional = optional
```
jcoyne, line 17: Isn't optional implied when required is false?

jcoyne (changes requested):

app/components/works/contributor_row_component.rb
aaron-collier: Doh, yes. I totally missed that. Will update.

jcoyne (approved):

PR #1457, opened 2021-05-14
Add draft label to collection name when needed on dashboard

### LINKED ISSUE #1049: When a collection is in draft state, include "Draft" after the collection title in the admin table on the dashboard

aaron-collier:
## Why was this change made?

Fixes #1049 

Screen Shot 2021-05-14 at 3 59 03 PM

Note: This is draft pending updated test(s).

### DIFF
--- a/app/components/dashboard/all_collections_row_component.html.erb
+++ b/app/components/dashboard/all_collections_row_component.html.erb
@@ -1,5 +1,5 @@
 <tr>
-  <td><%= link_to collection.head.name, collection %><strong><i><%= draft_label %></i></strong></td>
+  <td><%= link_to collection.head.name, collection %><% if draft? %><strong><i> <%= draft_label %></i></strong><% end %></td>
   <td><%= counts['total'] %></td>
   <td><%= counts['deposited'] %></td>
   <td><%= counts['first_draft'] %></td>
--- a/app/components/dashboard/all_collections_row_component.html.erb
+++ b/app/components/dashboard/all_collections_row_component.html.erb
@@ -1,0 +1,10 @@
+<tr>
+  <td><%= link_to collection.head.name, collection %><strong><i><%= draft_label %></i></strong></td>
+  <td><%= counts['total'] %></td>
+  <td><%= counts['deposited'] %></td>
+  <td><%= counts['first_draft'] %></td>
+  <td><%= counts['version_draft'] %></td>
+  <td><%= counts['pending_approval'] %></td>
+  <td><%= counts['rejected'] %></td>
+  <td><%= I18n.l(collection.updated_at.to_date, format: :abbr_month) %></td>
+</tr>
--- a/app/components/dashboard/all_collections_row_component.html.erb
+++ b/app/components/dashboard/all_collections_row_component.html.erb
@@ -1,5 +1,5 @@
 <tr>
-  <td><%= link_to collection.head.name, collection %><% if draft? %><strong><i> <%= draft_label %></i></strong><% end %></td>
+  <td><%= link_to collection.head.name, collection %><%= draft_label %></td>
   <td><%= counts['total'] %></td>
   <td><%= counts['deposited'] %></td>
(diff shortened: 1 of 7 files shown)


spec/requests/dashboards_spec.rb
```diff
@@ -174,28 +174,24 @@
...
-        <td><a href=\"#{collection_path(workful_collection)}\">MyString</a></td>
-        <td>5</td>
-        <td>1</td>
-        <td>1</td>
-        <td>1</td>
-        <td>1</td>
-        <td>1</td>
-        <td>Dec 02, 2020</td>
-      </tr>
+  <td><a href=\"#{collection_path(workful_collection)}\">MyString</a><strong><i></i></strong></td>
```
aaron-collier, line 177: These two tests wouldn't pass without spacing these over after moving them into a component. Same below.

app/components/dashboard/all_collections_row_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<tr>
+  <td><%= link_to collection.head.name, collection %><strong><i><%= draft_label %></i></strong></td>
```
jcoyne, line 2: Should there be a space before the draft_label?

jcoyne (changes requested):

spec/requests/dashboards_spec.rb
jcoyne: I feel like checking for exact html in a request spec is a bit of an anti-pattern, but I don't think we need to fix it here as it was a pre-existing condition.

app/components/dashboard/all_collections_row_component.html.erb
jcoyne: Do we want to have `<strong><i></i></strong>` on the page even when there is no draft label?

jcoyne (changes requested):

app/components/dashboard/all_collections_row_component.html.erb
```diff
@@ -0,0 +1,10 @@
+<tr>
+  <td><%= link_to collection.head.name, collection %><% if draft? %><strong><i> <%= draft_label %></i></strong><% end %></td>
```
jcoyne, line 2: How about putting all this HTML and the conditional into a method in the component?  so just `<= draft_label %>` here and then `draft_label` can `return unless draft?`.

app/components/dashboard/all_collections_row_component.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    attr_reader :collection, :counts
+
+    def draft?
+      return true if @collection.head&.state&.include? 'draft'
+
+      false
+    end
+
+    def draft_label
```
jcoyne, line 21: why not return all the html from this method?

```ruby
  tag.strong do
   tag.i  ' - Draft'
  end
 ```

File: app/components/dashboard/all_collections_row_component.rb [hunk 5]
```diff
@@ -0,0 +1,32 @@
...
+      case @collection.head&.state
+      when 'first_draft'
+        return '- Draft'
```
jcoyne, line 24: These returns aren't necessary if this is the last statement in the method. (e.g. get rid of the `nil`)

jcoyne (changes requested):

app/components/dashboard/all_collections_row_component.rb
jcoyne: better sill might be:
```ruby
tag.span ' - Draft', class: 'draft-tag'
```

and then add a `.draft-tag` css style.

aaron-collier: This is what I was looking for! Thanks, I also wasn't happy with the HTML that way.

jcoyne (approved):

jcoyne:
@aaron-collier looks like you need to clean up the typecheck (just set it to false).

PR #1462, opened 2021-05-17
Extract more-types controller from edit-deposit controller

jcoyne:
## Why was this change made?
This keeps the function of each controller clear.

### DIFF
--- a/app/components/works/available_date_component.html.erb
+++ b/app/components/works/available_date_component.html.erb
@@ -6,7 +6,7 @@
     </div>
   </div>
 
-  <div class="embargo-date mb-5" data-edit-deposit-target="embargo-dateField">
+  <div class="embargo-date mb-5">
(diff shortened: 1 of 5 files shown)


app/components/works/available_date_component.html.erb
```diff
@@ -6,7 +6,7 @@
     </div>
   </div>
 
-  <div class="embargo-date mb-5" data-edit-deposit-target="embargo-dateField">
```
jcoyne, line 9: This was cruft

app/components/works/contributor_row_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="row inner-container" data-controller="contributors" data-contributors-required-value="<%= required? %>" data-edit-deposit-target="contributorsField">
```
jcoyne, line 1: This was cruft

jermnelson (approved):

PR #1468, opened 2021-05-17
Depositors should only see their works in the collection table

### LINKED ISSUE #1274: Depositors should only be able to see their items in the collections tables on the dashboard.
The collection tables on the dashboard display the 4 most recent items in the collection. They appear to do this for depositors regardless of who made the items, and then allow users to click in to view those items. Users should only be able to see their items in this list on the dashboard.

Note that the full list of deposits appears to correctly show only the items for the depositor and no other items.

jcoyne:
## Why was this change made?

Fixes #1274

### DIFF
--- a/app/components/dashboard/collection_component.html.erb
+++ b/app/components/dashboard/collection_component.html.erb
@@ -1,38 +1,30 @@
-<section>
-  <table class="table">
-    <caption>
-      <%= render Dashboard::CollectionHeaderComponent.new collection_version: collection.head %>
-    </caption>
-    <thead class="table-light">
-      <tr>
-        <th>Deposits in collection</th>
-        <th><span class="visually-hidden">Actions</span></th>
-        <th>Deposit status</th>
-        <th>Last modified</th>
-        <th>Persistent link</th>
-        <th>Citation</th>
-      </tr>
-    </thead>
-    <tbody>
-      <% visible_deposits.take(MAX_DEPOSITS_TO_SHOW).each do |work| %>
+<li class="mb-5">
+  <section>
+    <table class="table">
+      <caption>
+        <%= render Dashboard::CollectionHeaderComponent.new collection_version: collection.head %>
+      </caption>
+      <thead class="table-light">
         <tr>
-          <%= render Works::LinkToShowComponent.new(work_version: work.head) %>
-          <td>
-            <%= helpers.turbo_frame_tag dom_id(work, :edit), src: edit_button_work_path(work), target: '_top' %>
-            <%= helpers.turbo_frame_tag dom_id(work, :delete), src: delete_button_work_path(work) %>
-          </td>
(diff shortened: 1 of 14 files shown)


app/components/dashboard/collection_component.html.erb
```diff
@@ -1,38 +1,30 @@
...
-        <th><span class="visually-hidden">Actions</span></th>
-        <th>Deposit status</th>
-        <th>Last modified</th>
-        <th>Persistent link</th>
-        <th>Citation</th>
-      </tr>
-    </thead>
-    <tbody>
-      <% visible_deposits.take(MAX_DEPOSITS_TO_SHOW).each do |work| %>
+<li class="mb-5">
```
jcoyne, line 1: Added this node and indented everything in this file.

app/components/dashboard/collection_component.html.erb
```diff
@@ -1,38 +1,30 @@
...
+          <th>Deposit status</th>
+          <th>Last modified</th>
+          <th>Persistent link</th>
+          <th>Citation</th>
         </tr>
-      <% end %>
-    </tbody>
-  </table>
+      </thead>
+      <tbody>
```
jcoyne, line 17: code changes begin here.

app/components/dashboard/collection_component.rb
```diff
@@ -1,12 +1,12 @@
...
+# typed: false
 # frozen_string_literal: true
 
 module Dashboard
   # Renders a collection and a summary table of works in the collection
   class CollectionComponent < ApplicationComponent
     MAX_DEPOSITS_TO_SHOW = 4
 
-    sig { params(collection: Collection).void }
+    # sig { params(collection: Collection).void }
```
jcoyne, line 9: This is incompatible with the check that view_component does to see if the initializer accepts a kwarg `collection`

app/components/dashboard/collection_component.rb
```diff
@@ -24,7 +24,8 @@ def visible_deposits
       # displayable works and add another query to get the overall count, but
       # that would double the number of queries run for each instance of this
       # component.
-      collection.works.order('updated_at desc').limit(MAX_DEPOSITS_TO_SHOW + 1)
+      scope = policy.apply_scope collection.works, type: :relation
```
jcoyne, line 27: This is the important change, using the policy to scope the results.

app/models/work_version.rb
```diff
@@ -36,7 +36,6 @@ class WorkTypeUpdateError < RuntimeError; end
   LINK_TEXT = ':link will be inserted here automatically when available:'
 
   after_update_commit -> { work.broadcast_update }
-  after_update_commit -> { work.collection.broadcast_update_collection_summary }
```
jcoyne, line 39: We no longer update the whole collection summary at once, because depositor users don't see all the stuff in the summary.

jermnelson (approved):

PR #1470, opened 2021-05-18
Remove dynamic delete

jcoyne:
## Why was this change made?
We no longer need this complexity, so it has been removed.


app/components/collections/delete_button_component.rb
```diff
@@ -0,0 +1,29 @@
+# typed: false
+# frozen_string_literal: true
+
+module Collections
+  # Render the delete button for a collection
+  class DeleteButtonComponent < ApplicationComponent
+    CONFIRM_MESSAGE = 'Are you sure you want to delete this draft collection? It cannot be undone.'
```
aaron-collier, line 7: Is this button only available for draft collections? Or should this read `... delete this collection?`

jcoyne: This is just a refactor, so I'm taking the existing text from https://github.com/sul-dlss/happy-heron/pull/1470/files#diff-251c3b4af41a707fdfad74ccb17786d67aad7db340ff96eb4a5c10310f87d975L3

aaron-collier (approved):

PR #1478, opened 2021-05-18
Show Work Type Modal in Work Detail if purl_reservation is True

### LINKED ISSUE #1319: Edit icon on item details page for reserved PURLs should force item type selection.
Edit icon is forcing type selection for PURL reservation as expected from the dashboard, but not from the item details page.

jermnelson:
## Why was this change made?
Fixes #1319 

Unit and feature tests

n/a

### DIFF
--- a/app/components/collections/works_component.html.erb
+++ b/app/components/collections/works_component.html.erb
@@ -1,4 +1,4 @@
-<table class="table" data-controller="datatables">
+<table class="table" data-controller="datatables work-type">
   <thead class="table-light">
     <tr>
       <th width="18%"><%= I18n.t 'collection.deposits' %></th>
@@ -15,4 +15,5 @@
   <tbody>
     <%= render Collections::WorkRowComponent.with_collection(works.map(&:head)) %>
   </tbody>
+  <%= render Works::WorkTypeModalComponent.new %>
 </table>
--- a/app/components/purl_check_link_component.html.erb
+++ b/app/components/purl_check_link_component.html.erb
@@ -1,0 +1,17 @@
+<% if work_version.purl_reservation? %>
+  <%= link_to '',
+        data: {
+          destination: update_type_work_path(work_version.work),
+          form_method: 'patch',
+          bs_toggle: 'modal',
+          bs_target: '#workTypeModal',
+          action: "work-type#set_collection"
+        },
+        aria: { label: choose_label } do %>
+    <span class="fas fa-pencil-alt <%= css_class %>"</span>
+  <% end %>
+<% else %>
+  <%= link_to edit_work_path(work_version.work, anchor: anchor), aria: { label: edit_label } do %>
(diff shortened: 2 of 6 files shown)


jcoyne (changes requested):
I pulled this down and I'm still not seeing the popup that makes you select a type when you are on the view page and click edit:

Screen Shot 2021-05-18 at 1 57 10 PM

jermnelson:
Thanks @jcoyne, I didn't understand from the ticket's description about where the problem was occurring. Have a fix that works but tracking down some testing errors. 

> I pulled this down and I'm still not seeing the popup that makes you select a type when you are on the view page and click edit:

app/components/collections/work_row_component.html.erb
```diff
@@ -2,9 +2,10 @@
   <%= render Works::LinkToShowComponent.new(work_version: work_version) %>
   <td>
     <% if allowed_to? :edit?, work_version %>
-      <%= link_to [:edit, work], aria: { label: "Edit #{Works::DetailComponent.new(work_version: work_version).title}" } do %>
-        <span class="fas fa-pencil-alt"></span>
-      <% end %>
+      <%= helpers.turbo_frame_tag dom_id(work, :edit),
```
jermnelson, line 5: Now display work_type modal when clicking on the Edit button, fixes similar problem as described in ticket #1319

jcoyne: I don't understand why this block needs to change.  The point of edit_button_work_path is to check the authorization of the current user, but here we're already doing that on line 4.  So, this is adding unnecessary complexity.

jcoyne (changes requested):

jcoyne (changes requested):
The show page for the work looks good, but I'm still not seeing the correct behavior on this page:

Screen Shot 2021-05-27 at 3 43 54 PM

app/components/collections/work_row_component.html.erb
jermnelson: Removed this from the current PR, my plan was to use the new `PurlCheckLinkComponent` instead but was running into a problem where Stimulus controller was properly injecting the modal action attribute into the modal. This can be addressed in a follow-up PR.

jermnelson:
> The show page for the work looks good, but I'm still not seeing the correct behavior on this page:
> 
> Screen Shot 2021-05-27 at 3 43 54 PM

See comment above; I was having issues getting the workTypeModal's data attributes to be injected correctly.

jcoyne (approved):
@jermnelson okay, you may merge after you write a new ticket for that troublesome spot.

jermnelson:
> @jermnelson okay, you may merge after you write a new ticket for that troublesome spot.

Done, created ticket https://github.com/sul-dlss/happy-heron/issues/1546

PR #1484, opened 2021-05-19
Layout subtype options column first

### LINKED ISSUE #1334: List subtype options in the work type modal in columns from top to bottom instead of left to right.
This applies to both the short lists and the larger list displayed when the user clicks on "see more options"

jcoyne:
## Why was this change made?

Fixes #1334

### DIFF
--- a/app/components/works/work_type_modal_component.html.erb
+++ b/app/components/works/work_type_modal_component.html.erb
@@ -14,7 +14,7 @@
       </div>
 
       <div class="modal-body">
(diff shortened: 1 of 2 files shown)


ndushay (commented):
I could review this better if there were screenshots.  Also, has PO reviewed and signed off?

jcoyne:
@ndushay our practice has been to pushing to main and then the PO reviews all the progress made over the week. Since this hasn't been released to users yet, there is little need for tight controls.

ndushay (approved):

PR #1493, opened 2021-05-20
Display embargo requirement inline

### LINKED ISSUE #1001: Embargo date validation should appear inline, not just at top of page
When I attempt to deposit a work with only the year for the embargo date, the error only appears at the top of the page, not inline but the date entry fields:

Screen Shot 2021-04-28 at 11 28 23 AM

jcoyne:
Fixes #1001

### DIFF
--- a/spec/components/works/available_date_component_spec.rb
+++ b/spec/components/works/available_date_component_spec.rb
@@ -14,9 +14,15 @@ RSpec.describe Works::AvailableDateComponent, type: :component do
     work_form.prepopulate!
   end
 
-  it 'checks the immediate release radio button' do
+  it 'checks the immediate release radio button by default' do
     expect(rendered.css('#release_immediate[@checked]')).to be_present
     expect(rendered.css('#release_embargo[@checked]')).not_to be_present
+  end
+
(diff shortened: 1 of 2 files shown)


spec/components/works/available_date_component_spec.rb
```diff
@@ -14,11 +14,17 @@
     work_form.prepopulate!
   end
 
-  it 'checks the immediate release radio button' do
+  it 'checks the immediate release radio button by default' do
     expect(rendered.css('#release_immediate[@checked]')).to be_present
     expect(rendered.css('#release_embargo[@checked]')).not_to be_present
   end
 
+  it 'requires the date fields to be filled in' do
```
ndushay, line 22: quibble:   requires _all_ the date fields to be filed in ?

ndushay (approved):
added a quibble about spec description.  approving.

spec/components/works/available_date_component_spec.rb
jcoyne: Sounds good. Change made.

aaron-collier (approved):

PR #1494, opened 2021-05-20
Require the search parameter

### LINKED ISSUE #1491: [happy-heron/stage] NoMethodError: undefined method `start_with?' for nil:NilClass
## Backtrace

    line 8 of [PROJECT_ROOT]/app/controllers/searches_controller.rb: show

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/79494411)

jcoyne:
## Why was this change made?
The UIT security scanning software tries to fuzz this which triggers a Nil pointer error.  This change instead raises ActionController::ParameterMissing which is handled by ActionDispatch::ExceptionWrapper: https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L21

Fixes #1491

### DIFF
--- a/app/controllers/searches_controller.rb
+++ b/app/controllers/searches_controller.rb
@@ -4,7 +4,7 @@
 # When a user searches for a druid, direct them to the correct path.
 class SearchesController < ApplicationController
   def show
-    druid = params[:q]
+    druid = params.require(:q)
     druid = "druid:#{druid}" unless druid.start_with?('druid:')
     result = Work.find_by(druid: druid) || Collection.find_by(druid: druid)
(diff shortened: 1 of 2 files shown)


ndushay (approved):
question:  will anyone want to intentionally search "everything" ?   I think in some systems users do that by having an empty search param.  Not sure if that applies here.   Approving, but also wondering.

jcoyne:
@ndushay this is "search for one thing" not "search for a set of results", so empty search parameter is not going to be something we'd allow.

PR #1524, opened 2021-05-24
Add rake task to clean up unattached files after 7 days

### LINKED ISSUE #481: Add a cron job to look for ActiveStorage objects that are not attached to works
...and delete them.

The use case here is that someone starts to create a draft, uploads some files and then walks away before they ever save the draft.  We need a way to ensure these files are cleaned up.

aaron-collier:
## Why was this change made?

Fixes #481

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -7,3 +7,8 @@
 every :day, at: '1:00am' do
   runner 'WorkReminderGenerator.send_draft_reminders'
 end
+
+# Remove any files that did not get attached to objects after 7 days
+every :day, at: '12:00am' do
+  rake 'cleanup:uploads'
+end
--- a/lib/tasks/cleanup.rake
+++ b/lib/tasks/cleanup.rake
@@ -1,0 +1,13 @@
+# typed: false
+# frozen_string_literal: true
+
+namespace :cleanup do
+  desc 'Remove unattached files'
+  task uploads: :environment do
+    # Purge all blob keys were created 7 days ago and do not have attachments
+    ActiveStorage::Blob.left_outer_joins(:attachments)
+                       .where('DATE(active_storage_blobs.created_at) = ?', 7.days.ago.to_date)
+                       .where(attachments: { id: nil })
+                       .find_each(&:purge_later)
+  end
+end


lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,20 @@
+# typed: false
+# frozen_string_literal: true
+
+namespace :cleanup do
+  desc 'Remove unattached files'
+  task uploads: :environment do
+    # Retrieve all blob keys that were created more than 7 days ago and are not attached
+    unattached_file_keys = ActiveStorage::Blob.where('created_at < ?', 7.days.ago)
+                                              .where.not(key: AttachedFile.all.map(&:key))
```
jcoyne, line 9: This is going to run the system out of memory. The `AttachedFile.all.map(&:key)` part will grow without bound.  Can you figure out how to do this in a single query?

jcoyne (changes requested):

lib/tasks/cleanup.rake
aaron-collier: I was worried about this, thanks for the feedback. I'll think about this.

aaron-collier: @jcoyne, working through this, I don't think a single query is possible (or more accurately, advisable). Since we need to base this on entries in the blob table that do not exist int he attachments table, we're going to face returning a lot of records. In this pass, I query for a single days worth of blob IDs and use that for the basis of the additional queries.

Totally open to suggestion or pairing if you have an idea of how to simplify.

lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,28 @@
...
+  task uploads: :environment do
+    # Retrieve all blob keys that were created 7 days ago
+    uploaded_blobs = ActiveStorage::Blob.where('DATE(created_at) = ?', 7.days.ago)
+
+    # Retrieve file attachments for the above blobs
+    attached_files = ActiveStorage::Attachment.where(blob_id: uploaded_blobs)
+
+    # Get the ids and keys for blobs that do not have an Attachment
+    unattached_blobs = uploaded_blobs.map(&:id) - attached_files.map(&:blob_id)
+    unattached_file_keys = ActiveStorage::Blob.find(unattached_blobs).map(&:key)
```
jmartin-sul, line 15: a few random thoughts:
1. with this version of the PR, i think it'd be a win just switching from `map` to `pluck` (and dropping the `&` -- e.g. `uploaded_blobs.pluck(:id)`).  i believe that'll get you the same result, ultimately, but it'll do it without instantiating an active record object for each DB row, and i think it'll also select solely the column you're interested in, as opposed to all the table's columns (not a notable difference for most queries, but might make a difference when iterating over a potentially large result set like this).
2. iirc from what i saw of yesterday's version, i _think_ the subquery approach could work as a single query if you traded what was a `map` for a `select` -- my recollection is that for subqueries like that, ActiveRecord is smart enough to use the `Relation` from the `select` call to build a single whole SQL query (instead of running the inner select, retrieving all the results, mapping over the result set to get IDs, and then feeding those IDs to the outer query and running that second query separately).
3. another approach that i think is usable as an alternative to the subquery approach would be an [outer join](https://apidock.com/rails/v6.1.3.1/ActiveRecord/QueryMethods/left_outer_joins) to the table where you want to find no entries, with `nil` as the matching condition  on that table.  something like... `ActiveStorage::Blob.left_outer_joins(:active_storage_attachments).where('DATE(active_storage_blobs.created_at) < ?', 7.days.ago).where(blob_id: nil)` -- or something, i'm writing that freehand in a comment box without testing, so grain of salt 🙂 

here's an example where a `not` subquery is built using a scope instead of a concrete retrieved result set:  https://github.com/sul-dlss/preservation_catalog/blob/2dc297dd416f5264902fabb919805cac666a26db/app/models/zip_endpoint.rb#L26-L38

my recollection is i went that way on that pres cat query because i compared the plan explanation for that with the outer join approach, and the outer join approach was less efficient.

if none of the single shot approaches work, you could always fall back to iterating over `uploaded_blobs` using `uploaded_blobs.find_each`, and querying for each blob within that -- unlikely to make anything fall over, but also seems likely to take quite a while to run (and less fun imho than building one nice clean SQL query, but i'm a weirdo who likes SQL 🙂).

lib/tasks/cleanup.rake
jmartin-sul: i think simply turning the inner `map(&:key)` into `select(:key)` might get you a single-shot query -- iirc, ActiveRecord is smart enough to use that inner `Relation` to build a single larger query, though here it can't do that here since the `map` gives a concrete list of already retrieved values.

config/schedule.rb
```diff
@@ -1,9 +1,17 @@
 # typed: false
 # frozen_string_literal: true
 
+# Need the rails app to get Settings
```
jcoyne, line 4: I don't see us using Settings here. still needed?

lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,22 @@
+# typed: false
+# frozen_string_literal: true
+
+namespace :cleanup do
+  desc 'Remove unattached files'
+  task uploads: :environment do
+    # Retrieve all blob keys that were created 7 days ago and do not have attachments
+    unattached_blobs = ActiveStorage::Blob.left_outer_joins(:attachments)
+                                          .where('DATE(active_storage_blobs.created_at) = ?', 7.days.ago)
```
jcoyne, line 9: Do we need `7.days.ago.to_date` here?

lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,22 @@
...
+                                          .where(attachments: { id: nil })
+                                          .pluck(:id, :key)
+
+    # Remove the blob and the file from the disk
+    unattached_blobs.each do |blob_id, key|
+      ActiveStorage::Blob.find(blob_id)
+      uploaded_file_path = File.join(Settings.file_uploads_root, key[0, 2], key[2, 2], key)
```
jcoyne, line 16: I'm pretty sure you can do `blob.purge_later` and then you don't have to worry about all this file stuff.

jcoyne (changes requested):

lib/tasks/cleanup.rake
jcoyne: https://github.com/rails/rails/blob/328113b0adfb516639ea0ad9845e15877b8c8141/activestorage/app/models/active_storage/blob.rb#L303

config/schedule.rb
aaron-collier: Nope - good call, will remove.

lib/tasks/cleanup.rake
aaron-collier: Good call, that simplified this into a one liner.

lib/tasks/cleanup.rake
```diff
@@ -0,0 +1,13 @@
...
+# frozen_string_literal: true
+
+namespace :cleanup do
+  desc 'Remove unattached files'
+  task uploads: :environment do
+    # Purge all blob keys were created 7 days ago and do not have attachments
+    ActiveStorage::Blob.left_outer_joins(:attachments)
+                       .where('DATE(active_storage_blobs.created_at) = ?', 7.days.ago.to_date)
+                       .where(attachments: { id: nil })
+                       .each(&:purge_later)
```
jmartin-sul, line 11: not sure what the expected number of results will look like... guessing it'll be in the tens or hundreds, which is prob fine for this, though if you expected it to be 1000+, you might want to `find_each` instead of `each`, so that it batches things.

jmartin-sul (approved):
one nitpicky suggestion, but LGTM.  also refraining from merging since i paired (or maybe that's not a general rule for all our projects, just a couple i've been on, couldn't remember for sure).

lib/tasks/cleanup.rake
aaron-collier: Done.

jmartin-sul (approved):

lib/tasks/cleanup.rake
jmartin-sul: approved again 🙂

jcoyne (approved):

Issue #540, opened 2020-11-20
Saving a collection as a draft should take you to the collection view page.


amyehodge:
User is currently taken back to the dashboard.

PR #1532, opened 2021-05-25
Redirect to the collection page on deposit

### LINKED ISSUE #1003: When a user deposits a collection, they should be taken to the collection view page, not the dashboard.
This would mimic the behavior of depositing an item, after which the user is taken to the item view page.

aaron-collier:
## Why was this change made?

Fixes #1003


jcoyne (approved):

jcoyne:
This complements #540

PR #1541, opened 2021-05-27
Fix deposit button alignment.

### LINKED ISSUE #1476: Correct display issue with "Edit or deposit" button so that it is always inline with the title and the other button on the top right.
On https://sul-h2-stage.stanford.edu/collections/560 you can see that this button is shifted below the other content.

justinlittman:
closes #1476

## Why was this change made?
Fix alignment when collection name is long.

Locally

NA

### DIFF
--- a/app/components/collections/show/details_header_component.html.erb
+++ b/app/components/collections/show/details_header_component.html.erb
@@ -1,25 +1,30 @@
 <header class="title">
+  <% if depositing? %>
+    <span class="state">Depositing <%= spinner %></span>
+  <% end %>
+  <div class="row">
+    <div class="col"> </div>
+    <div class="col-4">
+      <%= edit_button %>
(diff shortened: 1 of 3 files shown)


justinlittman:
Before:
image

justinlittman:
After:
image

jcoyne (changes requested):
Screen Shot 2021-05-27 at 3 27 22 PM
I'm seeing this layout on shorter titles.  Is that what is intended @astridu ?

justinlittman:
@astridu Is this acceptable?

astridu:
It's ok, but it would be nicer if the button was aligned with the title if you have the time. 
image

jcoyne (approved):

justinlittman:
Skill is the limiting factor @astridu.

PR #1544, opened 2021-05-27
Add sorting to the all collections table on the admin dashboard

### LINKED ISSUE #1025: Make all columns other than status in the Approvals table on the dashboard sortable
I believe the only items that show up here are "Pending approval" so there is no need to sort by status. If you see that this not the case, please let me know.

aaron-collier:
## Why was this change made?

Fixes #1025 

Default sorting is the first column ascending:

Screen Shot 2021-05-27 at 2 58 35 PM

Sorting columns:

Last modified:

Screen Shot 2021-05-27 at 2 56 43 PM

Total:

Screen Shot 2021-05-27 at 2 56 35 PM


app/components/collections/works_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<table class="table" data-controller="datatables">
+<table class="table" id="worksTable" data-controller="datatables" data-datatables-selector-value="works">
```
aaron-collier, line 1: Note: adding this `data-tables-selector-value` is required so that more than one table can be on a page. Otherwise the first table with the given class is styled.

app/packs/stylesheets/allCollections.scss
```diff
@@ -24,4 +24,8 @@ $collapse-icon: url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg'
       }
     }
   }
+
+  .dataTable-search {
```
aaron-collier, line 28: The `searching: false` configuration (from the documentation) does not work on DataTables so hiding the search within this page via css was the easiest way to disable it.

app/packs/controllers/datatables_controller.js
```diff
@@ -2,11 +2,24 @@ import { Controller } from "stimulus"
...
 export default class extends Controller {
+  static values = { selector: String }
+
   connect() {
-    new DataTable("table", {
-      columns: [
-        { select: 4, sort: "desc" },  // Sort the fifth column in ascending order
-        { select: [1, 5, 6, 7, 8], sortable: false } // Disable unsortable columns
-      ]})
+    if (this.selectorValue == 'works') {
```
jcoyne, line 8: Rather than having this conditional, how about splitting this into two separate controllers?  On for collections and one for works?

jcoyne (changes requested):

app/packs/controllers/datatables_controller.js
aaron-collier: good idea, done. Thanks :-)

jcoyne (approved):

PR #1558, opened 2021-06-01
DropZone checks for duplicate filenames

### LINKED ISSUE #1327: Provide feedback to user when they attempt to upload two files with the same name.
## Backtrace

    line 41 of [PROJECT_ROOT]/app/jobs/deposit_job.rb: create_or_update
    line 29 of [PROJECT_ROOT]/app/jobs/deposit_job.rb: deposit
    line 10 of [PROJECT_ROOT]/app/jobs/deposit_job.rb: perform

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/78735927)

jermnelson:
## Why was this change made?
Fixes #1327

Tested in a feature test.

n/a

### DIFF
--- a/app/packs/controllers/dropzone_controller.js
+++ b/app/packs/controllers/dropzone_controller.js
@@ -10,7 +10,7 @@ import {
 } from "helpers";
 
 export default class extends Controller {
-  static targets = ["input", "previewsContainer", "preview", "template", "feedback", "container"];
+  static targets = ["input", "previewsContainer", "preview", "template", "feedback", "container", "fileName"];
 
   connect() {
     this.dropZone = createDropZone(this, this.templateTarget.innerHTML)
@@ -36,12 +36,12 @@ export default class extends Controller {
   }
 
   checkForDuplicates(fileName) {
-    // Current files in dropZone
-    const fileNames = this.dropZone.files.map(file => { return file.name })
+    // Extract all filenames
+    const fileNames = this.fileNameTargets.map(target => { return target.innerText.trim() })
+
     // Remove current fileName
-    fileNames.pop(fileName)
-    // Existing files for this Work
-    this.previewTargets.map(elem => { fileNames.push(elem.children[1].innerText.trim()) })
+    fileNames.splice(fileNames.indexOf(fileName), 1)
+
     return fileNames.indexOf(fileName) > -1
   }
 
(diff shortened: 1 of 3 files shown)


app/packs/controllers/dropzone_controller.js
```diff
@@ -35,6 +35,16 @@ export default class extends Controller {
     }
   }
 
+  checkForDuplicates(fileName) {
+    // Current files in dropZone
+    const fileNames = this.dropZone.files.map(file => { return file.name })
+    // Remove current fileName
+    fileNames.pop(fileName)
```
jcoyne, line 42: Does pop take arguments?  https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/pop

app/packs/controllers/dropzone_controller.js
```diff
@@ -35,6 +35,16 @@ export default class extends Controller {
...
+    // Existing files for this Work
+    this.previewTargets.map(elem => { fileNames.push(elem.children[1].innerText.trim()) })
```
jcoyne, line 44: It looks like you are doing a map, but aren't doing anything with the result. Would `forEach` be a better choice?

jcoyne: Do we need line 40?  Don't each of them get added to the previewTargets?

jcoyne: Instead of walking the DOM tree here, can you add a stimulus target attributes (`data-dropzone-target="fileName"`) to the fields and use `this.fileNameTargets`?

jcoyne (changes requested):

app/packs/controllers/dropzone_controller.js
```diff
@@ -75,10 +87,14 @@ export default class extends Controller {
 
   bindEvents() {
     this.dropZone.on("addedfile", file => {
+      if (this.checkForDuplicates(file.name) === true) {
```
jcoyne, line 90: Do we need to compare here? Isn't this method returning a boolean?

jcoyne (changes requested):

app/packs/controllers/dropzone_controller.js
jermnelson: Fixed in latest commit; using `Array.splice`

app/packs/controllers/dropzone_controller.js
jermnelson: Using new stimulus target attributes, removed dropZone in latest commit.

app/packs/controllers/dropzone_controller.js
jermnelson: Updated in latest commit.

jcoyne (approved):

PR #1562, opened 2021-06-01
Ensure the embargo date is in the future

### LINKED ISSUE #1384: Display error message inline with embargo date field when an embargo date in the past is chosen for a versioned work
I attempted to version a work and reset a valid embargo date to a date in the past. When I click "Deposit", the page appears to reload with my "what's changing" comment gone and the embargo date reset to the previous value. No error messages are displayed.

EmbargoDate

jcoyne:
## Why was this change made?

Fixes #1384

Screen Shot 2021-06-01 at 5 24 25 PM

### DIFF
--- a/app/packs/controllers/available_date_controller.js
+++ b/app/packs/controllers/available_date_controller.js
@@ -1,0 +1,88 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = ["year", "month", "day", "error"]
+
+  validate() {
+    this.errors = {}
+    this.allPartsPresent()
+    this.dateInFuture()
+    this.displayErrors()
+  }
+
+  showError(err) {
+    if (err) {
+      this.errorTarget.innerHTML = err
+      this.errorTarget.style.display = 'block'
+    } else {
+      this.hideError()
+    }
+  }
+
+  hideError() {
+    this.errorTarget.style.display = 'none'
+  }
+
+  displayErrors () {
+    if (this.errors.year) {
+      this.yearTarget.classList.add('is-invalid')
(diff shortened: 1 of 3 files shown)


app/packs/controllers/available_date_controller.js
```diff
@@ -0,0 +1,88 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = ["year", "month", "day", "error"]
+
+  validate() {
```
justinlittman, line 6: (This is a general comment that does not apply to this specific line of code.)

This provides a message to the user, but does not ensure the embargo date is in the future. The user can still submit an embargo date from the past and it will be accepted.

justinlittman (changes requested):

app/packs/controllers/available_date_controller.js
jcoyne: Setting the setCustomValidity in displayErrors actually prevents the form from being submitted.

jcoyne: See https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Constraint_validation#complex_constraints_using_the_constraint_validation_api

justinlittman: I'm not following. I ran the code locally and was able to submit the form.

jcoyne: @justinlittman did you select "Save Draft" or "Deposit"?  The validations only run when you do "Deposit"

justinlittman: Oh ... testing again.

justinlittman (approved):

PR #1566, opened 2021-06-02
Send email to managers on collection deposit

### LINKED ISSUE #1302: Additional managers listed when a collection is created need to receive the notification
Currently, when a collection manager creates a new collection and includes additional managers at that time -- as opposed to adding them in later -- those additional managers are not receiving the email notifications that they have been added to the collection. They should be receiving this notification: https://docs.google.com/document/d/1TZVILUrORI25kRZyUZ70vMQYo90Ngp_tqCI7BdU_VyQ/edit#heading=h.wm12u0xf3u0l

This notification IS being properly sent when additional managers are added to an already deposited collection.

aaron-collier:
## Why was this change made?

Fixes #1302

### DIFF
--- a/app/models/collection_version.rb
+++ b/app/models/collection_version.rb
@@ -33,6 +33,11 @@ class CollectionVersion < ApplicationRecord
           CollectionsMailer.with(collection_version: collection_version, user: reviewer)
                            .review_access_granted_email.deliver_later
         end
+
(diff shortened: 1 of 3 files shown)


jcoyne (changes requested):
Would you please add a test?

jcoyne (approved):

PR #1576, opened 2021-06-03
file upload checkmark now persistent

### LINKED ISSUE #1378: Add in a persistent checkmark for completion of file upload
Several testers indicated they weren't sure that files were done uploading. To help make this clearer, we want to remove the black checkmark that appears in the file description box and then goes away, and replace it with one that displays persistently while the file upload is complete.

ndushay:
## Why was this change made?

Fixes #1378

Makes file upload checkmark persistent.    

image

NOTE:  Amy is aware that the checkmark position isn't consistent when the screen size changes, and isn't too concerned.  This is the reason the checkmark is black instead of the green in the original ticket.

Locally, so far.  On stage for @amyehodge  as I type.


app/packs/stylesheets/dropzone.scss
```diff
@@ -126,13 +126,15 @@
...
-  -moz-animation: passing-through 3s cubic-bezier(0.77, 0, 0.175, 1);
-  -ms-animation: passing-through 3s cubic-bezier(0.77, 0, 0.175, 1);
-  -o-animation: passing-through 3s cubic-bezier(0.77, 0, 0.175, 1);
-  animation: passing-through 3s cubic-bezier(0.77, 0, 0.175, 1); }
+  opacity: 1;
+  -webkit-animation: slide-in 3s cubic-bezier(0.77, 0, 0.175, 1);
+  -moz-animation: slide-in 3s cubic-bezier(0.77, 0, 0.175, 1);
+  -ms-animation: slide-in 3s cubic-bezier(0.77, 0, 0.175, 1);
+  -o-animation: slide-in 3s cubic-bezier(0.77, 0, 0.175, 1);
+  animation: slide-in 3s cubic-bezier(0.77, 0, 0.175, 1); }
```
jcoyne, line 134: Can you remove the now unused keyframes from dropzoneKeyframes.scss?

jcoyne (changes requested):

app/packs/stylesheets/dropzone.scss
ndushay: will do.

app/packs/stylesheets/dropzone.scss
```diff
@@ -166,14 +162,14 @@
   -o-animation: pulse 6s ease infinite;
   animation: pulse 6s ease infinite; }
 .dz-preview.dz-error .dz-error-message {
-  display: block; }
-.dz-preview.dz-error:hover .dz-error-message {
```
jcoyne, line 170: This is about errors, so can this be reverted?

app/packs/stylesheets/dropzone.scss
```diff
@@ -142,16 +144,10 @@
   pointer-events: none;
   opacity: 0;
   z-index: 500;
-  position: absolute;
   display: block;
-  top: 50%;
-  left: 50%;
-  margin-left: -27px;
```
jcoyne, line 149: This is about errors, can it be reverted?

jcoyne (changes requested):

app/packs/stylesheets/dropzone.scss
ndushay: it can be, but what it does is make the error message itself hovertext over the fileupload error-mark.   So I don't think we'll want it in any case.

app/packs/stylesheets/dropzone.scss
ndushay: it can be, but ... should it be?    There was code in main that dealt with the error case and stomps on dropzone default behavior ... so ... 🤷🏻

ndushay: that is ... it's not like this will harm anything, and the current behavior is not good in any case, and I made ticket #1577 ...

jcoyne: I'm fine if this is in the PR for 1577, but it's confusing when you look at history and see that this is about fixing the checkmark, but it's adjusting styles that are unrelated.  Let's keep each change in its own PR.

ndushay: actually, if you look at the whole file, these particular changes pertain to both success-mark and error-mark.

jcoyne: Got it, github had folded up the end of the last block.

jcoyne (approved):

Issue #1574, opened 2021-06-02
"What's changing" element on collections page should display error message when not filled in


amyehodge:
When it does not pass the validation it should display the same error message as is displayed on the items page.

Current collections validation error:
 ![Screen Shot 2021-06-02 at 5.00.39 PM.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/b38f663b-0458-47b4-9574-f18bbbba9791)!

Items validation error:
![Screen Shot 2021-06-02 at 5.05.51 PM.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/8fb5d10c-3c6f-43f7-b47b-2e2649e49436)

Text of error message: You must describe your changes.

ndushay:
what about the asterisk?  Should it be present for both?  Absent for both?

jcoyne:
Is "Version you **Work**" the correct text?  Should it be "Version your collection"?

amyehodge:
@ndushay Please asterisk both. And good catch, @jcoyne . This should say "Version your collection". It takes a village.

PR #1589, opened 2021-06-04
Update text for version your collection

jcoyne:
## Why was this change made?
See https://github.com/sul-dlss/happy-heron/issues/1574#issuecomment-854725760

### DIFF
--- a/spec/components/collections/version_description_component_spec.rb
+++ b/spec/components/collections/version_description_component_spec.rb
@@ -12,8 +12,8 @@ RSpec.describe Collections::VersionDescriptionComponent do
   context 'with a first draft' do
     let(:collection_version) { build_stubbed(:collection_version) }
 
-    it 'does not renders the component' do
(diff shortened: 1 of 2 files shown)


spec/components/collections/version_description_component_spec.rb
```diff
@@ -13,15 +13,15 @@
     let(:collection_version) { build_stubbed(:collection_version) }
 
     it 'does not renders the component' do
```
ndushay, line 15: "render" not "renders"

ndushay (approved):
quibble with spec description (that was wrong before);  fix or don't.

spec/components/collections/version_description_component_spec.rb
jcoyne: Thank you. I've made the fix. Also added the "*" to the expectation.

ndushay (approved):

PR #1595, opened 2021-06-04
approvals table: link collection names to collection view page

### LINKED ISSUE #1024: Link collection names in the Approvals table on the dashboard to the collection view page

ndushay:
## Why was this change made?

Fixes #1024

I asked Amy if we should truncate collection titles at 150.  "I just checked the collections we are migrating and the longest title is 139 characters."

### Before

image

### After

image

locally, and a unit test.

~~I wasn't sure what to test with code -- the analogous Works::LinkToShowComponent only tests that titles are truncated ... but collection titles don't need to be truncated ...~~ .


spec/features/mediated_deposit_versioning_spec.rb
```diff
@@ -37,6 +37,13 @@
       expect(page).to have_content 'You have successfully submitted your deposit'
       click_link 'Return to dashboard'
 
+      # check collection link in approvals table
+      within_table('Approvals') do
```
jcoyne, line 41: I'd prefer that we keep this test focused on doing a mediated deposit.  Can this be in a component test instead?

ndushay: FYI, as I mentioned, the analogous Works::LinkToShowComponent only tests that titles are truncated, nothing about the link.

ndushay:
@jcoyne ready for re-review.

app/components/collections/link_to_show_component.html.erb
```diff
@@ -0,0 +1 @@
+<td class="collection-title"><%= link %></td>
```
jcoyne, line 1: I'm not terribly fond of having the `<td>` in this component, as that limits where we're able to use it.  Can that be moved back into ApprovalsComponent?

jcoyne (changes requested):

app/components/dashboard/approvals_component.html.erb
```diff
@@ -16,7 +16,7 @@
       <tr>
         <%= render Works::LinkToShowComponent.new(work_version: submission) %>
```
ndushay, line 17: I heard what you are saying about the <td> but look at line 17 here ...

ndushay:
back to you, @jcoyne

jcoyne (approved):

PR #1606, opened 2021-06-08
Allow Authors to be reordered

### LINKED ISSUE #977: A depositor should be able to order the authors
Screen Shot 2021-02-01 at 12 35 40 PM

jcoyne:
## Why was this change made?
Fixes #977

Screen Shot 2021-06-08 at 1 38 36 PM


app/components/works/contributor_row_component.html.erb
```diff
@@ -1,4 +1,4 @@
-<div class="row inner-container" data-controller="contributors" data-contributors-required-value="<%= required? %>"">
+<div class="row inner-container" data-controller="contributors" data-contributors-required-value="false"">
```
aaron-collier, line 1: Extra double quote here?

jcoyne: Good catch. I fixed it.

Issue #264, opened 2020-10-28
Add autocomplete for keywords 


jcoyne:
These should come from OCLC FAST authority hosted by Cornell's QA.

See conversation in #82

ndushay:
preferred component for typeahead?   "something that doesn't include jQuery"

we might need to proxy the back end ... can the front end query OCLC FAST directly?  (Do we want to have a rails controller query the QA server and return results to the front end?)

Using Cornell's QA has a search; not sure if OCLC FAST does;  Cornell's QA is implemented with blacklight so can query Solr ...

Cornell's QA is associated with LD4P, so Sinopia may have some code we can use ...

ndushay:
Possible typeahead components:
- https://github.com/gch1p/bootstrap-5-autocomplete/
- https://www.algolia.com/doc/ui-libraries/autocomplete/introduction/what-is-autocomplete/ "This is a good one too, but more heavyweight" (JCoyne)
    - Source at https://github.com/algolia/autocomplete
-  https://github.com/alphagov/accessible-autocomplete "May also want to look at ... as it’s geared toward accessability.
" (JCoyne)

jcoyne:
@ndushay at some point the requirements for the front end, might become so divergent from what OCLC offers that it makes sense for us to build our own backend for the search.

ndushay:
@jcoyne understood.  Let's call this "phase 1" and see what happens when I make the requested changes.

ndushay:
if we did do our own backend we might want to use it for ETDs and possibly some other apps as well.

PR #1615, opened 2021-06-08
add autocomplete_controller as a proxy to QA lookup server

ndushay:
## Why was this change made?

Autocomplete will require a lookup to OCLC FAST authority hosted by Cornell's QA.  This adds a controller to perform that so the javascript code can use the local controller and not require as much code (e.g. for error responses).

Part of #264

request spec.

Added settings needed for accessing Cornell's QA server.


ndushay:
Note:  tests are passing;  the github hook must be slow ...

app/controllers/autocomplete_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+      req.params['maxRecords'] = Settings.autocomplete_lookup.max_records
+    end
+  end
+
+  def lookup_connection
+    @lookup_connection ||= Faraday.new(
+      url: Settings.autocomplete_lookup.url,
+      headers: {
+        'Accept' => 'application/json',
+        'User-Agent' => 'Stanford Self-Deposit (Happy Heron)'
```
jcoyne, line 31: 😄

jcoyne (approved):

PR #1621, opened 2021-06-09
Line break 'Software/Code' when necessary on the worktype modal

### LINKED ISSUE #1504: On work type modal, fix alignment of "software/code" text so it isn't up against the edge of the box
Can we get a little space in there somehow? I'm not sure if you need @astridu to make a suggestion here or not.

aaron-collier:
## Why was this change made?

Fixes #1504 

Screen Shot 2021-06-11 at 10 13 25 AM

Screen Shot 2021-06-11 at 10 13 31 AM

### DIFF
--- a/spec/services/types_generator_spec.rb
+++ b/spec/services/types_generator_spec.rb
@@ -523,7 +523,7 @@ RSpec.describe TypesGenerator do
               {
                 structuredValue: [
                   {
-                    value: 'Software/Code',
+                    value: 'Software/<wbr>Code',
                     type: 'type'
                   },
                   {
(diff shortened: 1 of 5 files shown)


spec/services/types_generator_spec.rb
```diff
@@ -523,7 +523,7 @@
               {
                 structuredValue: [
                   {
-                    value: 'Software/Code',
+                    value: 'Software/<wbr>Code',
```
jcoyne, line 526: This is incorrect. There should be no HTML here.

aaron-collier: We do the lookup based on the label, so without this, the look fails. I'm wondering, based on that, if we should just remove the `/Code` because it's the problem here and would then require code to strip the `<wbr>` before the lookup... seems wrong to me

jcoyne: Have you considered adding a new separate `html_label` property to the WorkType model?

jcoyne (approved):

PR #1624, opened 2021-06-09
Add datatable controller for sorting the approvals table

### LINKED ISSUE #1387: Make columns in the Approvals table on the dashboard sortable
I have separated this out from the original ticket #1025 it was on as this is lower priority than the other task there.

For Approvals table, all columns other than the status should be sortable. I believe the only items that show up here are "Pending approval" so there is no need to sort by status. If you see that this not the case, please let me know.

Approvals table: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.2qm8sw1ugwl

aaron-collier:
## Why was this change made?

Fixes #1387 

Screen Shot 2021-06-09 at 11 43 34 AM

### DIFF
--- a/app/packs/controllers/datatable_approvals_controller.js
+++ b/app/packs/controllers/datatable_approvals_controller.js
@@ -3,7 +3,7 @@ import { DataTable } from "simple-datatables"
 
 export default class extends Controller {
   connect() {
-    new DataTable("#approvalsTable", {
+    new DataTable(this.element, {
       paging: false,
       columns: [
         { select: 0, sort: "asc" },  // Sort the first column in ascending order
--- a/app/packs/controllers/datatable_approvals_controller.js
+++ b/app/packs/controllers/datatable_approvals_controller.js
@@ -1,0 +1,14 @@
+import { Controller } from "stimulus"
+import { DataTable } from "simple-datatables"
+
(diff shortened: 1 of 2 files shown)


app/packs/controllers/datatable_approvals_controller.js
```diff
@@ -0,0 +1,14 @@
+import { Controller } from "stimulus"
+import { DataTable } from "simple-datatables"
+
+export default class extends Controller {
+  connect() {
+    new DataTable("#approvalsTable", {
```
jcoyne, line 6: we can avoid the extra lookup and need for an id by passing `this.element` rather than a string. I know the others don't do this, but I feel this would be less fragile.

jcoyne (changes requested):

app/packs/controllers/datatable_approvals_controller.js
aaron-collier: @jcoyne thanks! This is done. However, I'll not that I tried this for the other tables and it didn't work (they did not get rendered actually). I'll take another poke at refactoring them in a separate PR if we want parity there.

jcoyne (approved):

PR #1635, opened 2021-06-10
Make the link-color digital-blue

jcoyne:
## Why was this change made?

This helps align with the mockups.


ndushay (approved):
In a perfect world, your commit comment would say WHICH links, but meh.

jcoyne:
@ndushay this is for all of them.

PR #1636, opened 2021-06-11
Autocomplete added to work form for keywords

### LINKED ISSUE #264: Add autocomplete for keywords

ndushay:
## Why was this change made?

It provides suggestions when users type in the keyword text box in the work form.

image

Closes #264 

locally, so far and unit tests for the autocomplete_controller.


app/views/autocomplete/show.erb
```diff
@@ -0,0 +1,3 @@
+<% @suggestions.each_with_index.map do |suggestion, i| %>
```
jcoyne, line 1: Is this file used by anything?

spec/requests/autocomplete_spec.rb
```diff
@@ -11,48 +11,73 @@
...
+        'Broncobusters',
+        'Information brokers',
+        'New York (State)--New York--Brooklyn'
+      ]
+    end
     let(:lookup_resp_body) do
       <<~JSON
-        [
-          {
-            "uri": "http://id.worldcat.org/fast/170115",
```
jcoyne, line 21: I believe we want to record the URI too. Does this still permit that?

app/views/autocomplete/show.erb
ndushay: yes - it's how the autocomplete#show command takes the suggestions and serves them up the way the stimulus-autocomplete plugin needs.

jcoyne: I see that conditional that would render this. Should it have an `html.erb` suffix?

ndushay: It didn't work with the html.erb suffix.

It's not html, it's an html fragement, so I typed it as "text".   If it were typed as html, but then I think we'd get the full html headers and such.

jcoyne: Did you try passing `layout: false`?

spec/requests/autocomplete_spec.rb
ndushay: The best we can do is get idroot back from the request https://www.oclc.org/developer/develop/web-services/fast-api/assign-fast.en.html

We can get "idroot" back from autosuggest, and then manipulate it, but it's not always available (maybe only for subjects)  https://fast.oclc.org/searchfast/fastsuggest?callback=&query=mcrae&queryIndex=suggestall&queryReturn=suggestall,idroot&rows=10

image

From their own examples, there seems to be a "suggest" param, but it's not documented:

`suggest=autosuggest` from their first example on their doc page doesn't work, and `suggest=autoSubject` does work but has different results from leaving it out.

app/views/autocomplete/show.erb
ndushay: I don't understand - would `layout : false` be in the controller?   

Are you suggesting I put a method in the controller to do the reformatting?  Or that I add `layout : false` and then use `html.erb` as the suffix?   Maybe a realtime conversation would be better.

spec/requests/autocomplete_spec.rb
jcoyne: @ndushay would the Linked Data API have what we need? https://www.oclc.org/developer/develop/web-services/fast-api/linked-data.en.html

jcoyne: @arcadia Do you have ideas about what we ought to do when there is no identifier for the topic in FAST?

ndushay: @arcadiafalcone ^^

arcadiafalcone: The ones without identifiers are alternate labels and have identifiers under the main label -- for example, "McRae, Col." is an alternate label for "McCrae, John, 1872-1918," which does have an identifier: http://id.worldcat.org/fast/170115. So there may need to be a second lookup to get the main label and identifier. Though that may be weird for the user, as the term will be different from what they select. I suspect that ideally we would only present the main terms to the user to reduce confusion, but I don't know if that's possible.

ndushay:
@jcoyne I already had to change the description generator spec, because I had to change the description generator to get this working.  Do we still need a separate ticket for adding tests for this?

amyehodge:
This looks good to me.

app/views/autocomplete/show.erb
jcoyne: Yes. You should have a `render layout: false` in the controller and then rename this with an `html.erb` suffix.

jcoyne (changes requested):

jcoyne (approved):

PR #1646, opened 2021-06-14
Remove indent under your Collections.

### LINKED ISSUE #1632: Remove indent under "Your collections"
I think this is currently on `ul.collections`

justinlittman:
closes #1632

## Why was this change made?
Fix indent.

Local

NA

### DIFF
--- a/app/packs/stylesheets/dashboard.scss
+++ b/app/packs/stylesheets/dashboard.scss
@@ -38,4 +38,5 @@
 
 ul.collections {
   list-style-type: none;
+  padding-left: 0;
 }


justinlittman:
Before:


After:

app/packs/stylesheets/dashboard.scss
```diff
@@ -38,4 +38,5 @@
 
 ul.collections {
   list-style-type: none;
+  padding-left: 0rem;
```
jcoyne, line 41: I believe the convention is to remove the units for a "0" value. See https://github.com/CSSLint/csslint/wiki/Disallow-units-for-zero-values

justinlittman: Ahhh ... fixed.

jcoyne (approved):

PR #1656, opened 2021-06-15
If providing a create date interval, requires start and finish year.

### LINKED ISSUE #914: Make sure date entries for a date range are saved when saving drafts.
The validation here should require that at least two years be entered for a date range. When I enter a single year...

The information is not saved as seen in the item view...

...and when I open the item back up to edit again.

The desired behavior is for the entered information to be saved for the drafts but for the item not to be depositable unless at least years are entered for the two ends of the range. 

In the current situation, the entered data is just lost to the user and the user is not even notified about this.

justinlittman:
closes #914

## Why was this change made?
* Validate date intervals.
* Do not discard interval dates that the user has entered.

Unit, local

NA

### DIFF
--- a/app/components/works/buttons_component.rb
+++ b/app/components/works/buttons_component.rb
@@ -18,7 +18,7 @@ module Works
     private
 
     delegate :object, to: :form
-    delegate :first_draft?, :version_draft?, :title, to: :work_version
+    delegate :title, to: :work_version
     delegate :collection, to: :model
 
     def work_version
@@ -32,6 +32,16 @@ module Works
     sig { returns(T.nilable(T::Boolean)) }
     def work_in_reviewed_coll?
       collection.review_enabled?
+    end
+
+    sig { returns(T::Boolean) }
+    def show_version_draft_cancel?
+      work_version.version_draft? && work_version.persisted?
+    end
+
+    sig { returns(T::Boolean) }
+    def show_first_draft_cancel?
+      work_version.first_draft?
     end
   end
 end
--- a/app/components/works/dates_component.html.erb
+++ b/app/components/works/dates_component.html.erb
@@ -26,7 +26,7 @@
             id: "#{prefix}_created_year",
             placeholder: 'year',
             class: "form-control", min: min_year, max: max_year %>
-        <div class="invalid-feedback">Created year must be between <%= min_year %> and <%= max_year %></div>
+        <div class="invalid-feedback" data-date-validation-target="error"></div>
       </div>
       <div class="month">
         <label for="<%= prefix %>_created_month" class="visually-hidden">Created month</label>
@@ -60,13 +60,14 @@
         <div class="year">
           <label for="<%= prefix %>_created_range_start_year" class="visually-hidden">Created range start year</label>
           <%= date_range_start_year %>
-          <div class="invalid-feedback">Created year must be between <%= min_year %> and <%= max_year %></div>
+          <div class="invalid-feedback" data-date-validation-target="error"></div>
+          <div class="invalid-feedback" data-date-range-target="startError"></div>
         </div>
         <div class="month">
           <label for="<%= prefix %>_created_range_start_month" class="visually-hidden">Created range start month</label>
           <%= select_month created_range_start_month,
               { prefix: prefix, field_name: 'created_range(2i)', prompt: 'month'},
-              data: { date_validation_target: 'month', action: 'date-validation#change' },
+              data: { date_validation_target: 'month', date_range_target: 'startMonth', action: 'date-range#change' },
               id: "#{prefix}_created_range_start_month",
               class: "form-control" %>
         </div>
@@ -74,7 +75,7 @@
           <label for="<%= prefix %>_created_range_start_day" class="visually-hidden">Created range start day</label>
           <%= select_day created_range_start_day,
               { prefix: prefix, field_name: 'created_range(3i)', prompt: 'day'},
-              data: { date_validation_target: 'day', action: 'date-validation#change' },
+              data: { date_validation_target: 'day', date_range_target: 'startDay', action: 'date-range#change' },
               id: "#{prefix}_created_range_start_day",
               class: "form-control" %>
         </div>
@@ -92,13 +93,14 @@
         <div class="year">
           <label for="<%= prefix %>_created_range_end_year" class="visually-hidden">Created range end year</label>
           <%= date_range_end_year %>
-          <div class="invalid-feedback">Created year must be between <%= min_year %> and <%= max_year %></div>
+          <div class="invalid-feedback" data-date-validation-target="error"></div>
+          <div class="invalid-feedback" data-date-range-target="endError"></div>
         </div>
         <div class="month">
           <label for="<%= prefix %>_created_range_end_month" class="visually-hidden">Created range end month</label>
           <%= select_month created_range_end_month,
               { prefix: prefix, field_name: 'created_range(5i)', prompt: 'month' },
-              data: { date_validation_target: 'month', action: 'date-validation#change' },
+              data: { date_validation_target: 'month', date_range_target: 'endMonth', action: 'date-range#change' },
               id: "#{prefix}_created_range_end_month", class: "form-control" %>
         </div>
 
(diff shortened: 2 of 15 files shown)


app/components/works/dates_component.html.erb
```diff
@@ -26,7 +26,7 @@
             id: "#{prefix}_created_year",
             placeholder: 'year',
             class: "form-control", min: min_year, max: max_year %>
-        <div class="invalid-feedback">Created year must be between <%= min_year %> and <%= max_year %></div>
+        <div class="invalid-feedback" data-date-validation-target="error"></div>
```
justinlittman, line 29: This is now set by the stimulus controller.

app/components/works/dates_component.html.erb
```diff
@@ -60,21 +60,22 @@
         <div class="year">
           <label for="<%= prefix %>_created_range_start_year" class="visually-hidden">Created range start year</label>
           <%= date_range_start_year %>
-          <div class="invalid-feedback">Created year must be between <%= min_year %> and <%= max_year %></div>
+          <div class="invalid-feedback" data-date-validation-target="error"></div>
+          <div class="invalid-feedback" data-date-range-target="startError"></div>
```
justinlittman, line 64: Each controller has its own place to report errors to avoid stepping on each other.

app/components/works/dates_component.rb
```diff
@@ -16,8 +16,8 @@ def date_range_start_year
       number_field_tag "#{prefix}[created_range(1i)]", created_range_start_year,
                        data: {
                          date_validation_target: 'year',
-                         date_range_target: 'year',
-                         action: 'date-validation#change date-range#change'
+                         date_range_target: 'startYear',
+                         action: 'date-range#change clear->date-validation#clear validate->date-validation#validate'
```
justinlittman, line 20: Notice the binding of the Date Validation controller to the events that will be emitted by the Date Range controller.

app/components/works/dates_component.rb
```diff
@@ -57,15 +57,15 @@ def created_approximate?
     delegate :published_edtf, to: :reform
 
     def created_range_start_year
-      created_range_start&.year
+      created_range_start&.year || reform.send(:'created_range(1i)').presence&.to_i
```
justinlittman, line 60: So that the original submitted values are returned when an EDTF couldn't be created.

app/forms/edtf_params_filter.rb
```diff
@@ -21,10 +20,10 @@ def deserialize(params, name, range)
   end
 
   def deserialize_edtf(params, date_attribute, offset = 0)
-    year = params.delete("#{date_attribute}(#{1 + offset}i)")
-    month = params.delete("#{date_attribute}(#{2 + offset}i)")
-    day = params.delete("#{date_attribute}(#{3 + offset}i)")
-    uncertain = params.delete("#{date_attribute}(approx#{offset})")
+    year = params["#{date_attribute}(#{1 + offset}i)"]
```
justinlittman, line 23: May need to return, so don't delete them.

app/components/works/buttons_component.rb
```diff
@@ -33,5 +33,10 @@ def model
     def work_in_reviewed_coll?
       collection.review_enabled?
     end
+
+    sig { returns(T::Boolean) }
+    def version_draft?
+      work_version.version_draft? && work_version.persisted?
```
jcoyne, line 39: Can you explain why this is necessary?

app/components/works/dates_component.rb
jcoyne: Can you add that as a comment in the code?

app/packs/controllers/date_range_controller.js
```diff
@@ -1,11 +1,107 @@
...
+    }
+
+    validate() {
+        this.startYearTarget.dispatchEvent(new Event('validate'))
+        this.endYearTarget.dispatchEvent(new Event('validate'))
+        const filled = this.dateFilled()
+        if(filled) this.dateOrder()
+    }
+
+    clear() {
```
jcoyne, line 23: Can this have a more descriptive name?  `clearValidationMessages()` maybe?

app/packs/controllers/date_validation_controller.js
```diff
@@ -1,71 +1,82 @@
...
+  }
 
-    if (this.errors.day) {
-      this.dayTarget.classList.add('is-invalid')
-      this.dayTarget.setCustomValidity(this.errors.day)
-    } else {
-      this.dayTarget.classList.remove('is-invalid')
-      this.dayTarget.setCustomValidity('')
-    }
+  clear() {
```
jcoyne, line 20: Can this have a more descriptive name?  It is only the validations being cleared, so maybe `clearValiations`?

app/packs/controllers/date_validation_controller.js
```diff
@@ -1,71 +1,82 @@
...
+    this.yearTarget.required = false
+    this.monthTarget.classList.remove('is-invalid')
+    this.monthTarget.setCustomValidity('')
+    this.monthTarget.required = false
+    this.dayTarget.classList.remove('is-invalid')
+    this.dayTarget.setCustomValidity('')
+    this.errorTarget.textContent = ''
   }
 
-  dateInPast() {
```
jcoyne, line 39: Was dateInPast removed from the front end validations? Why?

jcoyne (changes requested):

app/components/works/buttons_component.rb
justinlittman: Since it hasn't been saved (because an error was returned on form submission), the version draft cannot be discarded. The cancel button is still available for cancelling.

app/components/works/dates_component.rb
justinlittman: Added

app/packs/controllers/date_range_controller.js
justinlittman: Done.

app/packs/controllers/date_validation_controller.js
justinlittman: Done

app/components/works/buttons_component.rb
jcoyne: Would you be able to rename this as `display_cancel_button?` I think that would make it more clear why we care about the persistence.

app/components/works/detail_component.html.erb
```diff
@@ -28,7 +28,7 @@
         <th scope="row">Persistent link</th>
         <td>
           <% if purl %>
-            <%= link_to purl, purl %>
+            <%= link_to purl, purl %> <%= render Works::ShareComponent.new(link: purl) %>
```
jcoyne, line 31: Did this leak from another PR?

app/components/works/share_component.rb
```diff
@@ -0,0 +1,17 @@
+# typed: false
+# frozen_string_literal: true
+
+module Works
+  # Allows the user to share a link.
+  class ShareComponent < ApplicationComponent
```
jcoyne, line 6: Part of a different PR?

app/packs/controllers/share_controller.js
```diff
@@ -0,0 +1,17 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
```
jcoyne, line 3: Did this leak from a different PR?

app/packs/controllers/share_popover_controller.js
```diff
@@ -0,0 +1,19 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
```
jcoyne, line 3: Leaking from a different PR.

jcoyne (changes requested):

app/components/works/detail_component.html.erb
justinlittman: Fixed.

app/components/works/share_component.rb
justinlittman: Yup

app/packs/controllers/date_validation_controller.js
justinlittman: No, the validation was generalized to do some additional checking about the date range so that function name was updated to reflect that.

app/packs/controllers/share_controller.js
justinlittman: Gone.

app/packs/controllers/share_popover_controller.js
justinlittman: Gone

jcoyne (approved):

PR #1662, opened 2021-06-17
Makes title and related works into text areas.

### LINKED ISSUE #1032: Set title, default citation, and related published works fields to text area
Make the title, default citation, and related published works fields set to text area

Design Doc: https://docs.google.com/document/d/1fcr2DYo7OrX-qTdeUOTWKSS1rlq_47sehESmbvgUsrk/edit#heading=h.m9amxw2cnrun

justinlittman:
closes #1032

## Why was this change made?
Better data entry.

Local

NA

Note that default citation was already a text area.

### DIFF
--- a/app/components/works/title_component.html.erb
+++ b/app/components/works/title_component.html.erb
@@ -6,9 +6,13 @@
        <%= render PopoverComponent.new key: 'work.title' %>
      </div>
      <div class="col-sm-9">
-       <%= form.text_field :title, class: "form-control", required: true,
-                           data: { action: "change->auto-citation#updateDisplay",
-                                   auto_citation_target: 'titleField' } %>
+       <%= form.text_area :title, class: "form-control", required: true, rows: 1,
+                           data: { action: "no-newlines#change change->auto-citation#updateDisplay",
+                                   auto_citation_target: 'titleField',
+                                   no_newlines_target: 'input',
+                                   controller: 'no-newlines'
+
+                           } %>
        <div class="invalid-feedback">You must provide a title</div>
      </div>
    </div>
--- a/app/packs/controllers/no_newlines_controller.js
+++ b/app/packs/controllers/no_newlines_controller.js
@@ -1,0 +1,11 @@
+import { Controller } from "stimulus"
+
+// This controller removes newlines from an input such as a textarea.
+// This allows the input to wrap long text.
+export default class extends Controller {
+  static targets = ["input"]
+
+  change() {
+    this.inputTarget.value = this.inputTarget.value.replace( /\r?\n/gi, '' )
+  }
+}
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -57,7 +57,8 @@
       <div class="col-sm-10">
         <%= form.text_area :citation, class: "form-control",
                            aria: { label: 'Provided citation' },
-                           data: { auto_citation_target: 'manual' } %>
+                           data: { auto_citation_target: 'manual',
+                                   action: "no-newlines#change", no_newlines_target: 'input', controller: 'no-newlines' } %>
         <%= form.text_area :citation_auto,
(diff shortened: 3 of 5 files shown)


justinlittman:
image

justinlittman:
image

app/components/works/title_component.html.erb
```diff
@@ -6,7 +6,7 @@
        <%= render PopoverComponent.new key: 'work.title' %>
      </div>
      <div class="col-sm-9">
-       <%= form.text_field :title, class: "form-control", required: true,
+       <%= form.text_area :title, class: "form-control", required: true, rows: 1,
```
jcoyne, line 9: @amyehodge @astridu @andrewjbtw Can we handle work titles that have linebreaks in them?  I'm concerned not just with this system, but how these show up in argo/purl/sul_embed/searchworks.

jcoyne (changes requested):

app/components/works/title_component.html.erb
amyehodge: I don't think we should have titles with line breaks.

jcoyne: @amyehodge do you care to revise https://github.com/sul-dlss/happy-heron/issues/1032#issuecomment-830170689 then?  A textarea element is for multi-line input.

amyehodge: The request was changed to text area because that was easier to accomplish, but I'm happy to change this. What else should there be? I'm not familiar with what the options are or how they work specifically, so I didn't know textarea would allow multi-line, for example.

For these three fields on this ticket -- title, default citation, and related works -- the user should be able to expand the typing area so they can see the full entry at one time, but they should not be able to add line breaks. What request meets those parameters?

jcoyne: I'm not sure there is a way to accommodate this in a web form.  Have you seen another form like this somewhere on the web that we could look to for inspiration?

amyehodge: I don't have an example right off hand. Can we just make these boxes bigger to accommodate, say, 3 lines of display. And not have them be expandable? @astridu do you have any other ideas?

jcoyne: Setting aside implementation details, one of the problems here is that when we show multiple lines to the user, the user will have the expectation that they can enter in line breaks because this is how every form they have ever encountered has worked. So we would need to come up with a design that wouldn't frustrate them as our control wouldn't work the same way.

jcoyne: The comment boxes on https://stackoverflow.com, prevent you from entering line-breaks, but I'm not sure this is a good/complete solution, because you can still paste in text that has line breaks in it.

jcoyne: @justinlittman okay, maybe you can add something like this:
```
inputField.addEventListener('input', (e) => {
  inputField.value = inputField.value.replace( /\r?\n/gi, '' )
})
```

astridu: Any of these text fields in github describe the behavior Amy suggests. If you drag the 2 diagonal stripes in the bottom right corner, you can expand the field. 
You can even add a line break by clicking shift + enter.

amyehodge: @astridu We don't want them to be able to put line breaks in these particular fields.

astridu: I don't know if there's a way to prohibit line breaks.

app/packs/controllers/no_newlines_controller.js
```diff
@@ -0,0 +1,9 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
```
jcoyne, line 3: How about adding a comment about what this controller does.  Then feel free to merge.

jcoyne (approved):

PR #1673, opened 2021-06-18
Use block format for label with html

jcoyne:


jcoyne:
Can't be done this way due to https://github.com/github/view_component/pull/240

PR #1709, opened 2021-06-24
Disable reviewers when review workflow turned off.

### LINKED ISSUE #1702: Hide reviewers field on full collection edit and collection settings forms when review workflow is off
This box currently looks like this:

If the review workflow is off and the collection manager enters a SUNet in the box, it is not saved and there is no warning to the user that it isn't being saved. Entries should only be needed here if the workflow is on, so I would suggest that we do not display the text "Enter SUNet IDs of participants. Separate each one with a comma and a space. (e.g. janelath, lelandst)" or the entry box and label unless the review workflow option has been toggled to on.

justinlittman:
closes #1702.

## Why was this change made?
Confusing for user to be able to enter reviewers when review workflow turned off.

Local, unit

NA

### DIFF
--- a/app/components/first_draft_collections/create/settings_component.html.erb
+++ b/app/components/first_draft_collections/create/settings_component.html.erb
@@ -97,14 +97,14 @@
     </div>
   </section>
 
-  <section id="review">
+  <section id="review" data-controller="review-workflow">
     <header>Review Workflow</header>
 
     <div class="mb-3 row">
       <%= tag.span t('hints.collection.review_enabled'), class: 'col-sm-10 col-xl-8 col-form-label' %>
       <div class="col-sm-3">
         <div class="form-check form-switch">
-          <%= form.check_box :review_enabled, { class: 'form-check-input' }, 'true', 'false' %>
+          <%= form.check_box :review_enabled, { class: 'form-check-input', data: { action: 'review-workflow#toggle', review_workflow_target: 'enabled' } }, 'true', 'false' %>
           <%= form.label :review_enabled, 'Enable Review Workflow', class: "form-check-label" %>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


app/packs/controllers/review_workflow_controller.js
```diff
@@ -0,0 +1,17 @@
...
+export default class extends Controller {
+  static targets = [ "enabled", "reviewers" ]
+
+  connect() {
+    this.toggle()
+  }
+
+  toggle() {
+    if(this.enabledTarget.checked) {
+      this.reviewersTarget.removeAttribute('disabled')
```
jcoyne, line 12: This can be reduced to `this.reviewersTarget.disabled = !this.enabledTarget.checked`

jcoyne (changes requested):

app/components/first_draft_collections/create/settings_component.html.erb
```diff
@@ -97,14 +97,14 @@
...
-  <section id="review">
+  <section id="review" data-controller="review-workflow">
     <header>Review Workflow</header>
 
     <div class="mb-3 row">
       <%= tag.span t('hints.collection.review_enabled'), class: 'col-sm-10 col-xl-8 col-form-label' %>
       <div class="col-sm-3">
         <div class="form-check form-switch">
-          <%= form.check_box :review_enabled, { class: 'form-check-input' }, 'true', 'false' %>
+          <%= form.check_box :review_enabled, { class: 'form-check-input', data: {action: 'review-workflow#toggle', review_workflow_target: 'enabled'} }, 'true', 'false' %>
```
jcoyne, line 107: please add spaces inside of braces.

app/packs/controllers/review_workflow_controller.js
```diff
@@ -0,0 +1,17 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = [ "enabled", "reviewers" ]
+
+  connect() {
+    this.toggle()
+  }
+
+  toggle() {
```
jcoyne, line 10: Is there a better name?  Toggle to me means flip the state, but this is really just synchronizing with the UI.

app/components/first_draft_collections/create/settings_component.html.erb
justinlittman: I just assumed that rubocop fixes _everything_.

app/packs/controllers/review_workflow_controller.js
justinlittman: It is toggling disabled. But I'm open to other suggestions.

app/packs/controllers/review_workflow_controller.js
justinlittman: Huh. I thought the attribute actually had to be removed. But verified that simpler statement works.

Thanks!

jcoyne (approved):

PR #1714, opened 2021-06-24
Allow new-sdr for action cable

jcoyne:
## Why was this change made?
allow WS on prod


justinlittman (approved):
Approving, but unclear after changes made by ops it is necessary, so leaving unmerged.

jcoyne:
@justinlittman I think it is, because this was deployed to prod.

PR #1769, opened 2021-07-08
Create a DOI service

jcoyne:
Blocked by:
* a testing prefix

## Why was this change made?

This shows that we can interact with the Datacite DOI minting service.


jcoyne:
This code moved to DSA.

PR #1783, opened 2021-07-12
Switch datatables to disable search via config. Enable search for works.

### LINKED ISSUE #1742: Restore search box on collection deposits page
I'm pretty sure there used to be a search box on the right side above the table on the collection deposits page. Here's what I see at https://sul-h2-stage.stanford.edu/collections/1084/works:

justinlittman:
closes #1742

## Why was this change made?
Per ticket.

Local

NA


justinlittman:
image

jcoyne (approved):

PR #1788, opened 2021-07-12
Adds public domain mark.

### LINKED ISSUE #1747: Add Public Domain Mark to H2 license options
There is at least one staff-deposited collection (government documents) that uses the Public Domain Mark. This option cannot be selected in H2 or Hydrus, but I've learned that it is assigned within Argo via special APO settings. 
(see the APO here: https://argo.stanford.edu/view/druid:yf511vz3216 )

Since H2 does not use special APOs for each collection, this has made the government documents workflow more complex as new H2-deposited objects are not associated with a special APO. 

If this license could be selected in H2, then it would not need to be assigned in Argo and would simplify things considerably. I believe PDM is the only license option available in Argo that is not available in H2. Adding this option would mean we could have parity in the license menus once we add the H2 options to Argo.

justinlittman:
closes #1747

## Why was this change made?
Per ticket

Locally

### DIFF
--- a/app/models/license.rb
+++ b/app/models/license.rb
@@ -97,8 +97,11 @@ class License
       ]
     },
     {
-      label: 'No License',
-      options: ['none']
+      label: 'Other',
+      options: [
+        'none',
+        'PDM-1.0'
+      ]
     }
   ].freeze
 end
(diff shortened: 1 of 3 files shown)


justinlittman:
image

jcoyne (approved):

PR #1791, opened 2021-07-13
Close modals before turbo caches.

### LINKED ISSUE #1357: Prevent work type modal from showing up when you discard a draft
When I start a draft and then decide to navigate away from the page (in this case, back to the collection view page) and discard my current work, the work type modal flashes up during the page transition.

justinlittman:
closes #1357

## Why was this change made?
Turbo was caching with an open modal causing the modal to flash on page changes.

Local

NA

### DIFF
--- a/app/packs/entrypoints/application.js
+++ b/app/packs/entrypoints/application.js
@@ -27,3 +27,10 @@ import 'simple-datatables'
 // const imagePath = (name) => images(name, true)
 
 require('modules/validate-forms')
+
+window.addEventListener("turbolinks:before-cache", function() {
+    // Close modal window
+    document.querySelectorAll('div.modal').forEach(function (elem) {
+        bootstrap.Modal.getInstance(elem).hide()
+    })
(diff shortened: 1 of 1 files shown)


jcoyne (approved):

app/packs/entrypoints/application.js
```diff
@@ -27,3 +27,10 @@ import 'simple-datatables'
 // const imagePath = (name) => images(name, true)
 
 require('modules/validate-forms')
+
+window.addEventListener("turbolinks:before-cache", function() {
```
jcoyne, line 31: We use turbo, not turbolinks 😬 I missed that the first time.

jcoyne (changes requested):

app/packs/entrypoints/application.js
justinlittman: https://github.com/sul-dlss/happy-heron/pull/1796

PR #1812, opened 2021-07-15
Stop using publication year to construct citation.

### LINKED ISSUE #1801: Update rules for default citation date so that publication date is no longer used.
According to #655, we set up the rules for the date in the automatically-generated citation like this:

"Rules for the date:

- If there is a publication date entered, use that date, but only the year
- If there is no publication date but there is an embargo date, use that date, but the year only
- Otherwise use the current year" 

After some discussions recently with @arcadiafalcone, we realized that if we allow the publication date to be used, it will be likely that it won't match the publisher properly. That is, if something was published in 1972 and we use that date but *always* use SDR as the publisher, this doesn't make sense. It makes more sense to always use the embargo release date or deposit date (the current year), which would correspond to the publication of this digital copy of the object by the SDR. If the user wants to change the date and publisher, they have the ability to do so.

Therefore, we should take out the rule to use the publication date and the rules should be:

- If there is an embargo date, use that date, but only the year
- Otherwise use the current year

justinlittman:
closes #1801

## Why was this change made?
Per ticket.

Unit

NA

### DIFF
--- a/app/packs/controllers/auto_citation_controller.js
+++ b/app/packs/controllers/auto_citation_controller.js
@@ -3,7 +3,7 @@ import { Controller } from "stimulus";
 export default class extends Controller {
   static targets = ["titleField", "manual", "auto", "switch",
     "contributorFirst", "contributorLast", "contributorRole", "contributorOrg",
-    "year", "embargoYear", "embargo"];
+    "embargoYear", "embargo"];
 
   connect() {
     this.purl = this.data.get("purl") || ":link will be inserted here automatically when available:" // Use a real purl on a persisted item or a placeholder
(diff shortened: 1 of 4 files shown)


app/packs/controllers/auto_citation_controller.js
```diff
@@ -72,17 +72,17 @@ export default class extends Controller {
...
-    return this.yearTarget.value
+    return null
   }
 
   get date() {
     const date = new Date();
-    if (this.year) {
-      date.setYear(this.year)
+    if (this.embargoYear) {
+      date.setYear(this.embargoYear)
```
jcoyne, line 85: Should it be set to current year if the embargo date is not present?

jcoyne (changes requested):

app/packs/controllers/auto_citation_controller.js
justinlittman: It is when `date` is initialized.

jcoyne (approved):

PR #1819, opened 2021-07-15
Updates item deposit form with new DOI assignment section.

### LINKED ISSUE #1728: DOI: Update item deposit form with new DOI assignment section
This will allow the user to either make a selection with regard to DOI minting or see what the collection manager has chosen -- to assign a DOI or not.

Design Doc: https://docs.google.com/document/d/16zGrIpRgGkq8d4VDUjEstoB2p8xFEuiAmKJopcT1CaQ/edit#heading=h.9z9lyuj5vyhc

Mockups: https://docs.google.com/document/d/16zGrIpRgGkq8d4VDUjEstoB2p8xFEuiAmKJopcT1CaQ/edit#heading=h.lr01ej7okb5u -- in particular mockup II.

justinlittman:
closes #1728

## Why was this change made?
Per ticket

Unit

NA

### DIFF
--- a/db/migrate/20210715161256_change_add_doi_default.rb
+++ b/db/migrate/20210715161256_change_add_doi_default.rb
@@ -1,0 +1,5 @@
+class ChangeAddDoiDefault < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default :works, :assign_doi, true
+  end
+end
--- a/app/components/works/doi_component.html.erb
+++ b/app/components/works/doi_component.html.erb
@@ -1,0 +1,31 @@
+<fieldset class="mb-5">
+  <legend class="h5">
+    DOI assignment
+
+  </legend>
+  <%= link_to 'What is a DOI?', 'https://library.stanford.edu/data-services/doi-services' %>
+  <div class="mb-3 inner-container">
+    <% if doi %>
+      <p>The DOI assigned to this work is <%= link_to doi, doi %>.</p>
+    <% elsif collection_doi_option == 'yes' %>
+      <p>A DOI will be assigned.</p>
+    <% elsif collection_doi_option == 'no' %>
+      <p>A DOI will not be assigned.</p>
+    <% else %>
+      <div class="row">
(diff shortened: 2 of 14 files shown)


justinlittman:
image

db/migrate/20210715161256_change_add_doi_default.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeAddDoiDefault < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default :works, :assign_doi, true
```
jcoyne, line 3: Is this required? I'm concerned that this will affect all existing works.  We don't want this value to be true for works that were create without DOIs

jcoyne (changes requested):

db/migrate/20210715161256_change_add_doi_default.rb
justinlittman: Yes, it is required. The default is supposed to be true when depositor selects. Note that this is combined with the collection DOI option to determine whether a DOI is assigned. (That is, it is ignored when DOI option is yes or no.)

jcoyne: @justinlittman Can we move the default off of the model and onto the form?

jcoyne: Hmm, probably not since its a boolean.  Can/should we set the existing works to "DOI requested" or should they be "DOI refused"?

justinlittman: Existing works would be "no", which makes sense since the user did not request one or have the option of choosing. However, for future works the default would be "yes", but the user would have the option of refusing.

jcoyne (approved):

db/migrate/20210715161256_change_add_doi_default.rb
jcoyne: @justinlittman what I'm saying here is that by changing this default, it changes all of the existing to "yes"

jcoyne: Oh, actually, the other migration runs first, which sets them to "false", and only then is the default chagned, so this should be fine.

PR #1824, opened 2021-07-16
Set a DOI when requested on subsequent versions

jcoyne:

### DIFF
--- a/app/services/request_generator.rb
+++ b/app/services/request_generator.rb
@@ -52,8 +52,18 @@ class RequestGenerator
   def identification
     {
       sourceId: "hydrus:object-#{work.id}",
-      doi: work.doi
+      doi: doi
     }.compact
+  end
+
+  sig { returns(T.nilable(String)) }
+  # If the work already has a DOI, just return it.
+  # If the user decides they want a DOI minted on a new version, we need to set it here.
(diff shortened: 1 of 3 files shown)


app/services/request_generator.rb
```diff
@@ -52,10 +52,20 @@ def model_attributes
...
+      doi: doi
     }.compact
   end
 
+  sig { returns(T.nilable(String)) }
+  # If the work already has a DOI, just return it.
+  # If the user decides they want a DOI minted on a new version, we need to set it here.
+  def doi
+    return work.doi if work.doi
+    return if work_version.first_draft? || !work.assign_doi?
```
justinlittman, line 64: The logic should be to not assign DOI if collection DOI option is "no" or collection DOI option is "depositor-selects" and assign DOI is false.

justinlittman (changes requested):

app/services/request_generator.rb
jcoyne: @justinlittman isn't that logic encapsulated in `work.assign_doi?`

justinlittman: Yes.

justinlittman (approved):

PR #1838, opened 2021-07-19
store last time terms of agreement for a work were accepted; and use that for default on new works

### LINKED ISSUE #243: Record terms of deposit acceptance time in database at user level
once recorded, ToD checkbox on work edit form renders as already checked if its been less than 1 year since they last checked it manually

peetucket:
## Why was this change made?

Fixes #243 - record the last time a user accepted the terms of agreement for a work -- when a new work is created, default to a checked state and show a message instead of a checkbox if its been less than 1 year since they accepted, else default to a false state, forcing them to check again (which will then update the last time it was checked).

* use new timestamp column on the user model to record when terms agreement occurred
* remove column on work_version that used to store terms agreement
* if the time has been less than 1 year, show a message to the user on the work form that the agreement was accepted already (instead of showing a checkbox and using the currently persisted value from the work_version model)
* if the time has been more than 1 year (or never recorded), show an unchecked checkbox each time the work draft page is rendered
* if the user checks the checkbox when saving the work (draft or deposit, doesn't matter), reset the timestamp to the current time

Localhost browser, updating/add request specs


spec/requests/edit_work_spec.rb
```diff
@@ -81,7 +81,7 @@
           expect(work.reload.head).to be_version_draft
           # Only changed fields are recorded in event.
           expect(work.events.first.description).to eq('title of deposit modified, contact email modified, ' \
-            'authors modified, file description changed')
+                                                      'authors modified, file description changed')
```
peetucket, line 84: unrelated rubocop auto-correct

app/forms/draft_work_form.rb
```diff
@@ -158,6 +158,19 @@ def description
...
   end
 
+  # the terms agreement default to saved value for this object if persisted,
+  #  or false if (a) never accepted or (b) not accepted in the last year,
+  #  else it defaults to true
+  #  (i.e a new draft with a previously agreed work in the last year defaults to true)
+  def default_agree_to_terms
+    return model.fetch(:work_version).agree_to_terms if persisted?
+
+    last_work_terms_agreement = model.fetch(:work).collection.creator.last_work_terms_agreement
```
jcoyne, line 168: why look at the collection creator and not the work depositor?

peetucket: yes, and in fact this is wrong because the collection creator could be different than the work depositor...

the problem I am running into is that that the work depositor does not appear to be set in the work model for a brand new work until it is persisted:

```
(byebug) model.fetch(:work_version).work
#<Work id: nil, druid: nil, head_id: nil, collection_id: 1, depositor_id: nil, created_at: nil, updated_at: nil>
```

I also don't appear to have access to the current logged in user (`current_user`) in this class ... any thoughts on best way to get the current access to the current depositor from here?

jcoyne: You can do like this:
https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/works_controller.rb#L23

on this line:
https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/works_controller.rb#L14

app/components/works/agreement_component.html.erb
```diff
@@ -2,11 +2,16 @@
...
+        You agreed to the <%= link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'} %>
+        <%= time_ago_in_words form.object.model.fetch(:work).depositor.last_work_terms_agreement %> ago.
+      <% else %>
+        <%= form.check_box :agree_to_terms, required: true, class: 'form-check-input' %>
+        <%= form.label :agree_to_terms,
+        "I agree to the #{link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'}}".html_safe,
+        class: 'form-check-label' %>
+        <div class="invalid-feedback">You must agree to the Terms of Deposit.</div>
+      <% end %>
     </div>
```
peetucket, line 15: Selectively show checkbox based on if the user has previously accepted the agreements in the specified timeframe.

app/controllers/works_controller.rb
```diff
@@ -11,7 +11,7 @@ class WorksController < ObjectsController
   def new
     validate_work_types!
     collection = Collection.find(params[:collection_id])
-    work = Work.new(collection: collection)
+    work = Work.new(collection: collection, depositor: current_user)
     work_version = WorkVersion.new(work_type: params[:work_type], subtype: params[:subtype], work: work)
```
peetucket, line 15: we need this to ensure we have access to the depositor on the work form

app/models/user.rb
```diff
@@ -28,6 +28,12 @@ class User < ApplicationRecord
 
   devise :remote_user_authenticatable
 
+  # this is the timeframe after which the user must agree to the terms again on a work
+  sig { returns(ActiveSupport::TimeWithZone) }
+  def terms_agreement_renewal_timeframe
+    Time.zone.now.years_ago(1)
+  end
+
```
peetucket, line 36: specifies the timeframe after which the user must agree to the terms again

app/models/user.rb
```diff
@@ -42,4 +48,16 @@ def sunetid
...
+    return false unless last_work_terms_agreement
+
+    terms_agreement_renewal_timeframe < last_work_terms_agreement
+  end
+
+  sig { returns(T::Boolean) }
+  def agreed_to_terms
+    update(last_work_terms_agreement: Time.zone.now)
+  end
 end
```
peetucket, line 63: when the user agrees to the terms, we need to update the current time in the database

spec/components/dashboard/deposit_progress_component_spec.rb
```diff
@@ -27,7 +27,7 @@
 
   context 'when work has everything' do
     let(:work_version) do
-      build_stubbed(:valid_work_version, agree_to_terms: true, attached_files: [build_stubbed(:attached_file)])
+      build_stubbed(:valid_work_version, attached_files: [build_stubbed(:attached_file)])
```
peetucket, line 30: this column no longer exists in the database for the work_version

peetucket:
Note that the `create_new_work_spec.rb` and `edit_draft_work_spec.rb` are not as different as they are shown.  Basically: (1) All of the existing tests were wrapped in a new context that assumes the terms were already previously accepted and (2) new context blocks were added for the cases where the user had never previously accepted terms and accepted them more than 1 year ago, and then some of the same tests were repeated, but checking the "terms" box this time.

app/controllers/validates_controller.rb
```diff
@@ -9,7 +9,7 @@ def show
       work = Work.find(params[:work_id])
       work_version = work.head
     else
-      work = Work.new(collection_id: params[:collection_id])
+      work = Work.new(collection_id: params[:collection_id], depositor: User.new)
```
peetucket, line 12: the new `agree_to_terms` assumes there is an associated depositor, we need one for a new work

ndushay:
@peetucket this needs a rebase.  @jcoyne @amyehodge what is the review status of this PR?

peetucket:
Rebased

jcoyne:
@ndushay I guess I'm confused what you are asking me. It doesn't look like anyone has reviewed this.

ndushay:
jcoyne - would you like to review the code?   I *think* Amy reviewed this before.

peetucket:
A couple screenshots to show the change in the UI...

This is what it looks like on the work deposit page when you have never accepted the terms of agreement or its been more than one year since you've accepted.  There is no change:

Screen Shot 2021-08-04 at 3 33 12 PM

This is what it looks like on the work deposit page when it's been less than a year since you accepted the terms of agreement on any other work.  The "time ago" phrase is a friendly representation (e.g. "a month ago").  The Terms are still linked to the popup for viewing/printing but cannot be "unaccepted".

Screen Shot 2021-08-04 at 3 34 19 PM

jcoyne:
@peetucket Can you throw this on qa or stage for @amyehodge  to test?

peetucket:
Yup, I can deploy to stage, but there is one wrinkle in there is a migration that removes a database column, making it a bit more work to revert back to main.  You can't just deploy main because the column will be gone from the database, or at least I had this happen to me on localhost.  

Obviously if we merge after she reviews it there, it's not a problem.


I can also walk through with Amy on a screenshare if that makes it easier.

amyehodge:
@peetucket we can do a walkthrough on screen share. I don't like the sound of the other option.

app/components/works/agreement_component.html.erb
```diff
@@ -2,11 +2,16 @@
   <header>Terms of Deposit *</header>
   <div class="row">
     <div class="form-check col-auto">
-      <%= form.check_box :agree_to_terms, required: true, class: 'form-check-input' %>
-      <%= form.label :agree_to_terms,
-       "I agree to the #{link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'}}".html_safe,
-       class: 'form-check-label' %>
-       <div class="invalid-feedback">You must agree to the Terms of Deposit.</div>
+      <% if form.object.agree_to_terms %>
```
jcoyne, line 5: Can you make this into a method on the `AgreementComponent` so that it's just `agreed_to_terms?` here?

app/models/work_version.rb
```diff
@@ -124,6 +124,21 @@ def created_edtf=(edtf)
...
+    work.depositor.agreed_to_terms_recently?
+  end
+
+  # the terms agreement checkbox value is not persisted in the database with the work but instead at the user level
+  def agree_to_terms=(value)
+    return if value == false || value == '0' || work.depositor.agreed_to_terms_recently?
+
+    # update the last time the terms of agreement was accepted for this depositor
+    #  if it has not been accepted within the defined timeframe and the checkbox was checked
+    work.depositor.agreed_to_terms
```
jcoyne, line 139: Can we put the update here?  Otherwise it's unclear that this is mutating and persisting.

jcoyne (changes requested):
I have a few small changes that I think would help readability, otherwise looks good.

app/models/work_version.rb
peetucket: ok, removed the method on the depositor and just issued the update query here

app/components/works/agreement_component.html.erb
peetucket: add a helper method on the agreement component to make this more readable

peetucket:
I reviewed with Amy over screenshare and she approved the changes.

jcoyne (approved):

PR #1839, opened 2021-07-21
move grouped dates on work edit page into fieldsets

### LINKED ISSUE #1827: Accessibility: date fields need to be grouped and labeled
When a form contains elements that are logically connected to each other it is necessary to group them.

The form contains check boxes and/or radio buttons that are not grouped and named, using a 'fieldset' with a descriptive 'legend' or the WAI-ARIA attribute role="group" (for check boxes) or role="radiogroup" (for radio buttons) with a descriptive 'aria-labelledby' or 'aria-label'.

On the deposit page, these radio button form elements need to be grouped:
Creation date
image

Release date
image

peetucket:
## Why was this change made?

Fixes #1827 - move input controls that are related to the same field into a single fieldset (just wrap the existing div) and ensure they have a legend.

Localhost browser

### DIFF
--- a/app/components/works/available_date_component.html.erb
+++ b/app/components/works/available_date_component.html.erb
@@ -1,40 +1,43 @@
-<div data-controller="complex-radio available-date">
-  <div class="mb-3 row">
-    <div class="col-sm-10 release-option" data-complex-radio-target="selection">
-      <%= form.radio_button :release, 'immediate', class: 'form-check-input',
-                            data: {
-                              action: 'complex-radio#disableUnselectedInputs change->auto-citation#updateDisplay available-date#clearError'
-                            } %>
-      <%= form.label :release_immediate, 'Immediately once deposit is completed' %>
-    </div>
-  </div>
-
-  <div class="embargo-date mb-5">
-    <div class="row<%= ' is-invalid' if error? %>">
+<fieldset>
+  <legend class="h5">Release date</legend>
+  <div data-controller="complex-radio available-date">
+    <div class="mb-3 row">
       <div class="col-sm-10 release-option" data-complex-radio-target="selection">
-        <%= form.radio_button :release, 'embargo', class: 'form-check-input',
+        <%= form.radio_button :release, 'immediate', class: 'form-check-input',
                               data: {
-                                action: 'complex-radio#disableUnselectedInputs change->auto-citation#updateDisplay available-date#validate',
-                                auto_citation_target: 'embargo'
+                                action: 'complex-radio#disableUnselectedInputs change->auto-citation#updateDisplay available-date#clearError'
                               } %>
-        <%= form.label :release_embargo, 'On this date' %>
-
-        <div class="year">
-          <label for="work_embargo_year" class="visually-hidden">Embargo year</label>
-          <%= year_field %>
-        </div>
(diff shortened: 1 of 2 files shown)


peetucket:
This diff is really odd, since all I did was wrap the existing divs in a `<fieldset>` and add a legend.

app/components/works/available_date_component.html.erb
```diff
@@ -1,40 +1,43 @@
...
-                              action: 'complex-radio#disableUnselectedInputs change->auto-citation#updateDisplay available-date#clearError'
-                            } %>
-      <%= form.label :release_immediate, 'Immediately once deposit is completed' %>
-    </div>
-  </div>
-
-  <div class="embargo-date mb-5">
-    <div class="row<%= ' is-invalid' if error? %>">
+<fieldset>
+  <legend class="h5">Release date</legend>
```
jcoyne, line 2: Is this supposed to be hidden? I don't see it in the mockups.  Can you add screenshots that shows the diff?

jcoyne (changes requested):

app/components/works/dates_component.html.erb
```diff
@@ -6,117 +6,116 @@
...
-            class: "form-control", min: min_year, max: max_year %>
-        <div class="invalid-feedback" data-date-validation-target="error"></div>
-      </div>
-      <div class="month">
-        <label for="<%= prefix %>_created_month" class="visually-hidden">Created month</label>
-        <%= select_month created_month, { prefix: prefix, field_name: 'created(2i)', prompt: 'month'},
-                         data: { date_validation_target: 'month', action: 'date-validation#change' },
-                         id: "#{prefix}_created_month", class: "form-control" %>
-      </div>
+    <legend class="h5">Creation date <%= render PopoverComponent.new key: 'work.creation_date' %></legend>
```
jcoyne, line 11: Only "phrasing content" (https://html.spec.whatwg.org/multipage/dom.html#phrasing-content) is permitted within `<legend>`. I don't think the `PopoverComponent` is permitted.

jcoyne (changes requested):

app/components/works/available_date_component.html.erb
peetucket: indeed...I just hid both legends using a `visually-hidden` class and otherwise left the displays as before...so it now looks exactly the same as it did before (just verified against main branch on localhost)

app/components/works/dates_component.html.erb
```diff
@@ -6,117 +6,123 @@
                                                  min_year: min_year,
                                                  max_year: max_year) %>
 
-  <div class="mb-3 row">
-    <div class="col-sm-3 h6">
-      Creation date
-      <%= render PopoverComponent.new key: 'work.creation_date' %>
-    </div>
-  </div>
+<fieldset>
```
jcoyne, line 9: The indent level on this line does not match line 126. Can you make them the same?

jcoyne (changes requested):

app/components/works/dates_component.html.erb
peetucket: Done

jcoyne (approved):

PR #1860, opened 2021-07-29
Display file upload errors.

### LINKED ISSUE #1577: display of file upload errors
Current behavior for a file upload error:

image

There is no filename showing, and I'm sure this isn't the optimal display.

While working on #1378, I had upload errors looking like this:

image

@amyehodge is okay with the above.

Certainly we will want the filename to display, and I would suspect an error symbol in addition to the error message.   I took the default dropzone behavior and styled it a bit to get the above.

However, when I rebased my work on main, the `.on("error"` method had been added to `dropzone_controller.js`.   I started down this path:

image

and it was suggested by JCoyne that I split out the error handling from the success checkmark work in PR #1576.

For developers:
The way I've been able to force an upload error is to add this code around line 115 in the `dropzone_controller.js` file:

```
    this.dropZone.on("success", file => {
      file.controller.emitDropzoneError('this is a very long upload error asfasfasf asfasfasfsaf safasfsafs afsfsfsafa')
    })
```

And then in `file_row_component.html.erb` you will need to comment out the div with class `dz-success-mark` so you don't have both the success and error displays at the same time.

justinlittman:
closes #1577

## Why was this change made?
Sometimes users go down the unhappy path.

Local

image
image

NA

### DIFF
--- a/app/components/works/file_row_component.html.erb
+++ b/app/components/works/file_row_component.html.erb
@@ -11,10 +11,12 @@
     <% else %>
       <div class="dz-filename" data-dropzone-target="fileName"><span data-dz-name></span></div>
       <div class="row">
-        <div class="dz-progress col-md-2 align-self-start"><span class="dz-upload" data-dz-uploadprogress></span></div>
-        <div class="dz-success-mark col align-self-end"><span class="fas fa-check fa-sm" /></div>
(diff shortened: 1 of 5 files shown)


justinlittman:
BONUS: If you name the file `dropzone_error.txt` it will trigger an error.

justinlittman:
If there is an error for a file, the form will not submit until the user has removed the file.

jcoyne (approved):

PR #1862, opened 2021-07-29
Adds additional links to home page.

### LINKED ISSUE #1858: Add two new links from home page to library website and update order/layout of header links
We would like to insert two links on the home page. Both will go to https://library.stanford.edu/research/stanford-digital-repository.

Here's the mockup:
https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/455998487_home

The additions are:
1. On the top right, add link text "SDR Services". Note that the above mockup has the correct order for the links in this area and there should be dots between the links
2. Below the list under "The process is simple" add a button that says "Learn more about SDR Services"

justinlittman:
closes #1858

## Why was this change made?
Per ticket

Local

image

### DIFF
--- a/app/packs/stylesheets/frontpage.scss
+++ b/app/packs/stylesheets/frontpage.scss
@@ -63,9 +63,12 @@
     input:placeholder-shown {
        font-style: italic;
     }
-
   }
 
+  a.btn {
+    color: $white;
+    border-color: $white;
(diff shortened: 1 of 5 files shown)


ndushay:
oh no - failing build!

ndushay (approved):

PR #1868, opened 2021-07-29
Update general mappings to match dor-services-app

arcadiafalcone:
## Why was this change made?
So H2 produces the cocina that dor-services-app expects.

CircleCI

n/a


arcadiafalcone:
I'm not sure about 6133ac4 - basically I think it needs to be either "no URI" or "URI and FAST as source," as those are the two scenarios in the data.

PR #1869, opened 2021-07-29
Enables and tweaks to types_generator_spec.

justinlittman:
refs https://github.com/sul-dlss/dor-services-app/issues/2869


ndushay (approved):
approving with the caveat of @arcadiafalcone  must approve all these spec changes.  Also wondering about the can be deleted ... is the implication that the cocina_props should NOT be generated?

arcadiafalcone:
@ndushay No, that the cocina produced in these scenarios is covered elsewhere so these specs are redundant.

PR #1871, opened 2021-07-30
Re-revert contributor type changes. 

justinlittman:
## Why was this change made?
Confusion

Unit


justinlittman:
Note that this includes the revert and fixes for the issues that required the first revert.

jcoyne (approved):

PR #1881, opened 2021-08-03
Adds ORCID to cocina model when depositing.

### LINKED ISSUE #1843: Add ORCID information to descriptive metadata
ORCID IDs added to H2 should be stored in descMD.

@arcadiafalcone can you confirm the location where these should be stored? Thanks.

justinlittman:
closes #1843

## Why was this change made?
Per ticket

Unit

NA

### DIFF
--- a/spec/services/cocina_generator/description/contributors_generator_spec.rb
+++ b/spec/services/cocina_generator/description/contributors_generator_spec.rb
@@ -1064,6 +1064,48 @@ RSpec.describe CocinaGenerator::Description::ContributorsGenerator do
       end
     end
 
-    # NOTE: not included in xit: mappings with ORCID
+    context 'with contributor with ORCID' do
+      let(:author) do
+        build(:person_author, first_name: 'Jane', last_name: 'Stanford', orcid: 'https://orcid.org/0000-0000-0000-0000')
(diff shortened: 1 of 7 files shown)


spec/services/cocina_generator/description/contributors_generator_spec.rb
```diff
@@ -1064,6 +1064,48 @@
...
+                type: 'person',
+                status: 'primary',
+                role: [contributing_author_role],
+                identifier: [
+                  {
+                    value: '0000-0000-0000-0000',
+                    type: 'ORCID',
+                    source: {
+                      uri: 'https://orcid.org'
+                    }
```
ndushay, line 1101: This matches https://github.com/sul-dlss/dor-services-app/blob/main/spec/services/cocina/mapping/descriptive/h2/contributor_h2_spec.rb#L1133-L1140  so I think it is correct.

ndushay (commented):
were we going to have another feature branch for orcid, with a feature switch in config?  Or am I misremembering?   Not sure this PR would be implicated in any case.

ndushay (approved):
don't need orcid branch for this PR, only for ORCID UI changes

PR #1883, opened 2021-08-03
Adds display of ORCID to work detail page.

### LINKED ISSUE #1735: ORCID: Update item view page to accommodate display of ORCID information
Documentation and mockups: https://docs.google.com/document/d/1LXsN7yfzZCnBwHkI_L7sOni5mkecJ3t1fiyR_5BT4lQ/edit#

Once this information is entered, it will also be displayed on the item view page next to the individual's name. There is a standard from ORCID regarding how ORCIDs should be displayed; see https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier. 

The green ORCID icon can be found at https://orcid.figshare.com/articles/figure/ORCID_iD_icon_graphics/5008697.

justinlittman:
closes #1735

## Why was this change made?
Per ticket

Unit, local

image

NA

### DIFF
--- a/spec/components/works/detail_component_spec.rb
+++ b/spec/components/works/detail_component_spec.rb
@@ -74,6 +74,20 @@ RSpec.describe Works::DetailComponent, type: :component do
     end
   end
 
+  describe 'contributors' do
+    let(:work_version) { build_stubbed(:work_version, authors: [author1, author2], contributors: [contributor]) }
+    let(:author1) { build_stubbed(:person_author, orcid: 'https://orcid.org/0000-0002-1825-0097') }
+    let(:author2) { build_stubbed(:person_author) }
+    let(:contributor) { build_stubbed(:person_contributor, orcid: 'https://orcid.org/0000-0002-1825-0098') }
+
+    it 'renders the contributor' do
+      expect(rendered.css('#contributors a').to_html).to include('https://orcid.org/0000-0002-1825-0097')
+      expect(rendered.css('#contributors a').to_html).to include('https://orcid.org/0000-0002-1825-0098')
+      # Edit pencil and two links. No link for author without ORCID.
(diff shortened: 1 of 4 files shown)


app/components/works/orcid_component.rb
```diff
@@ -0,0 +1,13 @@
+# typed: true
+# frozen_string_literal: true
+
+module Works
```
jcoyne, line 4: Can you move this component into the `Works::Show` module?

jcoyne (changes requested):

spec/components/works/detail_component_spec.rb
```diff
@@ -74,6 +74,15 @@
...
   end
 
+  describe 'contributors' do
+    let(:work_version) { build_stubbed(:work_version, authors: [author]) }
+    let(:author) { build_stubbed(:person_author, orcid: 'https://orcid.org/0000-0002-1825-0097') }
+
+    it 'renders the contributor' do
+      expect(rendered.css('#contributors a').to_html).to include('https://orcid.org/0000-0002-1825-0097')
+    end
+  end
```
ndushay, line 84: would it be more complete to have also a spec for the contributor role?   And also to have a spec for a contributor without an ORCID?  And full name vs. first name and last name?  Asking for a friend.

ndushay (commented):
Were we going to have another feature branch for ORCID, with a feature switch in config? Or am I misremembering? 

Making this review a comment;  do with it what you will.

spec/components/works/detail_component_spec.rb
justinlittman: I'll add one for a contributor; I wasn't looking to write complete tests for the contributors section.

jcoyne (approved):

PR #1903, opened 2021-08-06
Remove Sorbet

jcoyne:
## Why was this change made?
Sorbet is pretty hard to deal with and doesn't work especially well with Rails


peetucket (approved):

peetucket:
I didn't review every thing of course... but a quick scan looks fine to me

PR #1910, opened 2021-08-09
allow reserved purls to be deleted

### LINKED ISSUE #1837: On dashboard for Reserved PURLs, add trash can icon next to pencil icon.
This should do the same thing that the trash can icon does for the Reserved PURL, which is that it gets rid of the object entirely (I'm not sure what's happening to it on the back end).

@jcoyne Please let me know if you have any questions about this. I want to get this in the Ready if possible.

peetucket:
## Why was this change made?

Fixes #1837 - allow reserved PURLs to be deleted.

~~Question: this just destroys the `Work` and `WorkVersion` in the database, it would for any work that is in a first draft state.  I think it should be fine even after we have a held a PID?~~

Localhost browser, added delete to existing purl reservation test.

### DIFF
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -12,7 +12,7 @@ class WorkPolicy < ApplicationPolicy
   end
 
   def destroy?
-    (administrator? || depositor?) && record.persisted? && record.head.first_draft?
+    (administrator? || depositor?) && record.persisted? && (record.head.first_draft? || record.head.purl_reservation?)
   end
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Please add a test to this PR.

peetucket:
Updated an existing test that reserves a purl to confirm it can be deleted and that it removes the row from the database.

jcoyne (approved):

PR #1935, opened 2021-08-11
align items show page table columns

### LINKED ISSUE #1336: Columns in the tables on the item view page should be aligned.
See wireframes: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/420930436

Currently, these are a bit all over the place. See https://sul-h2-stage.stanford.edu/works/13373 for an example.

peetucket:
## Why was this change made?

Fixes #1336 : align (most) fields on item view page with a fixed width column for the labels.

I didn't align the Files, Author and Contributors and History sections because they are currently displayed differently.  The aligned sections have a label and a value in each row, but the the other three sections noted above are tables, with each row containing only values, with the labels at the top of the table columns.  If we also want to align those, we may run into other issues with the display (for example, a person's name will be long and won't be suitable for the smaller column width that the labels in the other sections are constrained to):

**Current view on main branch:**

before

**View in this PR:**

Screenshot 2021-08-12 at 11-20-01 SDR A test collection purl reserved

Localhost browser

### DIFF
--- a/app/components/works/detail_component.html.erb
+++ b/app/components/works/detail_component.html.erb
@@ -25,7 +25,7 @@
     <caption>Details</caption>
     <tbody>
       <tr>
-        <th scope="row">Persistent link</th>
+        <th class="col-3" scope="row">Persistent link</th>
         <td>
           <% if purl %>
             <%= link_to purl, purl %>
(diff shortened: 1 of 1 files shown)


app/packs/stylesheets/work.scss
```diff
@@ -47,6 +47,10 @@
       margin-bottom: 0rem;
     }
 
+    th {
```
jcoyne, line 50: We don't want to apply this to the files table and only want to apply it to the first column of each table.  How about adding the `col-3` class just to the `th` tags that need it?

jcoyne (changes requested):

app/packs/stylesheets/work.scss
peetucket: Ok, switched to using col-3 class where needed instead of in css

peetucket:
Updated the screenshot, looks basically the same as the previous CSS based method.

jcoyne (approved):

PR #1947, opened 2021-08-12
Allow rejected or pending approval first drafts to be deleted

### LINKED ISSUE #1946: Include "Discard draft" button for items in the review workflow
Items that have been 
- submitted for approval
- returned for review

should still be able to be deleted entirely. These should have the "Discard draft" button at the bottom of the form, as well as the trash can icon on the relevant tables on the dashboard and collection deposits lists.

peetucket:
## Why was this change made?

Fixes #1946 - allow rejected (returned) or pending approval first drafts to be deleted

Localhost browser
Updated tests

### DIFF
--- a/app/components/works/buttons_component.rb
+++ b/app/components/works/buttons_component.rb
@@ -36,7 +36,7 @@ module Works
     end
 
     def show_first_draft_cancel?
-      work_version.first_draft?
+      work_version.deleteable?
     end
   end
 end
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -107,7 +107,7 @@ class WorkVersion < ApplicationRecord
   end
 
   def deleteable?
-    (first_draft? || purl_reservation?) || (version == 1 && (pending_approval? || rejected?))
+    first_draft? || purl_reservation? || (version == 1 && (pending_approval? || rejected?))
   end
 
   def add_purl_to_citation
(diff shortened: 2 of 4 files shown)


app/components/works/buttons_component.rb
```diff
@@ -36,7 +36,7 @@ def show_version_draft_cancel?
     end
 
     def show_first_draft_cancel?
-      work_version.first_draft?
+      work_version.deleteable?
```
peetucket, line 39: this method shows the discard draft button on the work edit page and needs to have additional logic now to deal with rejected items

app/models/work_version.rb
```diff
@@ -106,6 +106,10 @@ def draft?
     version_draft? || first_draft?
   end
 
+  def deleteable?
+    (first_draft? || purl_reservation?) || (version == 1 && (pending_approval? || rejected?))
+  end
+
```
peetucket, line 112: this could go on the `WorkPolicy` too but it's unclear to me how this would get accessed from the `ButtonsComponent` if so

justinlittman: Are the first set of parentheses necessary? If I'm reading correctly, there are 3 OR conditions.

justinlittman (approved):

peetucket:
I think you are right, removed the extra parens.

PR #1953, opened 2021-08-13
go back to collection details/settings when canceling editing

### LINKED ISSUE #1949: When in collection edit form and selecting "cancel", user should be taken to the Collection view page
Ideally, if they were editing settings they should go to the settings tab and if they were editing details they should be taken to the details tab.

Currently, it takes you back to the dashboard.

peetucket:
## Why was this change made?

Fixes #1949 - cancel target location should be collection view pages when editing

Added new expectations to edit collection spec

### DIFF
--- a/spec/features/create_new_collection_spec.rb
+++ b/spec/features/create_new_collection_spec.rb
@@ -43,22 +43,14 @@ RSpec.describe 'Create a new collection', js: true do
       expect(page).to have_content('Depositing')
 
       # Simulate the deposit process
-      collection_version = CollectionVersion.last
-      collection_version.collection.update!(druid: 'druid:dc224fz4940')
-      collection_version.deposit_complete!
+      CollectionVersion.last.tap do |version|
+        version.collection.update!(druid: 'druid:dc224fz4940')
+        version.deposit_complete!
+      end
 
       # This link appears when depositing is complete
       click_link "Edit #{name}"
       expect(page).not_to have_content('Depositing')
-
-      # Cancel editing to confirm we get back to the collection settings view page
-      click_link 'Cancel'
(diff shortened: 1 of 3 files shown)


spec/features/create_new_collection_spec.rb
```diff
@@ -43,14 +43,22 @@
       expect(page).to have_content('Depositing')
 
       # Simulate the deposit process
-      CollectionVersion.last.tap do |version|
-        version.collection.update!(druid: 'druid:dc224fz4940')
-        version.deposit_complete!
-      end
+      collection_version = CollectionVersion.last
```
peetucket, line 46: i'm now making use of this object in other expectations below, so I am saving it here

spec/features/create_new_collection_spec.rb
```diff
@@ -43,14 +43,22 @@
...
+      collection_version.collection.update!(druid: 'druid:dc224fz4940')
+      collection_version.deposit_complete!
 
       # This link appears when depositing is complete
       click_link "Edit #{name}"
       expect(page).not_to have_content('Depositing')
+
+      # Cancel editing to confirm we get back to the collection settings view page
```
jcoyne, line 54: Can we revert this test? I think that using a feature spec to test link resolution will unnecessarily slow the test suite.  This effectively tests that the browser follows links, which I believe is something we can take for granted.  If we want to add a test, how about checking that the link exists in the component test, which is much faster.

jcoyne (changes requested):

spec/features/create_new_collection_spec.rb
peetucket: Just removed the feature test.  No current component tests for the affected components ... none added at this point.

jcoyne (approved):

PR #1954, opened 2021-08-13
cancel work editing goes to work show page

### LINKED ISSUE #1950: When in work edit form and user selects "cancel", user should be taken to the work view page
Currently, user is taken to the collection deposits list.

peetucket:
## Why was this change made?

Fixes #1950 - cancel link on work edit form should go to work show page (if existing work) or collection works listing page (if new work)

Updated existing test, added new test

### DIFF
--- a/spec/features/create_new_work_spec.rb
+++ b/spec/features/create_new_work_spec.rb
@@ -228,21 +228,6 @@ RSpec.describe 'Create a new work in a deposited collection', js: true do
       user.update(last_work_terms_agreement: Time.zone.now.years_ago(2))
     end
 
-    context 'when user clicks cancel link' do
-      it 'goes to collection deposits page if you click cancel' do
-        visit dashboard_path
-
-        click_button '+ Deposit to this collection' # , match: :first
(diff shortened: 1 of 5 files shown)


spec/features/create_new_work_spec.rb
```diff
@@ -228,6 +228,21 @@
       user.update(last_work_terms_agreement: Time.zone.now.years_ago(2))
     end
 
+    context 'when user clicks cancel link' do
```
jcoyne, line 231: Can we revert this test? I think that using a feature spec to test link resolution will unnecessarily slow the test suite. This effectively tests that the browser follows links which I believe is something we can take for granted. If we want to add a test, how about checking that the link exists in the component test. The component tests are  much faster.

jcoyne (changes requested):

spec/features/create_new_work_spec.rb
peetucket: Ok, removed this feature test and added tests to the component test.

jcoyne (approved):

Issue #1951, opened 2021-08-13
User is unable to delete a reserved PURL


amyehodge:
Kristin Kutella is the manager of the Stanford Center for East Asian Studies Thesis Collection (https://sdr.stanford.edu/collections/23). A student reserved a PURL when they shouldn't have and she would like to delete it.

(Note: ticket 1837 adds trash can icons for these to the dashboard but is not in production yet.)

She has clicked the pencil icon and selected "Text" as the work type in order to get into the edit form. When she clicks on the "discard draft" button at the bottom of the page she sees the warning message about not being able to undo this action. She clicks "OK" (or whatever it says there) and then is dumped back onto the welcome page at sdr.stanford.edu with this error:
"You are not authorized to perform the requested action."
![Screen Shot 2021-08-13 at 12.29.43 PM.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/b365c92f-7081-4bd8-8644-94ce0c89909a)

Note that I was able to delete this reserved PURL using the same procedure without any trouble.

amyehodge:
@amyehodge Note to self to let Kristin know when we've fixed this.

peetucket:
This is a collection manager trying to delete someone else's reserved PURL.

peetucket:
I have a suspicion that the work already done in #1910 and in review in #1947 will fix this.

peetucket:
I'm wondering if we need to add collection managers in addition to depositors and administrators to this check here:  https://github.com/sul-dlss/happy-heron/blob/main/app/policies/work_policy.rb#L14-L16

amyehodge:
@peetucket Reviewers should also be able to do this. That may be another distinct set of user permissions?

amyehodge:
Here are some samples of reserved PURLs that can't be deleted:

Item: https://sdr.stanford.edu/works/4658
Collection: https://sdr.stanford.edu/collections/67
User: akrogman

Item: need to check
Collection: https://sdr.stanford.edu/collections/23
User: kkutella

jcoyne:
<img width="1158" alt="Screen Shot 2021-08-18 at 3 37 21 PM" src="https://user-images.githubusercontent.com/92044/129968920-2593a04f-5d38-478f-a58d-4449d5edb1a7.png">
It looks like you can delete it from the collection page.

peetucket:
[...]

PR #1964, opened 2021-08-16
allow reviewers and collection managers to delete works

peetucket:
## Why was this change made?

To address #1951 - add collection managers and reviewers to the work delete policy

Added new tests

### DIFF
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -12,10 +12,15 @@ class WorkPolicy < ApplicationPolicy
   end
 
   def destroy?
-    (administrator? || depositor?) && record.persisted? && record.head.deleteable?
+    (administrator? || depositor? || reviews_collection? || manages_collection?(collection)) &&
+      record.persisted? && record.head.deleteable?
   end
 
   delegate :administrator?, to: :user_with_groups
+
+  def reviews_collection?
+    allowed_to?(:review?, record.collection)
+  end
 
   def depositor?
     record.depositor == user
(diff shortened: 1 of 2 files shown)


jcoyne (changes requested):
Please add a test

spec/features/delete_draft_work_spec.rb
```diff
@@ -16,18 +16,116 @@
 
   before do
     create(:collection_version_with_collection, collection: collection)
-    sign_in user
   end
 
-  context 'when draft' do
+  context 'when depositor logged in' do
```
peetucket, line 21: These first set of tests are the same, I just wrapped in new context block for the depositor being logged in.

spec/features/delete_draft_work_spec.rb
```diff
@@ -16,18 +16,116 @@
...
+        accept_confirm do
+          click_link 'Discard draft'
+        end
+        expect(Work.exists?(work_rejected.id)).to be false
+        expect(page).to have_current_path(collection_works_path(collection.id))
+      end
+    end
+  end
+
+  # shared examples for collection managers and reviewers
```
peetucket, line 95: some shared examples used by new tests by collection managers and reviewers being logged in

peetucket:
Added tests for collection manager nd reviewer deletion of works

spec/features/delete_draft_work_spec.rb
jcoyne: This is not a new feature, so we don't need a new (slow) feature tests.  Have you considered adding to the policy unit tests instead?

jcoyne (changes requested):
Can we just do a policy test and not the (slow) feature tests?

peetucket:
Created new work policy spec and reverted feature tests.

jcoyne (approved):

PR #1965, opened 2021-08-17
Determines type of subject terms from FAST and include in cocina.

### LINKED ISSUE #1874: Determine type of subject terms from FAST and include in cocina
(cc @amyehodge )

Each FAST subject term falls into one of these categories:
* Personal Names (cocina type: person)
* Corporate Names (cocina type: organization)
* Event (cocina type: event)
* Uniform Titles (cocina type: title)
* Chronological (cocina type: time)
* Topical (cocina type: topic)
* Geographic (cocina type: place)
* FormGenre (cocina type: genre)
* Meeting (cocina type: conference)

It is important to know the category a given term belongs to because terms of different categories are treated in different ways in systems such as SearchWorks (such as the facet a subject term goes into, and what heading it displays under in the record). Currently all subject terms are being typed as "topic", which means that terms such as places, names, and titles will not be handled correctly.

H2 should determine the type of the subject in FAST and include the corresponding cocina subject.type according to the mapping above.

MODS mappings are already in place for the different subject types. Data created prior to resolving this issue will need to be reviewed and remediated.

justinlittman:
closes #1874

## Why was this change made?
Accurate subject metadata.

Unit

NA

### DIFF
--- a/app/components/works/keywords_row_component.html.erb
+++ b/app/components/works/keywords_row_component.html.erb
@@ -3,7 +3,11 @@
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
     <div data-controller="autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
       <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
-      <%= form.hidden_field :uri, { required: true, 'data-autocomplete-target': 'hidden', 'data-autocomplete-edit-target': 'uri', 'data-action': 'autocomplete-edit#change' }%>
+      <!-- This input is populated by autocomplete controller.-->
+      <!-- That triggers the autocomplete-edit controller, which populates the uri and cocina type inputs. -->
+      <input type="hidden" data-autocomplete-target="hidden" data-autocomplete-edit-target="value" data-action="autocomplete-edit#change" />
+      <%= form.hidden_field :uri, { required: true, 'data-autocomplete-edit-target': 'uri' }%>
+      <%= form.hidden_field :cocina_type, { required: true, 'data-autocomplete-edit-target': 'type' }%>
       <ul class="show dropdown-menu" data-autocomplete-target="results"></ul>
       <div class="invalid-feedback">Keyword must be filled in</div>
     </div>
--- a/app/controllers/autocomplete_controller.rb
+++ b/app/controllers/autocomplete_controller.rb
@@ -44,7 +44,7 @@ class AutocompleteController < ApplicationController
       req.params['start'] = 0
       req.params['version'] = '2.2'
       req.params['indent'] = 'on'
-      req.params['fl'] = 'id,fullphrase'
+      req.params['fl'] = 'id,fullphrase,type'
       req.params['sort'] = 'usage desc'
     end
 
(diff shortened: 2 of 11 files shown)


app/controllers/autocomplete_controller.rb
```diff
@@ -103,7 +104,34 @@ def sort_suggestions(suggestions)
...
 
-  def authority_uri(idroot)
-    "#{URI_PREFIX}#{idroot.delete_prefix('fst').to_i}/"
+  def key(idroot, type)
+    "#{URI_PREFIX}#{idroot.delete_prefix('fst').to_i}/::#{cocina_type(type)}"
   end
+
+  # rubocop:disable Metrics/CyclomaticComplexity
+  # rubocop:disable Metrics/MethodLength
+  def cocina_type(type)
```
jcoyne, line 113: Can this be a Hash rather than a case statement?

jcoyne (changes requested):

app/controllers/autocomplete_controller.rb
justinlittman: Yup. Fixed.

app/components/works/keywords_row_component.html.erb
```diff
@@ -3,7 +3,10 @@
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
     <div data-controller="autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
       <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
-      <%= form.hidden_field :uri, { required: true, 'data-autocomplete-target': 'hidden', 'data-autocomplete-edit-target': 'uri', 'data-action': 'autocomplete-edit#change' }%>
+      <%#= form.hidden_field :value, { 'data-autocomplete-target': 'hidden', 'data-autocomplete-edit-target': 'value', 'data-action': 'autocomplete-edit#change' }%>
```
jcoyne, line 6: Is this cruft?

app/components/works/keywords_row_component.html.erb
```diff
@@ -3,7 +3,10 @@
...
+      <input type="hidden" data-autocomplete-target="hidden" data-autocomplete-edit-target="value" data-action="autocomplete-edit#change" />
```
jcoyne, line 7: What is the purpose of this? How can a hidden field trigger a change?

jcoyne (changes requested):

app/packs/controllers/autocomplete_edit_controller.js
```diff
@@ -1,15 +1,19 @@
...
-  static targets = ["uri", "input"]
+  static targets = ["uri", "input", "value", "type"]
 
   connect() {
     this.change()
   }
 
   change() {
-    if(this.uriTarget.value) {
+    console.log(this.valueTarget.value)
```
jcoyne, line 11: remove this

justinlittman: Duh.

app/components/works/keywords_row_component.html.erb
justinlittman: This hidden input is populated by the autocomplete controller (recall, an instance of stimulus-autocomplete). In turn this triggers the autocomplete-edit controller to split the value and populate the uri and cocina type inputs.

The autocomplete controller changing the hidden field does trigger the change -- not sure how.

jcoyne: Can you add a comment about this mechanism?

justinlittman: Done.

jcoyne (approved):

PR #1980, opened 2021-08-18
allow reviewers and collection managers to also delete work_versions

peetucket:
## Why was this change made?

To address https://github.com/sul-dlss/happy-heron/issues/1951#issuecomment-901451204 -- it looks like perhaps our updating of the work policy in #1964 now allows the delete icon to be shown for everyone who can delete, but when the deletion is actually attempted, the work_version policy prevented the deletion (for a collection manager).  The theory is that it works for admins and depositors since they are currently allowed to delete work_versions, but not for collection managers or reviewers because they are not currently allowed to delete work_versions.  This code matches up the work and work_version policies so that they are the same group of people.

Added tests

### DIFF
--- a/app/policies/work_version_policy.rb
+++ b/app/policies/work_version_policy.rb
@@ -57,7 +57,8 @@ class WorkVersionPolicy < ApplicationPolicy
   end
 
   def destroy?
-    (administrator? || depositor?) && record.persisted? && record.version_draft?
+    (administrator? || depositor? || reviews_collection? || manages_collection?(collection)) &&
+      record.persisted? && record.draft?
   end
 
   private
--- a/spec/policies/work_version_policy_spec.rb
+++ b/spec/policies/work_version_policy_spec.rb
@@ -151,6 +151,100 @@ RSpec.describe WorkVersionPolicy do
     end
   end
 
+  describe_rule :destroy? do
+    context 'when persisted and version_draft' do
+      before { record.state = 'version_draft' }
+
+      failed 'when user is not a depositor, reviewer or manager for the collection'
+
+      # `succeed` is `context` + `specify`, which checks
+      # that the result of application wasn't successful
+      succeed 'when user is a depositor' do
+        let(:work) { build_stubbed :work, depositor: user }
+      end
+
+      succeed 'when user is a collection manager' do
+        before { collection.managed_by = [user] }
+      end
+
+      succeed 'when user is a reviewer' do
+        before { collection.reviewed_by = [user] }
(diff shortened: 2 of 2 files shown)


app/policies/application_policy.rb
```diff
@@ -6,6 +6,14 @@ class ApplicationPolicy < ActionPolicy::Base
 
   private
 
+  def can_delete_works?(collection)
```
jcoyne, line 9: I don't think that either of these methods should be inherited by all policies

jcoyne (changes requested):

jcoyne (commented):
This is challenging to review because it combines a refactoring with a behavior change.  Can you split into a PR for each?

app/policies/work_version_policy.rb
```diff
@@ -47,17 +47,17 @@ def update?
...
   # The collection reviewers can review a work
 
   def review?
-    record.pending_approval? && reviews_collection?
+    record.pending_approval? && reviews_collection?(collection)
   end
 
   def destroy?
-    (administrator? || depositor?) && record.persisted? && record.version_draft?
+    can_delete_works?(collection) && record.persisted? && record.version_draft?
```
jcoyne, line 60: Can this first part be `allowed_to?(:destroy?, record.head)` ?

jcoyne (changes requested):

peetucket:
Ok, I've removed the commit that did some refactoring and just left the change to the policy.  If this works, I can do a follow on with the refactoring if that's helpful.

app/policies/work_version_policy.rb
peetucket: I don't think so, because the work policy for deletion is more restrictive than the work_version policy for deletion (since you can delete a work version when its a version draft without also deleting the work itself, but not vice versa).  So there will be some common checks (like who is allowed to do it), but we can't do *all* of the work checks for the work_version.

app/policies/work_version_policy.rb
```diff
@@ -57,7 +57,8 @@ def review?
   end
 
   def destroy?
-    (administrator? || depositor?) && record.persisted? && record.version_draft?
+    (administrator? || depositor? || reviews_collection? || manages_collection?(collection)) &&
+      record.persisted? && record.draft?
```
peetucket, line 61: note: switched `version_draft?` to `draft?` as well as adding reviewers and collection managers ... I believe this is correct because we should be able to delete a first draft of a work_version if we delete the work itself

PR #1983, opened 2021-08-19
refactor policies to remove code duplication for delete works

peetucket:
## Why was this change made?

Follow on from #1980 to refactor work and work_version policy to avoid code duplication.  Create new base class for works and work_versions, move a couple methods there used by all work related policies.

Existing tests

### DIFF
--- a/app/policies/common_work_policy.rb
+++ b/app/policies/common_work_policy.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# Base class for work and work_version policies
+class CommonWorkPolicy < ApplicationPolicy
+  alias_rule :delete?, to: :destroy?
+
+  private
+
+  def administors_collection?
+    administrator? || depositor? || reviews_collection? || manages_collection?(collection)
+  end
+
+  def reviews_collection?
+    allowed_to?(:review?, collection)
+  end
+end
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
 
 # Authorization policy for Work objects
-class WorkPolicy < ApplicationPolicy
-  alias_rule :delete?, to: :destroy?
-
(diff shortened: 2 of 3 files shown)


app/policies/application_policy.rb
```diff
@@ -6,6 +6,14 @@ class ApplicationPolicy < ActionPolicy::Base
...
   private
 
+  def administors_collection?(collection)
+    administrator? || depositor? || reviews_collection?(collection) || manages_collection?(collection)
+  end
+
+  def reviews_collection?(collection)
+    allowed_to?(:review?, collection)
+  end
+
```
peetucket, line 16: This method was basically the same in both work_policy and work_version_policy, putting it here in the same way the manages_collection? method below is here.

app/policies/application_policy.rb
```diff
@@ -6,6 +6,14 @@ class ApplicationPolicy < ActionPolicy::Base
 
   private
 
+  def administors_collection?(collection)
+    administrator? || depositor? || reviews_collection?(collection) || manages_collection?(collection)
+  end
+
```
peetucket, line 12: New common method for both the work_policy and work_version_policy.  If we don't want to have a these methods available to all subclasses, we could move it and the two methods below to a new "work" related superclass

app/policies/common_work_policy.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Base class for work and work_version policies
+class CommonWorkPolicy < ApplicationPolicy
```
peetucket, line 4: suggestion for a better name ?

app/policies/common_work_policy.rb
```diff
@@ -0,0 +1,16 @@
...
+  alias_rule :delete?, to: :destroy?
+
+  private
+
+  def administors_collection?
```
peetucket, line 9: suggestion for a better method name?

justinlittman (approved):

PR #1986, opened 2021-08-19
Adds required license when created a reserve PURL work.

### LINKED ISSUE #1979: Reserved PURL is not displaying the right default license
Item: https://sdr.stanford.edu/works/4659/edit

This collection is set for depositor selects with a CC0 1.0 license as the default. When I reserve a PURL and then open it to edit, it has "no license" displayed as the default.

justinlittman:
closes #1979

## Why was this change made?
So that required license is required.

Unit


app/forms/reservation_form.rb
```diff
@@ -9,6 +9,15 @@ class ReservationForm < Reform::Form
   property :license, on: :work_version, default: 'none'
   property :title, on: :work_version
 
+  def deserialize!(params)
+    params['license'] = collection.required_license if collection.required_license
```
jcoyne, line 13: #1979 was about a default license.  Is required_license what we ought to use here?

justinlittman: Oh, I didn't realize there was a difference, but now I see there is. Let me check with @amyehodge.

jcoyne: @justinlittman I see you made a change here.  Did you talk to @amyehodge about it?

justinlittman: Done.

app/forms/reservation_form.rb
```diff
@@ -9,11 +9,28 @@ class ReservationForm < Reform::Form
   property :license, on: :work_version, default: 'none'
   property :title, on: :work_version
 
+  def deserialize!(params)
+    params['license'] = collection.default_license if collection.default_license && !work_version.license
```
jcoyne, line 13: Do we need the check for `!work_version.license`?  I thought this form would only show up for new work_versions (and none of them have a license)?

justinlittman: Not entirely sure. This form may also be instantiated on update in the WorksController, so I erred on the cautious side.

https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/works_controller.rb#L81

PR #1992, opened 2021-08-20
Switch from webp to jpeg.

### LINKED ISSUE #1978: Change webp images on home page to something Safari on Catalina can render properly

justinlittman:
closes #1978

## Why was this change made?
Support Catalina.

The trivial file size differences didn't seem worth any additional trickery/complexity so just switched to jpeg.

Local

NA


jcoyne (commented):
splash.jpg is 5x bigger than splash.webp . Is that trivial?

justinlittman:
Yes, I think a 113KB background image delivered on one page is trivial.

PR #1993, opened 2021-08-20
Switch form-text to black.

### LINKED ISSUE #1989: Accessibility: Change font color to increase contrast
This text is gray "Name associated with this ORCID iD is Astrid Usong."
Change to black.

justinlittman:
closes #1989

## Why was this change made?
Accessability

Local

image

### DIFF
--- a/app/packs/stylesheets/bootstrapVariables.scss
+++ b/app/packs/stylesheets/bootstrapVariables.scss
@@ -27,3 +27,6 @@ $breadcrumb-divider: quote(">");
 $navbar-padding-y: 0;
 
 $nav-link-padding-y: 0.3rem;
+
+// Forms
+$form-text-color: black; // Better contrast for accessability


app/packs/stylesheets/bootstrapOverrides.scss
```diff
@@ -8,3 +8,8 @@
 .table {
   border: 1px solid $table-border-color;
 }
+
+// Better contrast for accessability
+.form-text {
```
jcoyne, line 13: Instead of overriding, set $form-text-color (see https://github.com/twbs/bootstrap/blob/f61a0218b36d915db80dc23635a9078e98e2e3e0/scss/forms/_form-text.scss#L10)

justinlittman: I went looking for that. Where do you find those variables documented?

jcoyne (approved):

PR #1995, opened 2021-08-20
Adjust spacing for contributor.

### LINKED ISSUE #1988: Refine spacing of ORCID UI
Current
image

Proposed design
image

justinlittman:
closes #1988

## Why was this change made?
Per ticket

image

### DIFF
--- a/app/components/works/contributor_row_component.html.erb
+++ b/app/components/works/contributor_row_component.html.erb
@@ -1,4 +1,4 @@
-<div class="row inner-container" data-controller="contributors" data-contributors-required-value="<%= author? %>">
+<div class="row inner-container contributor-row" data-controller="contributors" data-contributors-required-value="<%= author? %>">
   <div class="col-md-11">
     <div class="row">
       <div class="col-md-3">
@@ -10,7 +10,7 @@
         is
       </div>
       <div class="col" data-contributors-target="person">
-        <div class="row" data-contributors-target="personName">
+        <div class="row person-name-row" data-contributors-target="personName">
           <div class="col-md-1">
             <%= form.radio_button :with_orcid, 'false', html_options_for_radio(true, !orcid?) %>
           </div>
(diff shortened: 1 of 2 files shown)


app/components/works/contributor_row_component.html.erb
```diff
@@ -1,16 +1,16 @@
...
+<div class="row inner-container contributor-row" data-controller="contributors" data-contributors-required-value="<%= author? %>">
   <div class="col-md-11">
     <div class="row">
       <div class="col-md-3">
         <%= form.label :role_term, role_term_label, class: 'form-label' %>
         <%= render PopoverComponent.new key: 'work.role_term' %>
         <%= select_role %>
       </div>
-      <div class="col-md-1 role-prefix">
+      <div class="col-sm-1 role-prefix">
```
jcoyne, line 9: This switch causes this to get awkwardly positioned:

Screen Shot 2021-08-23 at 12 17 07 PM

It looked like this before.
Screen Shot 2021-08-23 at 12 15 06 PM

app/packs/stylesheets/workEditor.scss
```diff
@@ -74,4 +69,24 @@
   div.embargo-date {
     margin-bottom: 1rem !important;
   }
+
+  .contributor-row.inner-container {
```
jcoyne, line 73: Is it necessary to have `.inner-container` here? Why can't we just put this under `.contributor-row`?

jcoyne (changes requested):

app/components/works/contributor_row_component.html.erb
justinlittman: Switched back to `col-md-1`.

app/packs/stylesheets/workEditor.scss
justinlittman: It gets trumped by `.work-editor section .inner-container` in `editor.scss`, which sets a different margin.

jcoyne: Please add a comment about this so there's no accidental refactoring to something simpler.

justinlittman: Done.

jcoyne (approved):

PR #2045, opened 2021-08-31
Remove extra modal class tag causing modal not to close properly

### LINKED ISSUE #2040: Clicking "X" on reserve a purl modal should close the modal
Currently, clicking the X does nothing. Clicking the Cancel button does close the modal, as does clicking outside the modal, so there are ways out still.

aaron-collier:
## Why was this change made?

Fixes #2040 

The extra modal was removing the ability for the normal data-close action to be linked to the appropriate div.


ndushay (approved):
LGTM;  not sure if @amyehodge wants to see this deployed and test it.

PR #2068, opened 2021-10-06
Upgrade webpacker to 6.0.0.rc5

jcoyne:


jcoyne:
No longer necessary since we switched to ESbuild.

PR #2080, opened 2021-10-19
embargo release criteria should match what the user set

### LINKED ISSUE #2069: Embargoed items that are supposed to be released to Stanford-only rights are instead being released to the world
**Description of the problem**

H2 allows depositors to embargo their items for up to 3 years. During the embargo period, the files are not supposed to be available and the object should have the rights of "citation-only". These rights are not being set correctly and the result is that data is being made available before it should be released.

This ticket is for embargoed objects that are supposed to be released to Stanford visibility. I'll file a separate ticket for embargoed objects that should be released to world visibility.

**Details**

I created a test item and embargoed it until October 31. When released, it should be available only to the Stanford community. Instead, on the Purl the files have been made available to the Stanford community immediately:
https://sul-purl-stage.stanford.edu/cz498dt9764
https://argo-stage.stanford.edu/view/druid:cz498dt9764

In the embargo settings, the release is set to make the files available to the world on the release date:
```json
"embargo": {
      "access": "world",
      "download": "world",
      "releaseDate": "2021-10-31T00:00:00.000+00:00"
    },
```
It looks like what's happening is that the object level rights are correct: citation-only. But the files are being given file-level rights of "Stanford" that are overriding the object level rights. The files should not be downloadable by anyone during the embargo period.

These are the settings I chose in H2:
https://sul-h2-stage.stanford.edu/works/18549
Screen Shot 2021-10-07 at 4 41 36 PM

peetucket:
## Why was this change made?

Fixes #2069 - embargo release access rights should match what the user set to ensure when the object is released, it has the correct access

Added new unit test

### DIFF
--- a/app/services/cocina_generator/access_generator.rb
+++ b/app/services/cocina_generator/access_generator.rb
@@ -32,7 +32,7 @@ module CocinaGenerator
       {
         access: 'citation-only',
         download: 'none',
-        embargo: { access: 'world', download: 'world', releaseDate: work_version.embargo_date.iso8601 }
+        embargo: regular_access.merge({ releaseDate: work_version.embargo_date.iso8601 })
       }
     end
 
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
LGTM.

PR #2102, opened 2021-11-16
Remove dlss-capistrano provided bundler2 config

jcoyne:
## Why was this change made?
This functionality is now handled by capistrano-bundler 2.0


jmartin-sul (approved):

jmartin-sul:
approving and merging because this seemed like an overlooked stray from a larger batch of these:  https://github.com/search?q=org%3Asul-dlss+Remove+dlss-capistrano+provided+bundler2+config&type=issues

that i discovered while catching up on communication missed during my recent/ongoing jury duty.

apologies if this merge was an error, seems easy enough to revert if needed.

PR #2116, opened 2021-12-16
Make ordered author button widgets Stimulus targets

### LINKED ISSUE #2114: Sometimes upButton is nil
https://github.com/sul-dlss/happy-heron/blob/cc548f4553e755ebfaa930a05956784efe665a49/app/javascript/controllers/ordered_nested_form_controller.js#L23-L26
Because the selector is looking for a specific action property, but it's actually: `data-action="click->ordered-nested-form#moveUp auto-citation#updateDisplay"`

It should use a stimulus target instead.
Screen Shot 2021-12-15 at 10 32 55 PM

mjgiarlo:
Fixes #2114

## Why was this change made?

☝🏻 

CI, deployed to stage to make sure console error no longer logged

None

### DIFF
--- a/app/javascript/controllers/nested_form_controller.js
+++ b/app/javascript/controllers/nested_form_controller.js
@@ -1,7 +1,10 @@
 import { Controller } from "stimulus"
 
 export default class extends Controller {
-  static targets = ["add_item", "template"]
+  // Declare targets for this controller and its children because adding to
+  // inherited targets in children causes Stimulus controller errors
+  // ("ordered_nested_form_controller_default is not defined")
+  static targets = ["add_item", "template", "upButton", "downButton"]
   static values = { selector: String }
 
   addAssociation(event) {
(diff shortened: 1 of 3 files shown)


jcoyne (approved):

app/javascript/controllers/nested_form_controller.js
```diff
@@ -1,7 +1,10 @@
 import { Controller } from "stimulus"
 
 export default class extends Controller {
-  static targets = ["add_item", "template"]
+  // Declare targets for this controller and its children because adding to
+  // inherited targets in children causes Stimulus controller errors
+  // ("ordered_nested_form_controller_default is not defined")
+  static targets = ["add_item", "template", "upButton", "downButton"]
```
jcoyne, line 7: Will it error on target not found when using `data-controller="nested-form"`?

mjgiarlo: @jcoyne I'm not sure I follow, but the build is already broken so I'm not exactly rushing to get this in. What all changes are you suggesting I might want to try?

jcoyne: I'm not suggesting any change in particular. I was just wondering if we've tested both the `action-controller="nested-form"` and `action-controller="ordered-nested-form"`

mjgiarlo:
@jcoyne should we get this in or would you like changes?

PR #2140, opened 2022-01-19
parse approximate dates in the same way as non-approximate dates

### LINKED ISSUE #2139: do not insert month/day data when the creation date is approximate and contains only year or year + month
**Description of the bug**

When a user enters a creation date for a deposit and marks the date as "Approximate", H2 is adding an "01" if the month or the month+day was not entered on the form.

Example:
https://sdr.stanford.edu/works/4939 (H2 entry)
https://argo.stanford.edu/view/nw987sj3048 (as deposited)

The date range for this item has been entered as an approximate year to another approximate year. But when writing to Cocina, an "01-01" for month and day was added to each date in the range.

**Expected behavior**

Users are given the option to enter a creation date for their deposits in the following forms:

- year
- year, month
- year, month, day

They can also:

- enter a date range, where the start and end dates each follow the same format as for a single date
- check a box labeled "Approximate" for any or all dates entered
Screen Shot 2022-01-18 at 3 05 54 PM

If the date entered does not include a month or day, no month or day should be added. If the creation date is "2020" it should not be transformed into "2020-01-01". Similarly, "ca. 2020" should not be transformed into "ca. 2020-01-01"

**Additional information**

An earlier PR fixed this issue for non-approximate dates but may not have fixed the approximate date case: https://github.com/sul-dlss/happy-heron/pull/2085

I've tested all of these creation date possibilities and how they get represented in Cocina after deposit:

For one creation date (i.e. not a range):

- year only: correct Cocina
- year-month: correct Cocina
- approx year: 01-01 is added
- approx year-month: 01 is added

For date ranges:

- year-to-year range: correct Cocina
- year-month to year-month range: correct Cocina
- approx year-to-year range: 01-01 is added to the approx date
- approx year-month-to-year-month range: 01 is added to the approx date

So I think it's really only the approx dates that have this issue now.

peetucket:
## Why was this change made?

Fixes #2139 - approximate dates should be parsed in the same way as non-approximate dates when added to cocina, they just get a qualifier added to the cocina metadata

Added new fixtures and tests (confirmed they failed before my code change and passed after it)

### DIFF
--- a/app/services/cocina_generator/description/generator.rb
+++ b/app/services/cocina_generator/description/generator.rb
@@ -134,7 +134,7 @@ module CocinaGenerator
         {
           type: type,
           qualifier: date.uncertain? ? 'approximate' : nil,
-          value: date.uncertain? ? date.to_s : date.edtf
+          value: date.edtf.chomp('?')
         }.compact
       end
 
--- a/spec/factories/work_versions.rb
+++ b/spec/factories/work_versions.rb
@@ -82,6 +82,14 @@ FactoryBot.define do
     created_edtf { EDTF.parse('2020-03-08?') }
   end
 
+  trait :with_approximate_creation_date_year_only do
+    created_edtf { EDTF.parse('2020?') }
+  end
+
+  trait :with_approximate_creation_date_year_month_only do
+    created_edtf { EDTF.parse('2020-06?') }
+  end
+
   trait :with_keywords do
     transient do
       keywords_count { 3 }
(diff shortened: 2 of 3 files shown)


app/services/cocina_generator/description/generator.rb
```diff
@@ -134,7 +134,7 @@ def date_props_for(date, type: nil)
         {
           type: type,
           qualifier: date.uncertain? ? 'approximate' : nil,
-          value: date.uncertain? ? date.to_s : date.edtf
+          value: date.edtf.chomp('?')
         }.compact
```
peetucket, line 138: edtf formatted dates include a `?` when approximate, just chomp it anytime it is there

jcoyne (approved):

PR #2181, opened 2022-03-05
set default time zone for the app to Pacific; revert embargo changes

peetucket:
## Why was this change made?

This reverts #2179 (use Pacific time for embargo dates), and instead just sets the application's default time zone to Pacific.  This will have the same effect as the reverted change for the embargo dates, and will also make the date displays for users show in the Pacific time zone.  This will prevent weird "from the future" things like below happening, where I edited an item on March 4 after 4pm, but it showed as last updated on March 5 (since anytime after 4pm Pacific is now the next day in UTC).  Since we display dates (and not timestamps) to users, I think this change should do the expected thing for anyone in North America.

Screen Shot 2022-03-04 at 4 29 10 PM

Tests and localhost browser.

### DIFF
--- a/app/validators/embargo_date_validator.rb
+++ b/app/validators/embargo_date_validator.rb
@@ -16,7 +16,7 @@ class EmbargoDateValidator < ActiveModel::EachValidator
   def valid_embargo_collection(collection, record, attribute, value)
     year_match = collection.release_duration.match(/(\d+)/)
     year_label = year_match[0].to_i < 2 ? 'year' : 'years'
-    future_date = current_date + year_match[0].to_i.years
+    future_date = Time.zone.today + year_match[0].to_i.years
     error_message = "must be less than #{year_match[0]} #{year_label} in the future"
     if value > future_date
       record.errors.add attribute, error_message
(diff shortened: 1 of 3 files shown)


jmartin-sul (approved):

jcoyne:
I would suggest that the correct way to fix that bug in the display would be to use the [LocalTimeComponent](https://github.com/sul-dlss/happy-heron/blob/main/app/components/local_time_component.rb). Then the dates show correct for whatever timezone the user happens to be in: https://github.com/sul-dlss/happy-heron/blob/main/app/components/local_time_component.rb

I would be *very* careful deploying the change you just made because this app was fully tested with the UTC timezone and changing zones possibly introduces a number of new bugs.

peetucket:
> I would suggest that the correct way to fix that bug in the display would be to use the [LocalTimeComponent](https://github.com/sul-dlss/happy-heron/blob/main/app/components/local_time_component.rb). Then the dates show correct for whatever timezone the user happens to be in: https://github.com/sul-dlss/happy-heron/blob/main/app/components/local_time_component.rb
> 
> I would be _very_ careful deploying the change you just made because this app was fully tested with the UTC timezone and changing zones possibly introduces a number of new bugs.

That makes sense, after making the change I wondered about other side effects.  I'm not sure really what could they be,  but would rather revert and tackle on the display side instead instead of risk adding new issues.  See #2183

peetucket:
Attempt at using the `LocalTimeComponent`: #2184

Issue #2198, opened 2022-03-21
Pin ViewComponent back to minor-level and troubleshoot the update


mjgiarlo:
When ViewComponent released version 2.51.0, it broke our build. Specifically, it broke the `spec/components/collections/show/details_header_component_spec.rb` test, with the following error message:

```
Failure/Error: expect(rendered.css('turbo-frame')[1]['src']).to eq "/collection_versions/#{collection_version.id}/edit_link"
  Expected \e[33m"/collections/1068/edit_link"\e[0m to eq \e[35m"/collection_versions/1069/edit_link"\e[0m.
./spec/components/collections/show/details_header_component_spec.rb:18:in `block (3 levels) in <top (required)>'
```

At that time, we decided to pin ViewComponent to 2.50.0 (the prior version, which worked fine). Pin ViewComponent *back* to the minor-level and troubleshoot the error above. (Could be a bug in ViewComponent; could be a bug in how H2 uses ViewComponent.)

jcoyne:
This seems to stem from https://github.com/github/view_component/issues/1315

mjgiarlo:
Closed by #2204

PR #2204, opened 2022-03-28
Unpin ViewComponent now that the bug introduced in 2.51.0 has been fixed

mjgiarlo:

No need to pin the dependency at patch-level now that the bug is fixed.

CI


jcoyne (approved):

mjgiarlo:
Closes #2198

PR #2233, opened 2022-04-26
add some HB context for when deposit jobs fail

### LINKED ISSUE #2217: Include the work number, depositor sunet, and (if available) the druid, in H2 error messages
Error messages like https://app.honeybadger.io/projects/77112/faults/84949137 include a WorkVersion but it's not clear to me if there's a way to get from the work version number to the work number without going through the H2 database. The URLs in H2 seems to be based on work number, though could be missing something there.

If the HB messages include the work number, then anyone with access to error message could go straight to the work URL in H2 and see the current status of the item.

peetucket:

Fixes #2217 - when a DepositJob fails, we can add some HB context (like druid, depositor, work id) to help debug.

Also -- add new rubocop cops and autofix errors.

It was not, but this is a pretty minor change.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,7 +40,7 @@ class ApplicationController < ActionController::Base
   def groups_from_request_env
     session['groups'] = begin
       raw_header = request.env[Settings.authorization_group_header]
-      raw_header = ENV['ROLES'] if Rails.env.development? # rubocop:disable Rails/EnvironmentVariableAccess
+      raw_header = ENV.fetch('ROLES', nil) if Rails.env.development? # rubocop:disable Rails/EnvironmentVariableAccess
       logger.debug("Roles are #{raw_header}")
       raw_header&.split(';') || []
     end
--- a/app/jobs/deposit_job.rb
+++ b/app/jobs/deposit_job.rb
@@ -4,7 +4,11 @@
 class DepositJob < BaseDepositJob
   queue_as :default
 
+  # rubocop:disable Metrics/AbcSize:
   def perform(work_version)
+    Honeybadger.context({ work_version_id: work_version.id, druid: work_version.work.druid,
+                          work_id: work_version.work.id, depositor_sunet: work_version.work.depositor.sunetid })
+
     request_dro = CocinaGenerator::DROGenerator.generate_model(work_version: work_version)
     blobs = work_version.attached_files.map { |af| af.file.attachment.blob }
 
@@ -21,6 +25,7 @@ class DepositJob < BaseDepositJob
       update(new_request_dro)
     end
   end
+  # rubocop:enable Metrics/AbcSize:
 
   private
 
--- a/config/initializers/devise_remote_user.rb
+++ b/config/initializers/devise_remote_user.rb
@@ -2,7 +2,7 @@
 
 DeviseRemoteUser.configure do |config|
   config.env_key = lambda do |env|
-    if Rails.env.development? && ENV['REMOTE_USER']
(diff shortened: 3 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -40,7 +40,7 @@ def ldap_groups
   def groups_from_request_env
     session['groups'] = begin
       raw_header = request.env[Settings.authorization_group_header]
-      raw_header = ENV['ROLES'] if Rails.env.development? # rubocop:disable Rails/EnvironmentVariableAccess
+      raw_header = ENV.fetch('ROLES', nil) if Rails.env.development? # rubocop:disable Rails/EnvironmentVariableAccess
```
mjgiarlo, line 43: AFAICT, this *should* operate the same as it did before. Can you say more about this change?

config/initializers/devise_remote_user.rb
```diff
@@ -2,7 +2,7 @@
 
 DeviseRemoteUser.configure do |config|
   config.env_key = lambda do |env|
-    if Rails.env.development? && ENV['REMOTE_USER']
+    if Rails.env.development? && ENV.fetch('REMOTE_USER', nil)
```
mjgiarlo, line 5: Can you say more about this change as well?

mjgiarlo (approved):
Approving with a couple questions about `ROLES` shenanigans

config/initializers/devise_remote_user.rb
peetucket: Should be the same as before (I poked around in the Rails console).  This and the previous ENV change were made by a `rubocop -a` after adding the new configs it showed on standard out when running rubocop.

app/controllers/application_controller.rb
ndushay: this is due to Style/FetchVar cop;  it's to avoid inadvertent nil coming from ENV vars, I think.

app/jobs/deposit_job.rb
```diff
@@ -4,7 +4,11 @@
 class DepositJob < BaseDepositJob
   queue_as :default
 
+  # rubocop:disable Metrics/AbcSize:
   def perform(work_version)
+    Honeybadger.context({ work_version_id: work_version.id, druid: work_version.work.druid,
+                          work_id: work_version.work.id, depositor_sunet: work_version.work.depositor.sunetid })
+
```
ndushay, line 11: this is all that's needed?  we don't need to explicitly pass the context somewhere?

config/initializers/devise_remote_user.rb
mjgiarlo: Ah, ok. Thanks, rubocop! Merging.

ndushay (commented):
FYI, I usually add new rubocops (and their concommitant code changes) in a separate commit ... possible fodder for the future.   (In fact, i try to do it as a separate PR so it's not entangled with other actual work.)

app/jobs/deposit_job.rb
peetucket: I believe that is all that is needed...HB will add this context to the alert if it sends a notification (here is an example from elsewhere in the code: https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/record_embargo_release_job.rb#L13)

peetucket:
> FYI, I usually add new rubocops (and their concommitant code changes) in a separate commit ... possible fodder for the future. (In fact, i try to do it as a separate PR so it's not entangled with other actual work.)

yes, should have been a separate PR...got lazy here

app/jobs/deposit_job.rb
mjgiarlo: Yes, that is the case AFAIK too

PR #2239, opened 2022-05-04
Fix bug:  agree_to_terms? is true when User.last_work_terms_agreement is nil (which should be impossible)

peetucket:

Occasional HB issue occurring: https://app.honeybadger.io/projects/77112/faults/85235636

Speculation about activerecord caching inconsistencies 

Related to #2235

### DIFF
--- a/app/components/works/agreement_component.html.erb
+++ b/app/components/works/agreement_component.html.erb
@@ -4,7 +4,7 @@
     <div class="form-check col-auto">
       <% if agree_to_terms? %>
         You agreed to the <%= link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'} %>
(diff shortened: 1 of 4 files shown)


app/components/works/agreement_component.html.erb
```diff
@@ -4,7 +4,7 @@
     <div class="form-check col-auto">
       <% if agree_to_terms? %>
         You agreed to the <%= link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'} %>
-        <%= time_ago_in_words form.object.model.fetch(:work).depositor.last_work_terms_agreement %> ago.
+        <%= time_ago_in_words date_last_agreed %> ago.
```
ndushay, line 7: trying to make this date lookup more like `agree_to_terms?`

ndushay (commented):
Partial attempt here;  it currently throws errors when run in the app and has no tests.

ndushay:
Closing this as it turned out not to address the problem.  Peter - you might want to release the brach.

PR #2241, opened 2022-05-05
fix bug with embargo release date and user without recent terms of agreement

### LINKED ISSUE #2235: [happy-heron/prod] ArgumentError: nil can't be converted to a Time value
## Backtrace

    line 7 of [PROJECT_ROOT]/app/components/works/agreement_component.html.erb: call
    line 32 of [PROJECT_ROOT]/app/components/works/form_component.html.erb: block in call
    line 2 of [PROJECT_ROOT]/app/components/works/form_component.html.erb: call

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/85235636)

ndushay:

To avoid https://app.honeybadger.io/projects/77112/faults/85235636

Fixes #2235

### DIFF
--- a/app/components/works/agreement_component.html.erb
+++ b/app/components/works/agreement_component.html.erb
@@ -2,7 +2,9 @@
   <header>Terms of Deposit *</header>
   <div class="row">
     <div class="form-check col-auto">
-      <% if agree_to_terms? %>
+      <%# TODO: create a method for form.object.model.fetch(:work).depositor.last_work_terms_agreement
+      and/or determine why we have to check it separately to avoid potential nil error if embargo release date is invalid (e.g. June 31) %>
+      <% if agree_to_terms? && form.object.model.fetch(:work).depositor.last_work_terms_agreement %>
         You agreed to the <%= link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'} %>
         <%= time_ago_in_words form.object.model.fetch(:work).depositor.last_work_terms_agreement %> ago.
       <% else %>
--- a/app/javascript/controllers/available_date_controller.js
+++ b/app/javascript/controllers/available_date_controller.js
@@ -99,7 +99,7 @@ export default class extends Controller {
     const year = this.yearTarget.value
 
     if (isNaN(new Date(`${year}-${month}-${day}`))) {
-      this.errors.day = 'this day not in selected month'
+      this.errors.day = 'day must exist in selected month'
     }
   }
 }
--- a/app/javascript/controllers/available_date_controller.js
+++ b/app/javascript/controllers/available_date_controller.js
@@ -7,6 +7,7 @@ export default class extends Controller {
     this.errors = {}
     this.dateInFuture()
     this.allPartsPresent()
+    this.validDate()
     this.displayErrors()
   }
 
@@ -88,6 +89,17 @@ export default class extends Controller {
       this.errors.month = 'all parts must be provided'
     } else if (!day) {
       this.errors.day = 'all parts must be provided'
+    }
+  }
+
+  // Validate that the day selected is in the month
+  validDate() {
+    const day = this.dayTarget.value
+    const month = this.monthTarget.value
+    const year = this.yearTarget.value
+
+    if (isNaN(new Date(`${year}-${month}-${day}`))) {
+      this.errors.day = 'this day not in selected month'
     }
   }
 }
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -42,7 +42,7 @@ class User < ApplicationRecord
   end
 
   def agreed_to_terms_recently?
-    return false unless last_work_terms_agreement
+    return false if last_work_terms_agreement.nil?
 
     terms_agreement_renewal_timeframe < last_work_terms_agreement
   end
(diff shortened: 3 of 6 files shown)


app/components/works/agreement_component.html.erb
```diff
@@ -2,7 +2,7 @@
   <header>Terms of Deposit *</header>
   <div class="row">
     <div class="form-check col-auto">
-      <% if agree_to_terms? %>
+      <% if agree_to_terms? && form.object.model.fetch(:work).depositor.last_work_terms_agreement %>
```
ndushay, line 5: this will avoid the nil error in the app.  It's possible there's a better way, but this gets the job done.  (It's not just good, it's good enough.)

app/javascript/controllers/available_date_controller.js
```diff
@@ -90,4 +91,15 @@ export default class extends Controller {
...
+
+  // Validate that the day selected is in the month
+  validDate() {
+    const day = this.dayTarget.value
+    const month = this.monthTarget.value
+    const year = this.yearTarget.value
+
+    if (isNaN(new Date(`${year}-${month}-${day}`))) {
+      this.errors.day = 'day must exist in selected month'
+    }
```
ndushay, line 103: oddly, this seemed to be needed as well;  we only triggered the nil error when we had a user with a nil "last_work_terms_agreement" AND an invalid or incomplete embargo release date.

app/models/user.rb
```diff
@@ -42,7 +42,7 @@ def to_honeybadger_context
   end
 
   def agreed_to_terms_recently?
-    return false unless last_work_terms_agreement
+    return false if last_work_terms_agreement.nil?
```
ndushay, line 45: I find this easier to grok.

app/validators/embargo_date_parts.rb
```diff
@@ -6,7 +6,7 @@ def validate(record)
     # If we've already successfully deserialized the date, there's no point to continuing
     return if record.embargo_date.instance_of?(Date)
 
-    # Adds error because embargo date is nil (may be missing month and day values in form)
-    record.errors.add(:embargo_date, 'must provide all parts')
+    # Adds error because embargo date is nil or invalid (may be missing month and day values in form)
+    record.errors.add(:embargo_date, 'must provide all date parts and must be a valid date')
```
ndushay, line 10: changed the error message to be more complete and correct

ndushay:
@amyehodge I would like your signoff on this;  I can deploy it to stage and set your user to have no approval of work agreement so you can take it for a drive.  Peter could also do this.

app/javascript/controllers/available_date_controller.js
peetucket: this gives a nice friendly error message to the user on the page if they select a bogus date (e.g. June 31) before they even submit the form

app/components/works/agreement_component.html.erb
peetucket: ugh, despite my better judgement i want to spend more time trying to understand why it doesn't work the original way ... though agreed this will at least avoid the nasty exception (maybe drop a comment in there pointing to this ticket so if we don't fix it now, we will know why it was done later)

peetucket:
Looks fine pending PO review.  Suggest a name change for the PR to reflect what it does now.

app/components/works/agreement_component.html.erb
ndushay: psst - peter - let it go.  :-)

jcoyne, line 5: Can you move this from the template to the class?

jcoyne (changes requested):

amyehodge:
@ndushay I'm not seeing the last update to the error message. So, when I enter an invalid date, I see this:

Screen Shot 2022-05-06 at 11 27 28 AM

It says I've agreed to the terms 9 months ago, so I'm not sure that I'm able to test that piece of the interaction.

app/components/works/agreement_component.html.erb
jcoyne: I think this can be folded right into the `agree_to_terms` method as I don't see any other callers.

amyehodge:
This looks good to me. Okay to merge!

app/components/works/agreement_component.html.erb
ndushay: Honestly, it was such a chore gettig the dang thing to work, and with it RIGHT BELOW on line 7, I'm gonna say no to this.  You can make a ticket for it if you like.

4 of us tried to figure out how `agree_to_terms?` could be true here and `form.object.model.fetch(:work).depositor.last_work_terms_agreement` could be nil ... but empirically it was so.

I also don't know why this problem was only triggered when the selected embargo release date was an invalid date ... but that was also true.

ndushay:
See also #2243

app/components/works/agreement_component.html.erb
ndushay: See also #2243.

jcoyne: I feel like you are contravening our team norms (https://github.com/sul-dlss/DeveloperPlaybook/blob/main/best-practices/testing.md#why) by not including a meaningful test with this change. This has been a problem for a long time, so let's not rush to push something out that's not ready.

jcoyne (approved):

PR #2246, opened 2022-05-06
check the depositor agreement directly instead of going through the form

### LINKED ISSUE #2243: create a test for fix in PR #2241
There should be a feature test or some such for the code changed in PR #2241

1. showing the text in the embargo release date when an invalid date is chosen
2. showing that the error block appears at the top of the page when the user has NOT agreed to the terms of deposit AND there is an invalid date.

peetucket:

Based on work done in #2241 and #2242, this fixes the underlying cause of the bug identified in #2235.

Also 

Fixes #2243 (add a new test for this edge case)

The issue is that the `agree_to_terms` property on the work form was overloaded and could refer to the value of the existing HTML form checkbox as well as the custom derived property we were hoping to use (which did some lookup on the user in the database).  So in cases where the form was re-rendered with an error message and the value of the HTML checkbox was unchecked, the `agree_to_terms` property was returning a "0" instead of `false`, triggering our exception.  The fix is to just bypass the form property and delegate this check straight to the user model, which is what we were originally trying to do anyway.  This lets us avoid the extra check in the component view.

Localhost browser and new test

### DIFF
--- a/app/components/works/agreement_component.rb
+++ b/app/components/works/agreement_component.rb
@@ -7,10 +7,6 @@ module Works
       @form = form
     end
 
-    def agree_to_terms?
-      work_form.agree_to_terms
-    end
-
     def work_form
       form.object
     end
@@ -20,7 +16,7 @@ module Works
     end
 
     delegate :depositor, to: :work
-    delegate :last_work_terms_agreement, to: :depositor
+    delegate :agreed_to_terms_recently?, :last_work_terms_agreement, to: :depositor
 
     attr_reader :form
   end
--- a/app/components/works/agreement_component.html.erb
+++ b/app/components/works/agreement_component.html.erb
@@ -2,9 +2,7 @@
   <header>Terms of Deposit *</header>
   <div class="row">
     <div class="form-check col-auto">
-      <%# TODO: create a method for form.object.model.fetch(:work).depositor.last_work_terms_agreement
-      and/or determine why we have to check it separately to avoid potential nil error if embargo release date is invalid (e.g. June 31) %>
-      <% if agree_to_terms? && form.object.model.fetch(:work).depositor.last_work_terms_agreement %>
+      <% if agreed_to_terms_recently?  %>
         You agreed to the <%= link_to 'SDR Terms of Deposit', '#termsOfDepositModal', data: { bs_toggle: 'modal', bs_target: '#termsOfDepositModal'} %>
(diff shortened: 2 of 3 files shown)


app/components/works/agreement_component.rb
```diff
@@ -20,7 +16,7 @@ def work
     end
 
     delegate :depositor, to: :work
-    delegate :last_work_terms_agreement, to: :depositor
+    delegate :agreed_to_terms_recently?, :last_work_terms_agreement, to: :depositor
```
peetucket, line 20: this goes straight to the underlying method on the user model, which is what we want anyway

ndushay: Were you able to prove that this change, with all the conditions to trigger the exception previous, keeps that exception from being triggered?

Or are you expecting the not-yet-written test to do it?

It would be great if you wanted to write the test as part of the PR ...

app/components/works/agreement_component.rb
```diff
@@ -7,10 +7,6 @@ def initialize(form:)
       @form = form
     end
 
-    def agree_to_terms?
-      work_form.agree_to_terms
-    end
-
     def work_form
       form.object
     end
```
ndushay, line 12: do we still use work_form?  I'm guessing it can be removed?

ndushay (changes requested):
see also my single comment.

app/components/works/agreement_component.rb
peetucket: the `work_form` appears to be used to fetch the `work`, which is then used to get to the depositor (some lines not visible in the diff)

app/components/works/agreement_component.rb
peetucket: I was able to prove this change didn't blow up in browser.  I would suggest we leave it uncoupled from the ticket for the added test (though I am happy to have a look at that next).

peetucket: Or not...i guess this fix isn't an urgent bug fix, so either way is fine.

peetucket:
Note: I have verified the new spec in this PR catches the edge case and fails with the code prior to the bug fix in #2241, but passes with that bug fix and with this PR.

ndushay (approved):

jcoyne:
Thanks @peetucket. I really like this change. 👍🏼

PR #2254, opened 2022-05-25
Add activeadmin gem

jcoyne:


jcoyne:
It feels like using activeadmin might be more trouble than it's worth, given that it currently depends on sassc and we don't use that here.  Additionally, it doesn't support Turbo, which is an important technolgy for h2.

PR #2259, opened 2022-05-26
start with blank version description for new versions

### LINKED ISSUE #2256: When creating a new version of a collection, the "What changed?" field should be blank
Currently, when I open a new version of a collection that has been previously versioned, the comments made for the previous version display. This would allow the user to deposit a new version with version comments that reflect the last change without updating to reflect what happened with this change. And they might not realize they've done it. This box should be blank when opening a new version.

Example. Here is the info on this collection:

Screen Shot 2022-05-26 at 2 52 52 PM

When I click on the pencil icon to edit, the "What's changing?" field is already filled in with the old comment:
Screen Shot 2022-05-26 at 2 53 00 PM

peetucket:

Fixes #2256 - start from blank description for new collection versions

also, rename some variables to be more clear

Localhost and added/updated tests

### DIFF
--- a/app/controllers/collection_versions_controller.rb
+++ b/app/controllers/collection_versions_controller.rb
@@ -69,9 +69,9 @@ class CollectionVersionsController < ObjectsController
 
   # Create the next CollectionVersion for this Collection
   def create_new_version(previous_version)
-    previous_version.dup.tap do |work_version|
-      work_version.state = 'version_draft'
-      work_version.version = previous_version.version + 1
+    previous_version.dup.tap do |collection_version|
+      collection_version.state = 'version_draft'
+      collection_version.version = previous_version.version + 1
     end
   end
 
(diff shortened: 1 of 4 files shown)


app/controllers/collection_versions_controller.rb
```diff
@@ -33,6 +33,9 @@ def edit
     # if we end up on the edit page for a first draft (non-deposited collection), redirect to first draft edit page
     redirect_to edit_first_draft_collection_path(collection_version.collection) if collection_version.first_draft?
 
+    # start with a blank version description for a new version (pre-fill for saved as draft)
+    collection_version.version_description = '' if collection_version.deposited?
```
jcoyne, line 37: Can this go in the `create_new_version` method?

jcoyne (changes requested):

app/controllers/collection_versions_controller.rb
peetucket: I don't think so, because we need the form to render with blank value (and it is rendered in `edit`, which will fetch the previous verison's description).

peetucket: Tested moving to the `create_new_version` method and it fails the new test I wrote.

jcoyne: The test is for a situation that would never arise if you move it into create_new_version

jcoyne (approved):

PR #2265, opened 2022-06-02
sanitize abstract

peetucket:

Potential suggestion to deal with https://app.honeybadger.io/projects/77112/faults/85827019, user pasted in text from a PDF into the abstract field, which breaks in postgres upon submission.  Seems like there should be a more uniform way of doing this.

Localhost

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -14,6 +14,8 @@ class WorkVersion < ApplicationRecord
     # So this seems to take care of it.
     AbstractContributor.where(work_version_id: id).destroy_all
   end
+  before_save :clean_abstract
+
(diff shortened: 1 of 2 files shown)


peetucket:
1. Added inline comment referencing HB and why we are sanitizing.
2. Added rails log when strip occurs

app/models/work_version.rb
```diff
@@ -98,6 +100,16 @@ class WorkVersion < ApplicationRecord
     end
   end
 
+  # 6/3/2022 : Added to prevent https://app.honeybadger.io/projects/77112/faults/85827019
```
jcoyne, line 103: Honeybadger URLs expire, so I was hoping you'd link somewhere on the web that describes the "String contains null bytes" problem.

PR #2268, opened 2022-06-06
Change relationship declaration to contributor model to match author model

peetucket:

~~Another attempt to fix https://github.com/sul-dlss/happy-heron/issues/2240 which causes https://app.honeybadger.io/projects/77112/faults/85307350 .... basically a contributor model being saved before the associated work version is saved, throwing a null violation constraint.~~

This change just makes the relationship declaration on both associated models match.  `inverse_of` should be inferred for authors as it is for contributors.

Existing tests.

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -6,7 +6,7 @@ class WorkVersion < ApplicationRecord
 
   belongs_to :work
   has_many :contributors, dependent: :destroy, class_name: 'Contributor'
-  has_many :authors, -> { order(weight: :asc) }, inverse_of: :work_version, dependent: :destroy, class_name: 'Author'
+  has_many :authors, -> { order(weight: :asc) }, dependent: :destroy, class_name: 'Author'
   before_destroy do
     # Unfortunately the STI relationships above, don't delete everything.
     # I first tried this approach, but it didn't work either


jcoyne:
@peetucket I don't understand how defining this relationship explicitly would impact the order in which things are saved in. It already is able to derive this relationship implicitly and none of that logic governs save order.

app/forms/draft_work_form.rb
```diff
@@ -103,7 +103,7 @@ def deserialize_license(params)
              populator: ContributorPopulator.new(:authors, Author),
              prepopulator: ->(*) { authors << Author.new if authors.blank? },
              on: :work_version,
-             &contributor
+             &author
```
peetucket, line 107: this doesn't seem right to me, the `authors` collection references the contributor -- this would fix it

app/models/work_version.rb
```diff
@@ -5,7 +5,7 @@ class WorkVersion < ApplicationRecord
   include AggregateAssociations
 
   belongs_to :work
-  has_many :contributors, dependent: :destroy, class_name: 'Contributor'
+  has_many :contributors, inverse_of: :work_version, dependent: :destroy, class_name: 'Contributor'
```
jcoyne, line 8: This is something that rails can infer without overriding the default:
```ruby
WorkVersion.reflect_on_association(:contributors).inverse_of
=>
#<ActiveRecord::Reflection::BelongsToReflection:0x000000011249a300
 @active_record=
  AbstractContributor(id: integer, work_version_id: integer, first_name: string, last_name: string, created_at: datetime, updated_at: datetime, contributor_type: string, role: string, full_name: string, type: string, weight: integer, orcid: string),
 @class_name="WorkVersion",
 @foreign_key="work_version_id",
 @inverse_name=nil,
 @klass=
  WorkVersion(id: integer, version: integer, title: string, work_type: string, created_edtf: string, abstract: text, citation: string, access: enum, embargo_date: date, license: string, created_at: datetime, updated_at: datetime, state: string, published_edtf: string, subtype: text, work_id: integer, description: string),
 @name=:work_version,
 @options={},
 @plural_name="work_versions",
 @scope=nil>
```

jcoyne (changes requested):

app/models/work_version.rb
peetucket: Do you know if is there a reason we have it on the `has_many :authors` declaration on the very next line?  Maybe we should remove it there so at least they are consistent.

peetucket:
Interesting -- rubocop doesn't like the missing `inverse_of` for the Author declaration.  https://www.rubydoc.info/gems/rubocop/0.61.1/RuboCop/Cop/Rails/InverseOf  Apparently using some options prevents Rails from inferring it automatically.  I guess we leave it in.

PR #2380, opened 2022-08-03
Embargo is considered released on today's date

### LINKED ISSUE #2366: Prevent error when user attempts to set embargo date for the future on the day that the embargo has already been lifted.
HB error: https://app.honeybadger.io/projects/77112/faults/80837103
There is a previous discussion on this error in the #dlss-infrastructure slack channel if you search for that link.

I have successfully recreated this error. In a nutshell, it seems to me like there is confusion when modifications are made on **the day the embargo is released** because the form still displays the embargo date instead of showing that it has been released. Today's date won't validate, because it's not in the future, but you can't change this to a future date because the item has already been released. Attempting to do so will trigger the error message.

Here are the steps I took. 

- I created new item https://sdr.stanford.edu/works/5764 (druid:wn318vm7113) on 7-25-2022 and assigned it an embargo date of 7-26-2022. 
- On 7-26-2022, I checked to see that the embargo had been lifted. The history indicated that it had been. 
Screen Shot 2022-07-26 at 10 10 01 AM
- I then clicked the pencil icon to edit the item. I entered a note in the "What's changing?" box at the top, then modified the abstract. Note that the section for the embargo showed that the embargo date was today, though the history said the embargo had been lifted:
Screen Shot 2022-07-26 at 10 03 01 AM
- When I clicked "Deposit", I got the same message as the user below -- the validations ran and it told me that the embargo date must be in the future. 
- I then changed the embargo date to **tomorrow**, and then clicked "Deposit" again. 
- I then got the blank page with the "That wasn't supposed to happen" error message. 
- I think I then clicked back on the browser (maybe I refreshed? I don't think this is critical) and the view page for the item reloaded.
- I clicked the pencil icon again, made a note at the top about the changes, then changed the embargo option to "immediate" release.
- I clicked "deposit" and it completed just fine.
- When I re-open the form the release information displays like this and the user can't make changes:
Screen Shot 2022-07-26 at 10 15 53 AM

lwrubel:

Fixes #2366. Shows embargo as already released if the embargo date is today. 

Unit

### DIFF
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -43,7 +43,7 @@ class Work < ApplicationRecord
   end
 
   def already_embargo_released?
-    head&.deposited? && head.embargo_date.present? && head.embargo_date < Time.zone.today
+    head&.deposited? && head.embargo_date.present? && head.embargo_date <= Time.zone.today
   end
 
   delegate :name, to: :depositor, prefix: true
--- a/spec/components/works/embargo_component_spec.rb
+++ b/spec/components/works/embargo_component_spec.rb
@@ -84,6 +84,14 @@ RSpec.describe Works::EmbargoComponent do
     end
   end
 
+  context 'when the collection has been deposited and embargo date is today' do
+    let(:work_version) { build(:work_version, work: work, state: 'deposited', embargo_date: Time.zone.today) }
+
+    it 'renders the component' do
+      expect(rendered.to_html).to include 'This item has been released from embargo.'
+    end
+  end
+
   context 'when the collection has been deposited and embargo has not elapsed' do
     let(:collection) { build(:collection, :depositor_selects_access, release_option: 'depositor-selects') }
     let(:work_version) { build(:work_version, work: work, state: 'deposited', embargo_date: Time.zone.today + 1.month) }


mjgiarlo (approved):

lwrubel:
@mjgiarlo I could use some advice on this. I see that the method is already covered (via CodeClimate) but am not sure where specifically in the tests. (This method at least, is not covered on its own in the existing specs.) Would it be useful to add a test of a work_version with today's date as the embargo date in this existing spec? 

https://github.com/sul-dlss/happy-heron/blob/c75193679f6f1e418d3eab9ee0a46595bbd132bc/spec/components/works/embargo_component_spec.rb#L79

mjgiarlo:
> @mjgiarlo I could use some advice on this. I see that the method is already covered (via CodeClimate) but am not sure where specifically in the tests. (This method at least, is not covered on its own in the existing specs.) Would it be useful to add a test of a work_version with today's date as the embargo date in this existing spec?
> 
> https://github.com/sul-dlss/happy-heron/blob/c75193679f6f1e418d3eab9ee0a46595bbd132bc/spec/components/works/embargo_component_spec.rb#L79

That sounds good to me 😄

mjgiarlo (approved):

lwrubel:
Thanks, added.

PR #2403, opened 2022-08-12
Default years to the max value

### LINKED ISSUE #2357: Year fields for publication date and creation date should not display "1000" when arrow buttons are clicked
All four of the year fields in this section display "1000" when the user clicks on the up or down arrows. I believe this is the minimum value set for this field, and if I recall correctly, a minimum value is required by the field.

Screen Shot 2022-07-08 at 4 25 28 PM

Users have provided us with feedback that this is not helpful and means that they end up having to erase and type in a value, since using the arrows once it says "1000" is not efficient.

Today I identified a published item where "1000" was listed inappropriately as the publication date, likely because of this weird interaction.

Ideally these should display the current date and the user can use the arrows to move backward in time from here (or forward again). If that's not possible, a better solution is to have this box be empty and not display the minimum. If for some reason having the arrows is what makes this have to display the minimum, then we should remove them entirely and default to just an empty box that the user has to type into.

jcoyne:

When the up button is first pressed, set the value to the max value of the field rather than the minimum (default).

Fixes #2357

Tested locally.

### DIFF
--- a/app/javascript/controllers/year_field_controller.js
+++ b/app/javascript/controllers/year_field_controller.js
@@ -1,0 +1,20 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+//   static targets = ["year", "month", "day", "error"]
+
+  connect() {
+    this.hasChanged = false
+  }
+
+  // On the first change, if they press the arrow keys or arrow buttons (on the page),
+  // Then default it to the max value.  The default behavior is to default to the minimum value.
+  change(evt) {
+    if (this.hasChanged)
+        return // nothing to do
+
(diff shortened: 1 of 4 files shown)


ndushay (approved):

app/javascript/controllers/year_field_controller.js
```diff
@@ -0,0 +1,20 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+//   static targets = ["year", "month", "day", "error"]
```
justinlittman, line 4: cruft

PR #2408, opened 2022-08-12
Store the publication date on a work version when it is published

### LINKED ISSUE #2404: Add a published date field to work_version
This should be populated during the "begin_deposit" event. See https://github.com/sul-dlss/happy-heron/wiki/State-machine-diagrams

jcoyne:
Fixes #2404

### DIFF
--- a/db/migrate/20220812205204_add_publication_date_to_work_version.rb
+++ b/db/migrate/20220812205204_add_publication_date_to_work_version.rb
@@ -1,0 +1,7 @@
+class AddPublicationDateToWorkVersion < ActiveRecord::Migration[7.0]
+  def change
+    change_table :work_versions do |t|
+      t.datetime :published_at
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20220812205204_add_publication_date_to_work_version.rb
```diff
@@ -0,0 +1,7 @@
+class AddPublicationDateToWorkVersion < ActiveRecord::Migration[7.0]
+  def change
+    change_table :work_versions do |t|
+      t.datetime :published_at
```
justinlittman, line 4: Do we want to pre-populate for existing work versions?

justinlittman (approved):

db/migrate/20220812205204_add_publication_date_to_work_version.rb
jcoyne: I don't think so. We never will re-publish them so I don't see a reason to do so.

Issue #2384, opened 2022-08-03
Fix flaky test


jcoyne:
The spec/features/mediated_deposit_versioning_spec.rb often fails on CI. Figure out why and fix the problem

justinlittman:
Here's an example: https://app.circleci.com/pipelines/github/sul-dlss/happy-heron/4190/workflows/03ea1005-5a66-4b7b-b1f4-8282d60982da/jobs/15847

justinlittman:
When it fails:
![screenshot_2022-08-15-05-15-22 830](https://user-images.githubusercontent.com/588335/184636295-7ea4cc32-b617-4f30-b3fb-7c871c9825eb.png)

justinlittman:
When it doesn't fail:
![screenshot](https://user-images.githubusercontent.com/588335/184636330-2965f49a-08d5-45dc-8856-96ce0d0c610d.png)

justinlittman:
The failure occurs in the code right after switching from depositor to reviewer. When it fails, it appears that the user is NOT signed in as the reviewer.

I will try adding an explicit sign out to see if it helps.

justinlittman:
![image.png](https://images.zenhubusercontent.com/5d8270f64def500001d12f23/d2580ae4-0073-4963-9755-4f53cca57eed)

PR #2410, opened 2022-08-15
Adds sign out to flaky features/mediated_deposit_versioning_spec.

justinlittman:
refs #2384

Flaky bad.

### DIFF
--- a/spec/features/mediated_deposit_versioning_spec.rb
+++ b/spec/features/mediated_deposit_versioning_spec.rb
@@ -45,6 +45,9 @@ RSpec.describe 'Edit a new version of a work in a collection using mediated depo
 
       # A work submitted for approval should not be editable.
       expect(page).not_to have_css("a[aria-label='Edit #{new_work_title}']", wait: 0)
+      sign_out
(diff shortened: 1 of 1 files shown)


spec/features/mediated_deposit_versioning_spec.rb
```diff
@@ -45,6 +45,8 @@
 
       # A work submitted for approval should not be editable.
       expect(page).not_to have_css("a[aria-label='Edit #{new_work_title}']", wait: 0)
+      sign_out
+      sleep(1)
```
justinlittman, line 49: Just `sign_out` didn't work, so added sleep.

jcoyne: Can you add a comment about why the sleep is necessary? Otherwise I foresee someone removing this for lack of understanding why it was here.

jcoyne (changes requested):

spec/features/mediated_deposit_versioning_spec.rb
justinlittman: Done.

jcoyne (approved):

PR #2411, opened 2022-08-15
Sets user for Deposited event to SDR.

### LINKED ISSUE #2402: Record the name of the user depositing collection in the history table
I edited the metadata on this collection and it shows my name for the "updated by user" and "submitted for deposit" lines but then inserted the name of the collection creator for "deposit completed." This should have had my SUNet ID as I was the person who was making the changes.

This was on https://sdr.stanford.edu/collections/199

Screen Shot 2022-08-10 at 12 46 50 PM

justinlittman:
closes #2402

Correctly display who/what is performing event.

Unit

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ bundle exec byebug -R localhost:8989
 
 Note that, by default, the debugger will run on `localhost` on port `8989`. To change these values, add the `DEBUGGER_HOST` and `DEBUGGER_PORT` environment variables when invoking `bin/dev` above and make sure to poing `byebug -R` at those same values when starting up the debugger.
 
-Start up dependencies with `docker compose up db` (with `-d` to run in background), create and migrate the database with `bundle exec rake db:create db:migrate` and then run tests with `bundle exec rspec`. If you also want to do style checks & linting, run Rubocop and RSpec serially via `bin/rake`.
+Start up dependencies with `docker compose up db` (with `-d` to run in background), create and migrate the database with `bundle exec rake db:prepare` and then run tests with `bundle exec rspec`. If you also want to do style checks & linting, run Rubocop and RSpec serially via `bin/rake`.
 
 ### Integration
 
--- a/app/jobs/deposit_status_job.rb
+++ b/app/jobs/deposit_status_job.rb
@@ -13,6 +13,7 @@ class DepositStatusJob
   # these messages in addition to those that result from depositing in h2.
   from_queue 'h2.deposit_complete', env: nil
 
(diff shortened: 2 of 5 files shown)


justinlittman:
SDR user has been created in all environments.

README.md
```diff
@@ -55,7 +55,7 @@ bundle exec byebug -R localhost:8989
 
 Note that, by default, the debugger will run on `localhost` on port `8989`. To change these values, add the `DEBUGGER_HOST` and `DEBUGGER_PORT` environment variables when invoking `bin/dev` above and make sure to poing `byebug -R` at those same values when starting up the debugger.
 
-Start up dependencies with `docker compose up db` (with `-d` to run in background), create and migrate the database with `bundle exec rake db:create db:migrate` and then run tests with `bundle exec rspec`. If you also want to do style checks & linting, run Rubocop and RSpec serially via `bin/rake`.
+Start up dependencies with `docker compose up db` (with `-d` to run in background), create and migrate the database with `bundle exec rake db:create db:migrate db:seed` and then run tests with `bundle exec rspec`. If you also want to do style checks & linting, run Rubocop and RSpec serially via `bin/rake`.
```
jcoyne, line 58: We can just do `bin/rake db:prepare`.  It does all of these and cuts down on carpal tunnel as well.

app/jobs/deposit_status_job.rb
```diff
@@ -40,4 +44,12 @@ def parse_message(msg)
 
     raise "Unable to find required field 'druid' in payload:\n\t#{json}"
   end
+
+  def parent(object)
+    object.head.try(:collection) || object.head.work
```
jcoyne, line 49: why is this necessary?  Isn't this just going from parent to child back to parent?

jcoyne (changes requested):

README.md
justinlittman: Fixed.

app/jobs/deposit_status_job.rb
justinlittman: That's a great question. If you set `event_context` on `object` directly, `object.head.deposit_complete!` does not get the `event_context`.

jcoyne: Can you add a comment about that here? Something like, "although both objects represent the same row in the database, they are not the same object in memory"

justinlittman: Comment added.

jcoyne (approved):

PR #2415, opened 2022-08-16
Moves All Collections table to Admin

### LINKED ISSUE #2307: Move existing All Collections table to new Admin dashboard page
This table and search functionality should be moved from the dashboard page you see when you log in and added to the Admin dashboard page. They should no longer be on the main dashboard page.

Screen Shot 2022-06-10 at 4 23 54 PM

The All Collections table should appear exactly as it is now at the bottom of the new page and be collapsed by default. 

Mockup for All Collections table: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/467473848

lwrubel:
Resolves #2307 to move All Collections table to Admin dashboard.

Unit, added tests, QA.

Screen Shot 2022-08-16 at 11 16 42 AM


app/assets/stylesheets/dashboard.scss
```diff
@@ -27,6 +27,10 @@
     font-style: italic;
   }
 
+  #admin-dropdown {
```
jcoyne, line 30: Rather than add a custom style here, can you add the bootstrap class `mb-4` (or similar) to the element's markup?

app/controllers/admins_controller.rb
```diff
@@ -7,5 +7,14 @@ class AdminsController < ApplicationController
 
   def show
     authorize! :admin_dashboard
+    @presenter = build_presenter
+
+    @presenter.work_stats = StatBuilder.build_stats if user_with_groups.administrator?
```
jcoyne, line 12: Is this conditional necessary?  Aren't all users of this route administrators?

app/controllers/admins_controller.rb
```diff
@@ -7,5 +7,14 @@ class AdminsController < ApplicationController
...
+  end
+
+  private
+
+  def build_presenter
```
jcoyne, line 17: I'm wondering if having this as a separate method is useful.  We're not encapsulating a lot of complexity or reusing this, so I'd say it's easier to read the code without it.

app/presenters/admin_presenter.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+# The page model for the admin page
+class AdminPresenter
+  attr_accessor :work_stats
```
jcoyne, line 5: Can the work stats be set as part of the initializer?

jcoyne (changes requested):

app/assets/stylesheets/dashboard.scss
lwrubel: Oh duh, yes, will do.

app/controllers/admins_controller.rb
lwrubel: That's true, got it.

app/controllers/admins_controller.rb
lwrubel: Do we anticipate needing to build out the presenter for the admin dashboard as more functionality is added there? That was the reason I kept using the `build_presenter` method. But I'm fine to remove if we're not needing that for now.

jcoyne: I'd remove it until we need it. It's really not much work to create a method.

app/presenters/admin_presenter.rb
lwrubel: Done.

lwrubel:
I think I've addressed the comments but let me know if you see anything else to improve.

app/components/admins/all_collections_component.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module Dashboard
+module Admins
```
jcoyne, line 3: Can this be `Admin` (singular) instead?

app/components/admins/all_collections_row_component.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module Dashboard
+module Admins
```
jcoyne, line 3: Same here.

jcoyne (changes requested):

app/components/admins/all_collections_component.rb
lwrubel: Updated here and in other filepaths in order to align with https://github.com/sul-dlss/happy-heron/pull/2421

lwrubel:
Rebased.

jcoyne (approved):
Thank you @lwrubel

jcoyne (commented):
It's getting really close. I just have one more question about testing.

jcoyne (changes requested):
It's getting really close, I just  have one question about testing.

spec/requests/dashboards_spec.rb
```diff
@@ -130,44 +130,13 @@
...
-      work6.update(head: work_version6)
-
       sign_in user, groups: ['dlss:hydrus-app-administrators']
     end
 
-    it 'shows a link to create collections, all collections table, and admin' do
+    it 'shows a link to admin' do
       get '/dashboard'
       expect(response).to have_http_status(:ok)
-      expect(response.body).to include 'Your collections'
```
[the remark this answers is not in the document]
lwrubel: I think I was too aggressive in stripping this down to testing the admin link. Added those expectations back.

jcoyne (approved):

PR #2417, opened 2022-08-16
Adds deposit date into cocina description.

### LINKED ISSUE #2368: Insert deposit date into description when each version is published
The problem: When a user does not enter a date in H2, no date is displayed in the metadata or indexed for the object.

The date of initial deposit should be entered into the descriptive metadata, both to record this information and to provide a fallback for discovery if the user does not enter a date. This date should not change when the object is versioned unless the user enters a publication date later.

The date of initial deposit should also be retained as the record creation date, and should not change if the object is versioned in future.

Example if deposit date is January 29, 2022 and the object was updated on January 31 and February 15; the user entered a publication date of November 30, 2021:
```
{
    "event": [
        {
            "type": "deposit",
            "date": [
                {
                    "type": "publication",
                    "value": "2022-01-29",
                    "encoding": {"code": "edtf"}
                }
            ]
        },
        {
            "type": "deposit",
            "date": [
                {
                    "type": "modification",
                    "value": "2022-01-31",
                    "encoding": {"code": "edtf"}
                }
            ]
        },
        {
            "type": "deposit",
            "date": [
                {
                    "type": "modification",
                    "value": "2022-02-15",
                    "encoding": {"code": "edtf"}
                }
            ]
        },
        {
            "type": "publication",
            "date": [
                {
                    "type": "publication",
                    "value": "2021-11-30",
                    "status": "primary",
                    "encoding": {"code": "edtf"}
                }
            ]
        }
    ],
    "adminMetadata": {
        "event": [
            {
                "type": "creation",
                "date": [
                    {
                        "value": "2022-01-29",
                        "encoding": {"code": "edtf"}
                    }
                ]
            }
        ]
    }
}
```
Dependent on https://github.com/sul-dlss/cocina-models/issues/502 for mapping spec.

Implementation details:
- Create a new timestamp on each version that is set during the `begin_deposit` event
- List all of version timestamps as event type deposit, date type modification in Cocina (see mapping)
- List the first version's timestamp as event type deposit, date type publication (see mapping)
- Also add initial deposit date as record creation date in DescriptiveAdminMetadata

justinlittman:
closes #2368

So that item has complete descriptive metadata.

Unit

### DIFF
--- a/app/services/cocina_generator/description/generator.rb
+++ b/app/services/cocina_generator/description/generator.rb
@@ -19,7 +19,7 @@ module CocinaGenerator
           contributor: ContributorsGenerator.generate(work_version: work_version).presence,
           subject: keywords.presence,
           note: [abstract, citation].compact.presence,
-          event: generate_events.presence,
+          event: EventsGenerator.generate(work_version: work_version).presence,
           relatedResource: related_resources.presence,
           form: TypesGenerator.generate(work_version: work_version).presence,
           access: access,
@@ -41,15 +41,6 @@ module CocinaGenerator
         [
           Cocina::Models::Title.new(value: work_version.title)
         ]
-      end
-
-      def generate_events
-        pub_events = ContributorsGenerator.events_from_publisher_contributors(work_version: work_version,
-                                                                              pub_date: published_date)
-        return [created_date] + pub_events if pub_events.present? && created_date
-        return pub_events if pub_events.present? # and no created_date
-
-        [created_date, published_date].compact # no pub_events
       end
 
       def related_resources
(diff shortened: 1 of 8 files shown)


app/services/cocina_generator/description/generator.rb
```diff
@@ -19,7 +19,7 @@ def generate
           contributor: ContributorsGenerator.generate(work_version: work_version).presence,
           subject: keywords.presence,
           note: [abstract, citation].compact.presence,
-          event: generate_events.presence,
+          event: EventsGenerator.generate(work_version: work_version).presence,
```
jcoyne, line 22: 👏

app/services/cocina_generator/description/generator.rb
```diff
@@ -194,7 +127,10 @@ def admin_metadata
           ]
         )
       end
-      # rubocop:enable Metrics/MethodLength
+
+      def admin_metadata_creation_date
+        work_version.work.work_versions.first&.published_at || work_version.work.created_at
```
jcoyne, line 132: When you call `work_versions` is the order guaranteed?  I'm not sure it is.

jcoyne (changes requested):

app/services/cocina_generator/description/generator.rb
justinlittman: Wouldn't it be in primary key order by default?

jmartin-sul: I think that usually ends up being the case, but iirc, there's no guaranteed ordering by default for SQL queries if no order is explicitly specified.  So I think it's possible that the result wouldn't be ordered by primary key (my very gut feeling is that 90-ish% of the queries I've run against a PG/oracle/mysql have come back in PK order if no sort was specified -- but I also think I've seen occasional query results that break that expectation).

ha, of course i should've just gone to the PG manual instead of dredging my memory:
https://www.postgresql.org/docs/12/queries-order.html 💬 
> After a query has produced an output table (after the select list has been processed) it can optionally be sorted. If sorting is not chosen, the rows will be returned in an unspecified order. The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. A particular output ordering can only be guaranteed if the sort step is explicitly chosen.

justinlittman: Ordering now done https://github.com/sul-dlss/happy-heron/pull/2417/files#diff-ebe838f9d71d26b6227d6d37c96a7e2d93fb68a80d55075c7f2374bfcb4688abR14

It seems sensible to have this as default ordering for this association.

jcoyne (approved):

PR #2424, opened 2022-08-16
validates uniqueness of keywords server side only upon depositing

### LINKED ISSUE #1502: Notify user and prevent saving when duplicate keyword has been entered
Our keywords map to DOI subjects, which has a uniqueness requirement in the DataCite DOI API. See https://github.com/datacite/schema/blob/master/source/json/kernel-4.3/datacite_4.3_schema.json#L298.

If a user enters duplicate keywords, the DOI is not created, though a DOI displays in H2 as if it has been.

The request here is to identify duplicates via a string match against other entries on the current form so that users can delete one to prevent the DOI minting from running into trouble. Users should be able to save as draft, submit for approval, or deposit when there are duplicate keywords. Display error message and red box around field similar to other validation errors.

image

peetucket:

Fixes #1502 - ensure user cannot have duplicate keywords when making a final submission.  See screenshot below for message when user submits a deposit with two (or more) keywords being the same: 

Screen Shot 2022-08-16 at 4 31 38 PM

Note: this is currently only implemented server side, since that was relatively easy to do.  Not sure if we also want to add client side validation (using stimulus likely?).

Localhost browser and updated existing test.

### DIFF
--- a/app/components/works/keywords_row_component.html.erb
+++ b/app/components/works/keywords_row_component.html.erb
@@ -2,7 +2,7 @@
   <div class="col-sm-6">
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
     <div data-controller="autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
-      <%= form.text_field :label, { id: 'keyword-input-field', class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
+      <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
       <!-- This input is populated by autocomplete controller.-->
       <!-- That triggers the autocomplete-edit controller, which populates the uri and cocina type inputs. -->
       <input type="hidden" data-autocomplete-target="hidden" data-autocomplete-edit-target="value" data-action="autocomplete-edit#change" />
@@ -13,7 +13,7 @@
     </div>
   </div>
   <div class="col-sm-1">
-    <%= button_tag type: 'button', class: 'btn btn-sm', id: 'keyword-remove-button', aria: { label: 'Remove' },
+    <%= button_tag type: 'button', class: 'btn btn-sm', aria: { label: 'Remove' },
     data: { action: "click->nested-form#removeAssociation" } do %>
       <span class="far fa-trash-alt"></span>
     <% end %>
--- a/app/components/works/keywords_row_component.html.erb
+++ b/app/components/works/keywords_row_component.html.erb
@@ -2,7 +2,7 @@
   <div class="col-sm-6">
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
     <div data-controller="autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
-      <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
+      <%= form.text_field :label, { id: 'keyword-input-field', class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
       <!-- This input is populated by autocomplete controller.-->
       <!-- That triggers the autocomplete-edit controller, which populates the uri and cocina type inputs. -->
       <input type="hidden" data-autocomplete-target="hidden" data-autocomplete-edit-target="value" data-action="autocomplete-edit#change" />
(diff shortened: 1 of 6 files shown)


app/components/works/keywords_row_component.html.erb
```diff
@@ -13,7 +13,7 @@
     </div>
   </div>
   <div class="col-sm-1">
-    <%= button_tag type: 'button', class: 'btn btn-sm', aria: { label: 'Remove' },
+    <%= button_tag type: 'button', class: 'btn btn-sm', id: 'keyword-remove-button', aria: { label: 'Remove' },
     data: { action: "click->nested-form#removeAssociation" } do %>
       <span class="far fa-trash-alt"></span>
     <% end %>
```
peetucket, line 19: adding IDs to the keyword text box and remove button to facilitate testing

app/forms/work_form.rb
```diff
@@ -45,7 +52,7 @@ def deserialize_doi(params)
 
   delegate :already_immediately_released?, :already_embargo_released?, to: :work
 
-  # This is responsible for setting the DOI if the request one on a new version.
+  # This is responsible for setting the DOI if the user requests one on a new version.
   def maybe_assign_doi
```
peetucket, line 56: just fixing the text of an unrelated comment that i spotted

app/forms/work_form.rb
```diff
@@ -6,6 +6,7 @@
 class WorkForm < DraftWorkForm
   validates :abstract, :access, :title, presence: true, allow_nil: false
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
+  validate :keywords_unique
   validates :attached_files, length: { minimum: 1, message: 'Please add at least one file.' }
```
peetucket, line 10: Before doing this custom validator method, I played around with the `unique: true` declaration on the `validates :keywords` method (which is a Reform validator: https://github.com/trailblazer/reform-rails/blob/master/lib/reform/form/validation/unique_validator.rb) but couldn't get it to work as expected (either on it's own or with a scope).

spec/features/create_new_work_spec.rb
```diff
@@ -117,14 +117,25 @@
...
 
-          fill_in 'Keyword', with: 'Springs'
+          # add the same keyword twice
+          all('#keyword-input-field')[0].set 'Springs'
+          click_button '+ Add another keyword'
+          all('#keyword-input-field')[1].set 'Springs'
 
-          find('label.switch').click # Use auto-generated citation
-          fill_in 'Provided citation', with: 'Citation from user input'
+          click_button 'Deposit'
```
peetucket, line 133: just moved this citation setting block up a bit so that all the keyword checking code is grouped together...order doesn't matter in this case

spec/factories/keywords.rb
```diff
@@ -1,9 +1,11 @@
...
+require 'faker'
+
 FactoryBot.define do
   factory :keyword do
     work_version { nil }
-    label { 'MyKeyword' }
+    label { "#{Faker::Food.spice} #{Faker::Food.ingredient}" }
     uri { 'http://example.org/uri' }
     cocina_type { 'place' }
   end
```
peetucket, line 11: other tests that create multiple keywords using factories need to be unique now, or else they will fail with the new unique constraint we added

spec/features/create_new_work_spec.rb
```diff
@@ -117,14 +117,25 @@
           fill_in 'Abstract', with: 'User provided abstract'
           check 'Oral history'
 
+          find('label.switch').click # Use auto-generated citation
```
jcoyne, line 120: This test, which is supposed to test the feature "a user can create a work" is really stretching to test this validation.  I think this test is getting too complicated and I urge you to instead create a lighter weight request spec that tests just this single validation.  Then if it breaks, it's much easier to spot which part broke.

jcoyne (changes requested):

spec/requests/edit_work_spec.rb
```diff
@@ -144,24 +144,50 @@
...
+              license: 'CC0-1.0',
+              release: 'immediate'
+            }
+          end
+
+          it 'returns a validation error' do
+            patch "/works/#{work.id}", params: { work: work_params, commit: 'Deposit' }
+            expect(response).to have_http_status :unprocessable_entity
+            expect(response.body).to include 'Please ensure all keywords are unique'
+          end
```
peetucket, line 190: this code doesn't change the existing validation test about missing title, it just adds a new context block which then groups the new dupe keyword validation test with the existing title missing test

spec/features/create_new_work_spec.rb
peetucket: Ok, reverted change to this feature spec and move the validation test to a new request spec

spec/features/create_new_work_spec.rb
```diff
@@ -121,7 +121,7 @@
 
           expect(page).to have_content('Keyword must be filled in')
 
-          fill_in 'Keyword', with: 'Springs'
+          fill_in 'keyword-input-field', with: 'Springs'
```
jcoyne, line 124: Is there a reason we can't use the label here?  Using the label in the test helps to test that the form is accessible.

jcoyne (changes requested):

spec/features/create_new_work_spec.rb
peetucket: once i added the ids to the elements, it stopped finding it by the label:

```
 Capybara::ElementNotFound:
       Unable to find field "Keyword" that is not disabled
```

but now that i moved this test to a request spec, we may not need the ids on the input elements anymore anyway...let me back that out

jcoyne: I believe this was because the element already had an ID that was automatically generated. This id matched the label. When you added manual ids, it no longer aligned with the label.  Accessability was broken.

jcoyne (approved):

PR #2425, opened 2022-08-17
Displays owner on item detail page.

### LINKED ISSUE #2314: Change owner: On item view page add a row for "current owner"
On the item view page, if the ownership of the item has changed hands, include a line under the "Depositor" line with 

label: Current owner
content: SUNet ID (name) -- same as how the Depositor info is displayed.

Mockup: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/467473850

Screen Shot 2022-06-28 at 3 12 41 PM

All users should see this.

justinlittman:
closes #2314

Part of adding support for an owner being a different role than depositor.

Unit

### DIFF
--- a/app/components/works/detail_component.rb
+++ b/app/components/works/detail_component.rb
@@ -44,6 +44,14 @@ module Works
       "#{work.depositor.sunetid} (#{work.depositor.name})"
     end
 
+    def owner
+      "#{work.owner.sunetid} (#{work.owner.name})"
+    end
+
+    def show_owner?
+      work.owner != work.depositor
+    end
+
     def collection_name
       collection.head.name
     end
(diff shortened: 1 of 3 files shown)


app/components/works/detail_component.rb
```diff
@@ -44,6 +44,14 @@ def depositor
       "#{work.depositor.sunetid} (#{work.depositor.name})"
     end
 
+    def owner
+      "#{work.owner.sunetid} (#{work.owner.name})"
+    end
+
+    def owner?
```
jcoyne, line 51: Can this be `show_owner?` or `owner_different_from_depositor?`?

jcoyne (changes requested):

app/components/works/detail_component.rb
justinlittman: Yup. Switched to `show_owner?`

jcoyne (approved):

PR #2426, opened 2022-08-17
Loading seed into test db

lwrubel:

To load the required User seed into the test database. 

Running tests

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -41,6 +41,9 @@ rescue ActiveRecord::PendingMigrationError => e
   exit 1
 end
 
+# load required seed in test database
+Rails.application.load_seed
+
 WebMock.disable_net_connect!(allow_localhost: true)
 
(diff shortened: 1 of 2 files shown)


spec/rails_helper.rb
```diff
@@ -41,6 +41,7 @@
   exit 1
 end
 
+Rails.application.load_seed # loading seeds
```
jcoyne, line 44: Doesn't this cause an error the second time you run the suite?

lwrubel: Yes, I'm seeing it does in Circle CI but not locally.

lwrubel: Should we instead put into the README to run ` bundle exec RAILS_ENV=test rake db:seed`?

jcoyne: sure, but `RAILS_ENV=test bin/rails db:seed` is how I'd do it.

PR #2431, opened 2022-08-17
Change email address for sdr user.

justinlittman:
So that the sunetid is properly displayed.


jcoyne (approved):

justinlittman:
Fix was also made in all environments.

PR #2434, opened 2022-08-17
Display non-namespaced druids on the user search result

### LINKED ISSUE #2433: Remove "druid:" prefix from the "Druid" field in the results tables
For the results tables for the "Search for user" function, please remove the "druid:" prefix that displays in the "Druid" field. This will make these tables easier to read. Thanks.

jcoyne:
Fixes #2433 

### DIFF
--- a/spec/models/collection_spec.rb
+++ b/spec/models/collection_spec.rb
@@ -11,8 +11,11 @@ RSpec.describe Collection do
 
   describe '#purl' do
     context 'with a druid' do
+      before do
+        collection.druid = 'druid:hb093rg5848'
+      end
+
       it 'constructs purl' do
-        collection.update(druid: 'druid:hb093rg5848')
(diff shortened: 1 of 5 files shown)


spec/models/work_spec.rb
```diff
@@ -16,8 +16,11 @@
 
   describe '#purl' do
     context 'with a druid' do
+      before do
+        work.druid = 'druid:hb093rg5848'
```
jcoyne, line 20: This is a speed optimization, because we don't need to persist to the DB.

spec/models/collection_spec.rb
```diff
@@ -11,8 +11,11 @@
 
   describe '#purl' do
     context 'with a druid' do
+      before do
+        work.druid = 'druid:hb093rg5848'
```
jcoyne, line 15: This is a speed optimization, because we don't need to persist to the DB.

ndushay (approved):
looks good but tests are failing.

PR #2436, opened 2022-08-18
Adds admin collection report.

### LINKED ISSUE #2310: Create ability to generate collection report
This is one of the options in the drop-down menu on the Admin page that is available only to administrators.

Report should include the following information in csv format for all collections in all states in H2.

- collection title
- collection id 
- collection druid when available
- state of each collection (draft, deposited, version draft, etc.) 
- version number
- SUNet ID of user who created the collection
- SUNet IDs of all current collection managers
- date created
- date last modified
- date last deposited
- release setting
- visibility setting
- license
- DOI yes/no
- review workflow

Default sort by collection title.

We also request options to restrict the results based on the following:

- range of dates created
- range of dates last modified
- range of dates last deposited
- H2 state, selected via checkboxes, multiple boxes can be checked
-- Initial draft
-- Deposited
-- Version draft
-- Decommissioned

Mockup: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/467473849

Screen Shot 2022-06-28 at 3 18 03 PM

For the three date range options:
- require year, month, and day for start and end dates
- second date must be after the first date
Screen Shot 2022-06-20 at 2 24 36 PM

justinlittman:
closes #2310

Unit, Amy.

### DIFF
--- a/app/services/admin/collections_csv_generator.rb
+++ b/app/services/admin/collections_csv_generator.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+module Admin
+  # Converts a set of collections to CSV.
+  class CollectionsCsvGenerator
+    def self.generate(collections)
+      new(collections).generate
+    end
+
+    def initialize(collections)
+      @collections = collections
+    end
+
+    def generate
+      CSV.generate do |csv|
+        csv << HEADERS
+        collections.each do |collection|
+          csv << row(collection)
+        end
+      end
+    end
+
+    private
+
+    attr_reader :collections
+
+    HEADERS = [
+      'collection title',
+      'collection id',
+      'collection druid',
+      'state',
+      'version number',
+      'creator',
+      'managers',
+      'date created',
+      'date last modified',
+      'release setting',
+      'release duration',
+      'visibility setting',
+      'license setting',
+      'required license',
+      'default license',
+      'DOI yes/no',
+      'review workflow'
+    ].freeze
+
+    # rubocop:disable Metrics/AbcSize
+    # rubocop:disable Metrics/MethodLength
+    def row(collection)
+      collection_version = collection.head
+      [
+        collection_version.name,
+        collection.id,
+        collection.druid,
+        collection_version.state,
+        collection_version.version,
+        collection.creator.sunetid,
+        collection.managed_by.map(&:sunetid).join('; '),
+        collection.created_at,
+        collection_version.updated_at,
+        collection.release_option,
+        collection.release_duration,
+        collection.access,
+        collection.license_option,
+        collection.required_license,
+        collection.default_license,
+        collection.doi_option,
+        collection.review_enabled
+      ]
+    end
+    # rubocop:enable Metrics/AbcSize
+    # rubocop:enable Metrics/MethodLength
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -53,6 +53,7 @@ Rails.application.routes.draw do
   resource :admin, only: :show
   namespace :admin do
     resources :users, only: :index
+    resources :collection_reports, only: %i[new create]
   end
 
   # This route is used by the emails for the contact the SDR team link.
--- a/app/components/admin/navigation_component.rb
+++ b/app/components/admin/navigation_component.rb
@@ -4,7 +4,9 @@ module Admin
   # Displays the drop down navigation for the admin pages
   class NavigationComponent < ApplicationComponent
     def dropdown
-      options = options_for_select([['Admin page', admin_path], ['Search for user', admin_users_path]],
+      options = options_for_select([['Admin page', admin_path],
+                                    ['Search for user', admin_users_path],
+                                    ['Generate collections report', new_admin_collection_report_path]],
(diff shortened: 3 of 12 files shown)


.rubocop.yml
```diff
@@ -35,6 +35,8 @@ Metrics/ClassLength:
 
 Metrics/MethodLength:
   Max: 15
+  Exclude:
+    - app/services/admin/collections_csv_generator.rb
```
jcoyne, line 39: Please move this exclusion online

app/controllers/admin/collections_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Admin
+  # Generates collection reports
+  class CollectionsController < ApplicationController
```
jcoyne, line 5: Can this be CollectionReportsController?

app/controllers/admin/collections_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      if params[:commit]
+        generate_report
+      else
+        @collections_report = CollectionsReport.new
+        @collection_results = nil
+      end
+    end
+
+    def report
```
jcoyne, line 21: Then this can be `create`

app/controllers/admin/collections_controller.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Admin
+  # Generates collection reports
+  class CollectionsController < ApplicationController
+    before_action :authenticate_user!
+    verify_authorized
+
+    def index
```
jcoyne, line 9: This can be `new`

app/controllers/admin/collections_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      authorize! :admin_dashboard
```
jcoyne, line 10: This should have its own policy

app/services/admin/collections_csv_generator.rb
```diff
@@ -0,0 +1,69 @@
...
+    # rubocop:disable Metrics/AbcSize
+    def row(collection)
+      [
+        collection.head.name,
+        collection.id,
+        collection.druid,
+        collection.head.state,
+        collection.head.version,
+        collection.creator.sunetid,
+        collection.managed_by.map(&:sunetid).join(', '),
```
jcoyne, line 57: Comma separation in a CSV string might be confusing. I might use a semicolon instead.

config/routes.rb
```diff
@@ -53,6 +53,11 @@
   resource :admin, only: :show
   namespace :admin do
     resources :users, only: :index
+    resources :collections, only: %i[index report] do
```
jcoyne, line 56: This would become :collection_reports

jcoyne (changes requested):

app/controllers/admin/collections_controller.rb
justinlittman: UserController uses the admin dashboard policy: https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/admin/users_controller.rb#L12

Since authorization will always and only be for admins, is it worth proliferating policies?

app/controllers/admin/collections_controller.rb
justinlittman: It can, but this matches the naming conventions of `UsersController`.

jcoyne: For UsersController, the resource being controlled was a User, in this case the resource being controlled is the CollectionReport (a report about collections).

app/controllers/admin/collections_controller.rb
jcoyne: I think it's getting very confusing becase we are adding aliases like `report?` that don't make sense in the case of a DashboardPolicy.

app/controllers/admin/collections_controller.rb
justinlittman: Changed.

app/controllers/admin/collections_controller.rb
justinlittman: Done.

config/routes.rb
justinlittman: Done.

app/services/admin/collections_csv_generator.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+      end
+    end
+
+    private
+
+    attr_reader :collections
+
+    # rubocop:disable Metrics/MethodLength
+    def headers
```
jcoyne, line 30: Could this be a constant rather than a method call?

app/services/admin/collections_csv_generator.rb
```diff
@@ -0,0 +1,71 @@
...
+        'visibility setting',
+        'license',
+        'DOI yes/no',
+        'review workflow'
+      ]
+    end
+
+    # rubocop:disable Metrics/AbcSize
+    def row(collection)
+      [
```
jcoyne, line 51: It might be easier to read if you have:
```
collection_version = collection.head
```

here at the beginning, then use that for cols 0, 3, 4, 8

jcoyne (approved):

PR #2439, opened 2022-08-18
Adding druid search to admin

### LINKED ISSUE #2382: Move existing druid search to Admin dashboard
This search functionality should be moved from the dashboard page you see when you log in and added to the Admin dashboard page. They should no longer be on the main dashboard page.

Screen Shot 2022-06-10 at 4 23 54 PM

The druid search should be available from the drop-down menu at the top of the page and should find druids for both items and collections. When the druid is found, user should be taken directly to the item view page or the collection details page (note that this currently goes to the collection settings page but should be changed).

Mockup showing Druid search in drop-down list along with error message if druid is not found: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/467760849

Screen Shot 2022-06-16 at 3 40 11 PM

lwrubel:
Resolves #2382 to add a druid search to the Admin section. Also changes the behavior of the All collections druid search to go to a collection's details (not settings) tab. Both search boxes behave the same way in terms of showing an error on the Search for DRUID page if the druid is not found. 

Unit and QA. 

Screen Shot 2022-08-19 at 7 28 00 AM


app/controllers/admin/druids_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    def index
+      authorize! :admin_dashboard
+      return unless params[:query]
+
+      result = lookup_druid(params[:query])
+      return @failure = true if result.failure?
+
+      item = result.value!
+      redirect_to collection_version_path(item.head) if item.instance_of?(Collection)
```
lwrubel, line 19: Can't use polymorphic_path since we need to redirect to the collection_version_path instead of collection_path (which would go to the settings tab). Is there some better way to do this under the circumstances?

app/controllers/searches_controller.rb
```diff
@@ -6,8 +6,9 @@ def show
     druid = params.require(:q)
     druid = "druid:#{druid}" unless druid.start_with?('druid:')
     result = Work.find_by(druid: druid) || Collection.find_by(druid: druid)
-    return render js: "window.location='#{polymorphic_path(result)}'" if result
+    return render json: { message: 'not found' }, status: :not_found unless result
 
-    render json: { message: 'not found' }, status: :not_found
+    path = result.instance_of?(Work) ? work_path(result) : collection_version_path(result.head)
+    render js: "window.location='#{path}'"
```
lwrubel, line 12: Had to remove polymorphic_path since we're changing which tab to go to for a collection.

app/controllers/admin/druids_controller.rb
jcoyne: I would have done:
```ruby
redirect_to item.is_a?(Collection) ? item.head : item
```

app/controllers/searches_controller.rb
```diff
@@ -6,8 +6,9 @@ def show
     druid = params.require(:q)
     druid = "druid:#{druid}" unless druid.start_with?('druid:')
     result = Work.find_by(druid: druid) || Collection.find_by(druid: druid)
-    return render js: "window.location='#{polymorphic_path(result)}'" if result
+    return render json: { message: 'not found' }, status: :not_found unless result
```
jcoyne, line 9: Can we remove this whole controller?

jcoyne (changes requested):

app/controllers/admin/druids_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+
+    before_action :authenticate_user!
+    verify_authorized
+
+    def index
+      authorize! :admin_dashboard
+      return unless params[:query]
+
+      result = lookup_druid(params[:query])
+      return @failure = true if result.failure?
```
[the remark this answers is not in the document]
jcoyne: I think this is fine.  Sandi Metz has a rule of thumb that a controller method should set only a single instance variable, which is what you've done here.

app/controllers/searches_controller.rb
lwrubel: Do we still want it to do a popup if the druid is not found or go to the same result you get from the "search for druid" page?

lwrubel: I went ahead and made this box behave like the other one for consistency, using the same controller as you suggested.

app/components/admin/all_collections_component.html.erb
```diff
@@ -1,10 +1,9 @@
 <section class="mb-5 allCollections">
-  <%= form_with url: search_path, method: :get, local: false,
-                data: { controller: 'find-druid', type: 'html', action: 'ajax:error->find-druid#onFailure' },
+  <%= form_with url: admin_druid_searches_path, method: :get,
```
jcoyne, line 2: I think the idea was to remove the form from here.

app/policies/admin_dashboard_policy.rb
```diff
@@ -2,6 +2,8 @@
 
 # Defines who is authorized to see the admin dashboard
 class AdminDashboardPolicy < ApplicationPolicy
+  alias_rule :index?, to: :show?
```
jcoyne, line 5: Can you make a new policy for the Druid search?

jcoyne (changes requested):

app/components/admin/all_collections_component.html.erb
lwrubel: Thanks, I hadn't understood that since it was in older mockups. Will remove from that location.

lwrubel: Removed the search box.

config/routes.rb
```diff
@@ -52,6 +52,7 @@
 
   resource :admin, only: :show
   namespace :admin do
+    resources :druid_searches, only: :index
```
jcoyne, line 55: Please remove the old search route too.

jcoyne (changes requested):

config/routes.rb
lwrubel: This is the new route for the new item in the dropdown.

lwrubel: Oh wait, do you mean line 18?

lwrubel: I've removed the searches route.

jcoyne: Looks good.

jcoyne (approved):

PR #2441, opened 2022-08-18
changes around sunet validation and entry on collection edit form

### LINKED ISSUE #2278: Update validations for SUNet ID search on collection form
This is for the initial full collections form and the collection settings form and applies to the fields used to add manager, depositors, and reviewers. 

We still have a lot of users entering strings that aren't one single SUNet ID into these boxes. Sometimes this information manages to get saved, but of course doesn't provide the user access.

In the screenshot below, I was able to type in people, which was not found in the search, but the box still highlighted in green and I was able to save the form. However, the entry itself was not saved and the user was not alerted that it was not saved.

Screen Shot 2022-06-08 at 2 02 28 PM

Each box should have one single SUNet ID in it, so we should add validations that make sure this is the case:

- string must be 3-8 characters in length
- string must start with a letter
- string must contain only letters and numbers

Even with this, we won't prevent the issue with "people" above, so the validation should also require that the user has clicked the "Add" button in the search results in order to be able to deposit (for the initial form) or save (for the collection settings form).

### LINKED ISSUE #2279: Prevent situation where valid SUNet ID is not found in SUNet ID search
This is for the initial full collections form and the collection settings form and applies to the fields used to add manager, depositors, and reviewers.

Sometimes a valid SUNet ID does not return results in the search.
Screen Shot 2022-06-08 at 2 07 38 PM

If the user continues to try, the search does always seem to eventually find the user so they can be added.

While I suspect this is the result of a connection issue with the LDAP server that we can't do anything about, it would be good to prevent this from happening or, if we can't do that, to provide some feedback to the user that there was an issue and they should try it again.

peetucket:

Fixes #2278 and fixes #2279

Changes:

- Updated instruction text above entry form
- Updated message shown when no results are found
- Clear sunet entry text box when focus is lost (unless there is a found search result)
- Hide search result modal when focus is lost (unless there is a found search result)
- Prevent return from being pressed in text box (stops form from being prematurely saved if the user has not clicked the "Add" button)
- Stop user from entering text that does not look like a sunet (must be between 3-8 characters, no special characters, doesn't start with a number)
- Double check that values are not saved unless you click the "Add" button first

New error message when results not found:

Screen Shot 2022-08-18 at 10 31 00 AM

Localhost

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -15,7 +15,7 @@ class AccountsController < ApplicationController
   # Does a lookup from the account service in production mode, otherwise returns a stub value
   def lookup
     return AccountService.new.fetch(params[:id]) if Rails.env.production?
-    return {} if params[:id].size < 6
+    return {} unless params[:id].start_with?('jcoyne85')
 
     {
       'name' => 'Coyne, Justin Michael',
--- a/app/components/collections/settings_component.html.erb
+++ b/app/components/collections/settings_component.html.erb
@@ -72,7 +72,7 @@
   <section id="participants">
     <header>Collection participants *</header>
     <div class="mb-3 row">
-      <p class="col-sm-12">Enter SUNet IDs of participants. Separate each one with a comma and a space. (e.g. janelath, lelandst)</p>
+      <p class="col-sm-12">Select the appropriate option below to add a manager or depositor. Enter one participant SUNet ID per box and then click "Add" to save the entry.</p>
     </div>
 
     <%= render Collections::Update::ManagersComponent.new(form: form) %>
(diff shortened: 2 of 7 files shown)


app/controllers/accounts_controller.rb
```diff
@@ -15,7 +15,7 @@ def show
   # Does a lookup from the account service in production mode, otherwise returns a stub value
   def lookup
     return AccountService.new.fetch(params[:id]) if Rails.env.production?
-    return {} if params[:id].size < 6
+    return {} unless params[:id].start_with?('jcoyne85')
```
peetucket, line 19: in development mode, we only have a single fixture faked result, so let's only return it if it looks like it matches (otherwise it can be confusing to developers)

jcoyne (approved):

PR #2442, opened 2022-08-18
Ensure that Chrome also defaults years to the max value

### LINKED ISSUE #2432: Year field fix from #2357 should also work in Chrome
Currently when you use Firefox and click the button, it defaults to the current year, in Chrome it defaults to 1000.

See #2357 

Alternative approach:
- get rid of arrows by getting rid of input field type being a number (just make a plan text input field)
- ensure we still have validation to prevent a user from entering anything other than a 4 digit number in the range of 1000 up to the current year

jcoyne:
This also switches from cuprite to selenium-webdriver for JS testing.  This is what we use in Argo.

Fixes #2432

Tested manually on different browsers.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,7 +37,8 @@ end
 group :test do
   gem 'capybara', '~> 3.34'
   gem 'capybara-screenshot'
-  gem 'cuprite'
+  gem 'selenium-webdriver' # for js testing
+  gem 'webdrivers' # installs the chrome for selenium tests
 end
 
 group :deployment do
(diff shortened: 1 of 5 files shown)


ndushay (approved):
this looks good but the tests are failing.

jcoyne:
@ndushay I know, but I'm a bit at a loss for what to do.  Cuprite can't support the events we need.

peetucket:
Trying to decide what to do with this, had to rebase and force push to fix conflicts.  I'm also not clear on how to deal with failing tests.

peetucket:
Moving fix for #2432 only to #3109 ... leaving out changes from cuprite to selenium-webdriver  ... can be done later in separate PR as needed (tests currently failing in this PR)

PR #2445, opened 2022-08-18
show client side validation for duplicate keyword entry

### LINKED ISSUE #1502: Notify user and prevent saving when duplicate keyword has been entered

peetucket:

Fixes #1502 - adds client side duplicate keyword validation

Notes:
- for simplicity of implementation, I reused the existing "invalid selection" that typically shows for blank keywords and expanded the error message from "Keyword must be filled in" to "Keyword must be filled in and be unique" so there is a single client side error message that will show in either case
- again for simplicity, the error shows next to the last text box that lost focus when a duplicate was detected, so it may be possible to get into some situations where the error shows on the wrong text box, though i think this simple method should work fine in most cases

Example:

H2-keywords

Localhost

### DIFF
--- a/app/components/works/keywords_row_component.html.erb
+++ b/app/components/works/keywords_row_component.html.erb
@@ -3,7 +3,7 @@
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
     <div data-controller="keywords autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
       <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input',
-                'data-action': 'blur->keywords#checkForDuplicates' } %>
+                'data-keywords-target': 'input', 'data-action': 'blur->keywords#checkForDuplicates' } %>
       <!-- This input is populated by autocomplete controller.-->
       <!-- That triggers the autocomplete-edit controller, which populates the uri and cocina type inputs. -->
       <input type="hidden" data-autocomplete-target="hidden" data-autocomplete-edit-target="value" data-action="autocomplete-edit#change" />
--- a/app/components/works/keywords_row_component.html.erb
+++ b/app/components/works/keywords_row_component.html.erb
@@ -1,15 +1,16 @@
-<div class="mb-3 row plain-container g-0">
+<div class="mb-3 row plain-container g-0 keyword-row">
   <div class="col-sm-6">
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
-    <div data-controller="autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
-      <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
+    <div data-controller="keywords autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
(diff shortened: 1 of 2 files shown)


app/components/works/keywords_row_component.html.erb
```diff
@@ -1,15 +1,16 @@
-<div class="mb-3 row plain-container g-0">
+<div class="mb-3 row plain-container g-0 keyword-row">
   <div class="col-sm-6">
     <%= form.label :label, 'Keyword', class: 'col-form-label sr-only' %>
-    <div data-controller="autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
-      <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input' } %>
+    <div data-controller="keywords autocomplete autocomplete-edit" data-autocomplete-url-value="/autocomplete" data-autocomplete-min-length-value="3" class="dropdown">
+      <%= form.text_field :label, { class: "form-control#{ 'is-invalid' if error?}", required: true, 'data-autocomplete-target': 'input', 'data-autocomplete-edit-target': 'input',
+                'data-action': 'blur->keywords#checkForDuplicates' } %>
       <!-- This input is populated by autocomplete controller.-->
```
peetucket, line 7: just added a new action to the keyword text box to check for duplicates on losing focus

app/javascript/controllers/keywords_controller.js
```diff
@@ -23,16 +23,23 @@ export default class extends Controller {
     this.containerTarget.classList.remove('is-invalid')
   }
 
-  checkForEnter(event) {
-    if (event.key === 'Enter') {
-      event.preventDefault() // prevent form submission
-      this.add(event);
```
peetucket, line 29: i spotted this similar method to what we just implemented for sunetids, but could not find any references to this particular "checkForEnter()" method anywhere in the code, so I removed it

app/javascript/controllers/keywords_controller.js
```diff
@@ -23,16 +23,23 @@ export default class extends Controller {
...
+  checkForDuplicates(event) {
+    const input = event.target.closest('input[type="text"]')
+    const keywords = Array.from(document.querySelectorAll('.keyword-row:not([style*="display: none"]) input[type="text"]'), input => input.value)
```
peetucket, line 28: this is the gnarly selector -- basically look for any non-hidden `.keyword-row` elements that have a text box, and map out their values into an array

app/javascript/controllers/keywords_controller.js
```diff
@@ -23,16 +23,23 @@ export default class extends Controller {
     this.containerTarget.classList.remove('is-invalid')
   }
 
-  checkForEnter(event) {
-    if (event.key === 'Enter') {
-      event.preventDefault() // prevent form submission
-      this.add(event);
+  checkForDuplicates(event) {
+    const input = event.target.closest('input[type="text"]')
```
jcoyne, line 27: Could you use a stimulus target for this?

app/javascript/controllers/keywords_controller.js
jcoyne: How about extracting this into a hasDuplicateKewords method?  This looks pretty similar to what we have in https://github.com/sul-dlss/argo/blob/6139bc519f915c25cd8e55eccd36a6708b22ef49/app/javascript/controllers/registration_item_row_controller.js#L94-L100

jcoyne (changes requested):

app/javascript/controllers/keywords_controller.js
peetucket: Yup - done

jcoyne (approved):

Issue #2312, opened 2022-06-13
Change owner: Add to Item view pages a function that enables a  change in item ownership


amyehodge:
**Background**
We need to be able support use cases where the depositor of an item no longer needs access to the item and wants another user to be able to manage it. 

This should be a "current owner" field (i.e. a new field in the database) so that we can retain the information about who the original depositor was. The "current owner" should have all the same permissions that the original depositor had with respect to that item.

The original depositor (or other previous owners) should no longer see the item on their dashboard or be able to perform any actions with the item, and they should no longer receive any email notifications about the item.

We should be able to accommodate more than one change of ownership, i.e. item is deposited by Person A, then Person B becomes the current owner, and sometime later Person C becomes the current owner.

amyehodge:
We also want to display an error message to the user if the work ID is not found:

Work ID not found. Please try searching for another work ID number.

![Screen Shot 2022-06-15 at 12 22 54 PM](https://user-images.githubusercontent.com/5558511/173877477-193e6000-d4f3-4d0d-adea-c1d03857db62.png)

amyehodge:
This feature will go on the item view page along with two other features. This is the what that page looks like before this feature is selected:
![image (3).png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/82b4fd90-fe2d-483a-8fd6-e59cc7ee4033)

And this is what the change owner function looks like. The admin will perform the same kind of SUNet ID search that we are using on the collection form to make sure they are entering a valid SUNet and that it is for the correct person. When they identify the correct person, they will click the Add button. And then the "change owner" button.
![image (2).png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/635cef2c-db55-43a2-aead-70d1b3184806)

justinlittman:
All additional work related to this ticket has been split into additional tickets.

PR #2454, opened 2022-08-22
Support changing work owners.

justinlittman:
refs #2312

Support admin function.

Unit

### DIFF
--- a/app/assets/stylesheets/work.scss
+++ b/app/assets/stylesheets/work.scss
@@ -57,4 +57,12 @@
   .edit {
     margin-left: 10px;
   }
+
+  .admin-section {
+    background-color: $silver-sand;
+
+    label.admin-functions {
+      @extend h5;
+    }
+  }
 }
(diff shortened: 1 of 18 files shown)


app/controllers/application_controller.rb
```diff
@@ -15,7 +15,10 @@ class ApplicationController < ActionController::Base
   authorize :user_with_groups
 
   def user_with_groups
-    UserWithGroups.new(user: current_user, groups: ldap_groups) if current_user
+    UserWithGroups.new(user: current_user, groups: ldap_groups)
+  rescue Devise::MissingWarden
+    # A nil UserWithGroups, indicating that not logged in
```
jcoyne, line 20: This should never happen.   If we are seeing this. Then components will not be showing up on the page (after a live update) when they should

jcoyne (changes requested):

app/controllers/application_controller.rb
jcoyne: Can this change be reverted?

jcoyne (changes requested):

app/controllers/application_controller.rb
justinlittman: What? I thought I did. Looking.

justinlittman: Apparently it only works if you hit save in VSCode ...

jcoyne (approved):

PR #2467, opened 2022-08-23
Make duplicate file prevent saving or depositing

### LINKED ISSUE #2336: When there is a "duplicate file name" error, do not allow saving or depositing
If the user has uploaded a file with the same name as another file in the deposit, they see the error message below.

Screen Shot 2022-06-16 at 1 57 15 PM

They should then not be able to save or deposit the item in this state, but they currently are able to both these things.

lwrubel:
Fixes #2336. If a file marked as a duplicate exists, the work cannot be saved or deposited.

Unit and Amy reviewed on stage.

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -6,7 +6,7 @@
                               bs_toggle: 'modal',
                               bs_target: '#contactUsModal' }) %>
   </p>
-  <div class="dropzone dropzone-default<%= ' is-invalid' if error? %>" data-dropzone-target="container">
+  <div class="dropzone dropzone-default" data-dropzone-target="container">
     <%= form.file_field :files, multiple: true, direct_upload: true,
                         data: { dropzone_target: 'input' } %>
     <div class="dz-message needsclick text-secondary">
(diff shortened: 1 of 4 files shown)


app/components/works/file_row_component.html.erb
```diff
@@ -43,7 +43,7 @@
 
   <div class="col-md-1">
     <%= button_tag type: 'button',
-                   class: "dz-remove pull-right btn",
+                   class: "dz-remove pull-right btn remove-file",
```
lwrubel, line 46: Added this class to provide a handle for testing.

app/components/works/add_files_component.html.erb
```diff
@@ -6,7 +6,7 @@
                               bs_toggle: 'modal',
                               bs_target: '#contactUsModal' }) %>
   </p>
-  <div class="dropzone dropzone-default<%= ' is-invalid' if error? %>" data-dropzone-target="container">
+  <div class="dropzone dropzone-default" data-dropzone-target="container">
```
lwrubel, line 9: It seemed that this div was not included in what `validate-forms.js` checks when looking for 'is-invalid' on form elements.

justinlittman (approved):

PR #2477, opened 2022-08-24
Changes collection report sort order to collection name.

### LINKED ISSUE #2460: Collection report should by default be sorted by collection title

justinlittman:
closes #2460

Per ticket

Unit

### DIFF
--- a/app/services/admin/collections_report_query.rb
+++ b/app/services/admin/collections_report_query.rb
@@ -20,7 +20,7 @@ module Admin
       add_date_created_end_filter
       add_date_modified_start_filter
       add_date_modified_end_filter
-      query.order('users.email ASC')
+      query.order('collection_versions.name ASC')
     end
 
     private
@@ -31,7 +31,7 @@ module Admin
     def add_status_filter
       return if collections_report.statuses.empty?
 
-      self.query = query.where(head: { state: collections_report.statuses })
+      self.query = query.where(collection_versions: { state: collections_report.statuses })
     end
 
     def add_date_created_start_filter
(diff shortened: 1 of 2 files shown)


app/services/admin/collections_report_query.rb
```diff
@@ -31,7 +31,7 @@ def generate
     def add_status_filter
       return if collections_report.statuses.empty?
 
-      self.query = query.where(head: { state: collections_report.statuses })
+      self.query = query.where(collection_versions: { state: collections_report.statuses })
```
jcoyne, line 34: will this pull in all versions rather than just the head version?  Is this following the right relationship?

jcoyne (changes requested):

app/services/admin/collections_report_query.rb
justinlittman: You're right.
```
SELECT "collections".* FROM "collections" INNER JOIN "collection_versions" ON "collection_versions"."id" = "collections"."head_id" INNER JOIN "users" ON "users"."id" = "collections"."creator_id" WHERE "collection_versions"."state" = $1 ORDER BY collection_versions.name ASC
```

justinlittman: Oh, no wait: That is correct: `INNER JOIN "collection_versions" ON "collection_versions"."id" = "collections"."head_id"`

jcoyne: Thank you for reassuring me.

jcoyne (approved):

PR #2482, opened 2022-08-24
Support decommissioning of works.

### LINKED ISSUE #2375: Decommission: Create method to decommission H2 items
Currently, decommissioning items that were deposited via H2 requires:

1. The "normal" decommission process, carried out via Argo
2. Extra steps to delete from H2, which requires updating the H2 database directly

For the upcoming workcycle, we'd like to make the H2 steps a function available in the web UI to H2 administrators. It is not a requirement to connect the H2 steps with the Argo steps (which are standard for any SDR content regardless of source).

**Proposal**

Implement decomissioning as a "soft delete":

- Lock item to prevent future edits
- Change ownership to an H2 admin so original depositor no longer has it on their dashboard
- Remove from collection (open question: does this mean item needs to move to a decommission/"graveyard" collection)?
- Do a regular delete of the files

Advantages of the soft delete approach:

- Retains a record of decommissioned items
- Potential to run reports on decommissioned items as an H2 admin

justinlittman:
closes #2375

Errors happen.

Unit, Amy.

### DIFF
--- a/app/components/collections/works_component.rb
+++ b/app/components/collections/works_component.rb
@@ -13,7 +13,7 @@ module Collections
 
     def works
       policy.authorized_scope(collection.works, as: :edits, with: WorkVersionPolicy)
-            .order('updated_at DESC')
+            .order('works.updated_at DESC')
     end
 
     def hide_depositor?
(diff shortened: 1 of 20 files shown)


jcoyne (approved):

jcoyne:
@justinlittman this requires a rebase.

justinlittman:
@amyehodge approved.

PR #2487, opened 2022-08-24
add ability for admins to lock/unlock item

### LINKED ISSUE #2317: Lock editing: Add function to Item view pages to lock an item
This will allow admins (only) to lock an item that is in any state, so that it can't be edited by anyone (#2381), including admins, within the H2 interface.

Admins should be able to unlock an item that was previously locked by this feature and by any admin.

This feature will go on the item view page along with two other features. This is the what that page looks like before this feature is selected:

And this is what the lock/unlock function looks like, though the selection in the dropdown should say **Lock/Unlock**. 

When the item is already locked, the user will instead see a button that says "Unlock" and the message should say "Unlocking this item will allow the depositor to make changes to this item. You may relock this item in the future."

peetucket:

Fixes #2317 - add lock/unlock function for admins to the existing admin item menu.  Toggles a new field at the work model level to set if an item is locked or not.

Open to suggestions on ways to make this more idiomatic/cleaner.  I structured the controller under the assumption that we may want more "admin item" tasks to add to the same controller later, but maybe that won't happen and/or we will just create new controllers for each new admin item function.

Screen Shot 2022-08-24 at 4 40 02 PM

Localhost and new feature test.


app/components/works/admin_component.html.erb
```diff
@@ -7,8 +7,11 @@
...
       </select>
     </div>
   </div>
-  <turbo-frame id="workAdminSection"></turbo-frame>
-</section>
\ No newline at end of file
+  <div class="row pt-3">
+    <turbo-frame id="workAdminSection"></turbo-frame>
+  </div>
+</section>
```
peetucket, line 17: just adding some padding to this area

app/components/works/admin_component.html.erb
```diff
@@ -7,8 +7,11 @@
...
         <option value='select' selected>Select...</option>
         <option value='<%= edit_owners_path(work) %>'>Change owner</option>
+        <option value='<%= edit_lock_work_path(work) %>'>Lock</option>
       </select>
     </div>
   </div>
-  <turbo-frame id="workAdminSection"></turbo-frame>
-</section>
\ No newline at end of file
+  <div class="row pt-3">
```
jcoyne, line 14: I don't think you want to put this here because for a non-admin, the padding still displays, but the frame may not have anything in it.

app/controllers/work_admins_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# a controller for the admins for a work.
+class WorkAdminsController < ApplicationController
```
jcoyne, line 4: Can you follow the pattern established in #2454 and make this a LocksController or similar?

app/controllers/work_admins_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  before_action :authenticate_user!
+  verify_authorized
+
+  def edit_lock
```
jcoyne, line 8: Then this can be just `edit`

app/controllers/work_admins_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    authorize! :work_admin
+    @work = Work.find(params[:id])
+  end
+
+  def update_lock
```
jcoyne, line 13: and this can be just `update`

app/policies/work_admin_policy.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Authorization policy for Work admins
+class WorkAdminPolicy < ApplicationPolicy
```
jcoyne, line 4: Make this the `LockPolicy`.

jcoyne (changes requested):

app/components/works/admin_component.html.erb
peetucket: This entire component should not render unless you are admin, and this padding is contained in the body of that component where the action specific content appears: https://github.com/sul-dlss/happy-heron/blob/main/app/components/works/admin_component.rb#L12-L14

app/controllers/work_admins_controller.rb
peetucket: Ok, did it this way under the assumption that we may want to use this controller for more than just locking (assuming there are other work specific admin functions), but we could also just create more controllers in that case.

peetucket:
Renamed to "WorkLocks" controller, updated action names, renamed policy to match.

app/views/work_locks/edit.html.erb
```diff
@@ -0,0 +1,17 @@
+<turbo-frame id='workAdminSection'>
+  <%= form_with url: locks_path(params[:id]), method: :put, data: {turbo: false} do |form| %>
```
jcoyne, line 2: Why are you disabling turbo here?

app/policies/work_lock_policy.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Authorization policy for Work admins to lock/unlock
+class WorkLockPolicy < ApplicationPolicy
+  def edit?
+    administrator?
```
jcoyne, line 6: Since edit/update are closely coupled, can you make edit an `alias_rule`?  Then there's no duplication.

app/components/works/admin_component.html.erb
jcoyne: I see. This is great then.

jcoyne (changes requested):

app/views/work_locks/edit.html.erb
peetucket: After the item is locked/unlocked by the controller, there is a re-direct and a flash message is shown, so we break out of the turbo frame for that.  Same pattern as the similarly implemented "change owner" feature: 

https://github.com/sul-dlss/happy-heron/blob/main/app/controllers/work_owners_controller.rb#L12-L23
https://github.com/sul-dlss/happy-heron/blob/main/app/views/work_owners/edit.html.erb#L2

app/policies/work_lock_policy.rb
peetucket: Yup.  Also changed the similar `WorkOwnerPolicy` to match.

app/views/work_locks/edit.html.erb
jcoyne: Ah, okay, I must not have the latest code when I looked.

jcoyne (approved):

PR #2494, opened 2022-08-25
Require other for other type.

### LINKED ISSUE #2277: Change error text and where it is displayed when user does not specify "Other" type
In the work type modal, if the user does not enter text in the box for "Specify "Other" type" and they click on Continue, the following should happen:

- text box should be outlined in red
- error message should display below the text near the top that says "*required field". This text should read ""Specify "Other" type is a required field."

Screen Shot 2022-06-08 at 1 48 06 PM

Currently what happens when the user doesn't fill in this box and clicks Continue is that they are taken back to the dashboard and an error message appears below the page header that says "Invalid subtype value for work of type 'other': missing". This text needs to be clearer and ideally within the modal as shown above so the user can complete what they are doing there instead of being ejected back to the dashboard.

### LINKED ISSUE #2777: [happy-heron/stage] NotImplementedError: NotImplementedError
## Backtrace

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/88770493)

justinlittman:
closes #2277

Local, Amy

image

### DIFF
--- a/app/components/works/work_type_modal_component.html.erb
+++ b/app/components/works/work_type_modal_component.html.erb
@@ -52,7 +52,7 @@
 
           <template data-work-type-target="otherTemplate">
             <div class="subtype-item">
-              <input type="text" name="subtype[]" class="form-control" id="subtype_other">
+              <input type="text" name="subtype[]" class="form-control" id="subtype_other" required>
             </div>
           </template>
 


justinlittman:
@amyehodge I took the easy approach using native browser validation functionality. If this isn't adequate, let me know.

justinlittman:
@amyehodge approved.

jcoyne (approved):

PR #2497, opened 2022-08-25
Change layout of other subtype input.

### LINKED ISSUE #2276: In work type modal, update the free text field that displays when "Other" is selected
When a user selects the "Other" work type option, the free text field that displays should:

1. have an asterisk to the right of the "Specify "Other" type" text, and
2. be centered under the work type options and the same width as those three columns are.

Screen Shot 2022-06-08 at 1 02 25 PM

justinlittman:
closes #2276

Local

image

### DIFF
--- a/app/components/works/work_type_modal_component.html.erb
+++ b/app/components/works/work_type_modal_component.html.erb
@@ -46,12 +46,11 @@
             <h6>Select at least two terms below:</h6>
           </template>
 
-          <template data-work-type-target="otherTemplateHeader">
-            <h5>Specify "Other" type</h5>
-          </template>
-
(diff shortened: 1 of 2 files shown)


justinlittman:
@amyehodge This is the best that I can do with my CSS skills. Is it close enough?

justinlittman:
@jcoyne I can't get this test to pass. It works fine in the local browser, but isn't displaying in the test.

image

justinlittman:
@amyehodge approved.

jcoyne:
@justinlittman needs a rebase.

jcoyne (approved):

PR #2500, opened 2022-08-25
Use approver for depositing event.

### LINKED ISSUE #2094: Reviewer's SUNet ID should appear in item history for approval/deposit
For an example of this issue, see https://sdr.stanford.edu/works/4672, history section.

The review workflow is on for this collection. User lauraand submitted this work for approval on August 26. The work was reviewed and approved by thalassa on Sep 13, 2021, but it shows "lauraand" for these steps as well. This is confusing and makes it appear as if lauraand approved her own work.

When a reviewer is the one that sends the work for deposit after approval, the reviewer's SUNet should appear in the history.

justinlittman:
closes #2094

Less confusing

Unit

### DIFF
--- a/app/controllers/reviews_controller.rb
+++ b/app/controllers/reviews_controller.rb
@@ -8,13 +8,18 @@ class ReviewsController < ApplicationController
   def create
     version = Work.find(params[:work_id]).head
     authorize! version, to: :review?
+    record_review(version)
+
+    redirect_to dashboard_path
+  end
+
+  def record_review(version)
     if params[:state] == 'approve'
+      version.work.event_context = { user: current_user }
       version.begin_deposit!
     else
       version.work.event_context = { description: params[:reason], user: current_user }
       version.reject!
     end
-
-    redirect_to dashboard_path
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/reviews_controller.rb
```diff
@@ -8,13 +8,18 @@ class ReviewsController < ApplicationController
   def create
     version = Work.find(params[:work_id]).head
     authorize! version, to: :review?
+    create_event(version)
+
+    redirect_to dashboard_path
+  end
+
+  def create_event(version)
```
jcoyne, line 16: I'm not in love with this name because creating an "Event" is one of two things that happens here (and it's the less important thing).  The other is that it causes a transition.  How about `record_transition` instead?

jcoyne (changes requested):

app/controllers/reviews_controller.rb
justinlittman: Oh, you're right. That is a terrible method name. Fixing.

jcoyne (approved):

PR #2501, opened 2022-08-25
In work type modal, update options for Mixed Materials.

### LINKED ISSUE #2275: In work type modal, update options for "Mixed Materials"
When a user selects "Mixed Materials" in the work type modal, update the text that displays in red

from: Select at least two terms below:
to: Select at least two terms from any of the options below:

Also, the "See more options" section should be expanded by default for the "Mixed Materials" work type (but not for any other work type).

Remove the javascript that collapses the options. There should no longer be the "see more/fewer options"

Screen Shot 2022-06-08 at 12 59 46 PM

justinlittman:
closes #2275

Local, Amy

image

### DIFF
--- a/app/components/works/work_type_modal_component.html.erb
+++ b/app/components/works/work_type_modal_component.html.erb
@@ -65,8 +65,8 @@
 
           <div class="mb-4 subtype-container" data-work-type-target="moreTypes" hidden></div>
 
-          <button type="submit"
-            class="btn btn-primary mx-auto d-block"
+          <button type="submit" 
+            class="btn btn-primary mx-auto d-block" 
             data-action="work-type#checkSubtypes"
             data-work-type-target="continueButton">Continue</button>
         </form>
--- a/app/components/works/work_type_modal_component.html.erb
+++ b/app/components/works/work_type_modal_component.html.erb
@@ -65,8 +65,8 @@
 
           <div class="mb-4 subtype-container" data-work-type-target="moreTypes" hidden></div>
 
-          <button type="submit" 
-            class="btn btn-primary mx-auto d-block" 
+          <button type="submit"
+            class="btn btn-primary mx-auto d-block"
             data-action="work-type#checkSubtypes"
             data-work-type-target="continueButton">Continue</button>
         </form>
--- a/app/models/work_type.rb
+++ b/app/models/work_type.rb
@@ -61,8 +61,6 @@ class WorkType
     'White paper', 'Working paper'
   ].freeze
 
-  MIXED_TYPES = (%w[Music Software/Code Sound Video] + MORE_TYPES).sort.freeze
-
   attr_reader :id
 
   attr_reader :label
@@ -113,7 +111,7 @@ class WorkType
       new(id: 'music', label: 'Music', html_label: 'Music', icon: 'music',
           subtypes: MUSIC_TYPES, cocina_type: Cocina::Models::ObjectType.object),
       new(id: 'mixed material', label: 'Mixed Materials', html_label: 'Mixed Materials', icon: 'play',
-          subtypes: MIXED_TYPES, cocina_type: Cocina::Models::ObjectType.object),
+          subtypes: MORE_TYPES, cocina_type: Cocina::Models::ObjectType.object),
       new(id: 'other', label: 'Other', html_label: 'Other', icon: 'archive',
           subtypes: [], cocina_type: Cocina::Models::ObjectType.object)
     ]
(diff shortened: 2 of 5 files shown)


justinlittman:
@amyehodge approved.

app/models/work_type.rb
```diff
@@ -63,6 +61,8 @@ class InvalidType < StandardError; end
     'White paper', 'Working paper'
   ].freeze
 
+  MIXED_TYPES = (%w[Data Image Portfolio Music Software/Code Sound Text Video] + MORE_TYPES).uniq.sort.freeze
```
jcoyne, line 64: Can we remove "Data", "Text" and "Video" from this list since they are already in `MORE_TYPES`?

jcoyne (changes requested):

app/models/work_type.rb
jcoyne: The screenshot doesn't display "Sound", should it?

justinlittman: Removing dupes and `uniq`. Screenshot is from earlier code.

jcoyne (approved):

jcoyne (changes requested):
Waiting to hear back from @arcadiafalcone about similar types:
Software/code ~ Software
Sound ~ Sound recording
Video ~ Video recording

app/components/works/work_type_modal_component.html.erb
```diff
@@ -65,8 +65,8 @@
 
           <div class="mb-4 subtype-container" data-work-type-target="moreTypes" hidden></div>
 
-          <button type="submit" 
-            class="btn btn-primary mx-auto d-block" 
+          <button type="submit"
+            class="btn btn-primary mx-auto d-block"
```
mjgiarlo, line 69: Is this a whitespace-only change?

peetucket: looks like it, possibly came from the rebase I just did?

looks like the a cypress test is failing, and may need some attention too (assuming this PR is still relevant after 9 months)

amyehodge:
I'm not sure what the whitespace issue or the similar terms are about, but I definitely do want the change to the instructions and the display to change to show all these right away without needing the user to expand the "More options" text. I'm not clear what is in the PR.

arcadiafalcone:
Looks like this was hashed out on Slack but never made its way back here -- when the user chooses Mixed Materials as the type, do not include the following options in the subtypes:
“Sound” (keeping "Sound recording")
“Video” (keeping "Video recording")
“Software/Code” (keeping “Software” and “Code)

From this Aug. 2022 thread:
https://stanfordlib.slack.com/archives/C69UZCJ8M/p1661786289068899?thread_ts=1661534612.213819&cid=C69UZCJ8M

cypress/spec/work_type_modal.cy.js
```diff
@@ -62,7 +62,7 @@ describe('Work type modal', () => {
       cy.get('div.subtype-container[data-work-type-target="subtype"] div.subtype-item input').last().click()
       cy.get('button[type="submit"]').click()
     })
-    cy.url().should('include', `/collections/${collection_id}/works/new?work_type=mixed+material&subtype%5B%5D=Data&subtype%5B%5D=Video`)
+    cy.url().should('include', `/collections/${collection_id}/works/new?work_type=mixed+material&subtype%5B%5D=3D+model&subtype%5B%5D=Working+paper`)
   })
```
peetucket, line 66: the first and last subtypes selected are slightly different now, and the test needs to reflect this

spec/validators/work_subtype_validator_spec.rb
```diff
@@ -54,7 +54,7 @@
     end
 
     context 'with two valid subtypes and one or more valid "more" types' do
-      let(:value) { %w[Data Sound Preprint] }
+      let(:value) { %w[Data Speech Preprint] }
```
peetucket, line 58: `Sound` is no longer a valid subtype for mixed media, just replace it with a different one that is valid

peetucket:
I made some changes to the PR to include the results of the discussion above, adjusted a test, and rebased from main.  This is now the current list of subtypes shown when you select `Mixed Materials` type.  If correct, this PR is ready for a review and merge:

Screen Shot 2023-06-13 at 11 35 45 AM

amyehodge:
@peetucket  "Music" appears to be missing. Otherwise, looks good.

peetucket:
> @peetucket "Music" appears to be missing. Otherwise, looks good.

Good catch, thanks.  I added it back in.

mjgiarlo:
@peetucket what's the status of this PR? Does it just need review + a conflict resolution?

mjgiarlo (approved):

PR #2506, opened 2022-08-25
Show modal for draft collections and modal for draft works

### LINKED ISSUE #2291: Add collection draft alert when user logs in if they have a collection in draft state
This should be similar to the alert that depositors get when they have a deposit in progress.

At login, **any** manager of a collection that is in a draft state (initial or version draft) should see a modal that looks like the one for items in progress but says:

Continue your deposit
Would you like to continue working on your draft collection
"Collection title"
Yes!/No 

Yes button takes users to the Collection details form
No button takes users to the Dashboard.

image

If a user has both multiple collections and/or items in draft, they should see the most recently touched draft collection. They should only be shown one of these modals, which is the same way we have this implemented when there are just items.

peetucket:

Fixes #2291 - add a new collection draft modal (offer to let you continue) that takes priority over the existing draft works in progress modal.  Only one modal is every shown.

Note this also re-orders collections in progress on the dashboard in most recent descending order (verified as OK with PO), as well as adds the type of deposit being continued in the modal (e.g. "collection" or "work").

Also did some related refactoring to make the display of titles easier and more consistent across the app, and used this update title presenter in more places to ensure we show a page title (currently missing in some spots in the production app when a collection draft in progress has no title).

~~Need to verify one change with PO.~~

Updated component test, and localhost browser.


app/components/breadcrumb_nav_component.rb
```diff
@@ -8,7 +8,7 @@ def initialize(breadcrumbs: [], admin: false)
   end
 
   def full_title(breadcrumb)
-    breadcrumb.presence || 'No title'
+    breadcrumb.presence || Settings.h2.no_title
   end
```
peetucket, line 12: Put all of these "No title" strings in a setting to DRY them up.

app/components/collections/show/settings_header_component.rb
```diff
@@ -16,7 +16,7 @@ def initialize(collection_version:)
       delegate :depositing?, :draft?, :collection, to: :collection_version
 
       def name
-        collection_version.name.presence || 'No Title'
+        DepositTitlePresenter.show(collection_version)
       end
```
peetucket, line 20: refactored the existing `WorkTitlePresenter` component so we can pass it a collection too (see below) and then used that throughout

app/components/dashboard/continue_deposit_modal_component.rb
```diff
@@ -0,0 +1,33 @@
...
+    # the last draft work (if any)
+    def work_version
+      @presenter.in_progress.first
+    end
+
+    def title
+      @title ||= DepositTitlePresenter.show(collection_version || work_version)
+    end
+  end
+end
```
peetucket, line 33: Renamed the `ContinueWorkModalComponent` to `ContinueDepositModalComponent` since we can now show a work or a collection here

app/controllers/dashboards_controller.rb
```diff
@@ -29,6 +29,7 @@ def build_presenter
       collection_managers_in_progress: CollectionVersion.with_state(:first_draft, :version_draft)
                                          .joins(:collection).left_outer_joins(collection: :managed_by)
                                          .where('managers.user_id' => current_user)
+                                         .order('collection_versions.updated_at desc')
     )
```
peetucket, line 33: Most recent at the top

app/presenters/deposit_title_presenter.rb
```diff
@@ -0,0 +1,12 @@
...
+# A helper for displaying a title for a Work or Collection
+class DepositTitlePresenter
+  def self.show(deposit)
+    if deposit.instance_of?(WorkVersion)
+      deposit&.title.presence || Settings.h2.no_title
+    elsif deposit.instance_of?(CollectionVersion)
+      deposit&.name.presence || Settings.h2.no_title
+    end
+  end
+end
```
peetucket, line 12: refactored and renamed this from `WorkTitlePresenter` so it can be used for works or collections

spec/components/dashboard/continue_deposit_modal_component_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    let(:show_popup) { false }
+    let(:presenter) do
+      instance_double(DashboardPresenter, show_popup?: show_popup, in_progress: [], collection_managers_in_progress: [])
+    end
+
+    it 'does not render the component' do
+      expect(rendered.to_html).to eq ''
+    end
+  end
+end
```
peetucket, line 50: spec renamed and new tests added to verify it works with collections and works

app/presenters/deposit_title_presenter.rb
jcoyne: I'm not sure there is commonalty enough to introduce  this abstraction. This adds an additional conditional every time it's called, even though we typically already know what type it is.

jcoyne (changes requested):

app/components/breadcrumb_nav_component.rb
peetucket: Done.

app/presenters/deposit_title_presenter.rb
peetucket: Revert the `WorkTitlePresenter` back to it's current form, added a new `CollectionTitlePresenter` to use in other parts of the code too, and then added this conditional to the `ContinueDepositModalComponent` to pick the right one.

app/presenters/deposit_title_presenter.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# A helper for displaying a title for a Work or Collection
+class DepositTitlePresenter
+  def self.show(deposit)
+    if deposit.instance_of?(WorkVersion)
+      deposit&.title.presence || Settings.h2.no_title
+    elsif deposit.instance_of?(CollectionVersion)
+      deposit&.name.presence || Settings.h2.no_title
```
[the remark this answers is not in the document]
peetucket: it was being done in the `WorkTitlePresenter` before -- not sure why though.  I removed, and seems to work in localhost but not sure if I've hit all the edge cases of new works (though again, I don't **think** it will be nil).

spec/components/dashboard/continue_deposit_modal_component_spec.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Dashboard::ContinueDepositModalComponent, type: :component do
+  let(:rendered) { render_inline(described_class.new(presenter: presenter)) }
+  let(:work_version) { build_stubbed(:work_version) }
+  let(:collection_version) { build_stubbed(:collection_version) }
```
jcoyne, line 8: Can you move these into the contexts where they are used?

jcoyne (changes requested):
Just one more small change to the tests then it looks good.

spec/components/dashboard/continue_deposit_modal_component_spec.rb
peetucket: Yes, I actually added another test as well to ensure we capture the three scenarios (only a work in progress, only a collection in progress, both collection and work in progress).

jcoyne (approved):

PR #2514, opened 2022-08-26
Don't display the lock on each section of the view page

### LINKED ISSUE #2320: Lock editing: On item view page, replace pencil icon with lock
On the item view page, if the item is locked, replace the pencil icon at the top of the page next to the title with a lock icon. This should be true for all users.

When the user clicks on the lock, they should see a tooltip that says "This deposit can no longer be edited in this application. If you need to make changes to this deposit, please use the Help link at the top of this page to contact us."

All other pencil icons on this page should be removed when the item is locked.

Mockup: https://projects.invisionapp.com/share/EQXC9CLKCR2#/screens/467473850

Screen Shot 2022-06-28 at 3 11 25 PM

jcoyne:

Fixes #2320

locally & new request spec

### DIFF
--- a/app/views/works/_edit_button.html.erb
+++ b/app/views/works/_edit_button.html.erb
@@ -17,7 +17,7 @@
         <span class="fa-solid fa-pencil-alt" aria-hidden="true"></span>
       <% end %>
     <% end %>
-  <% elsif work.locked? %>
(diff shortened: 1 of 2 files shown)


ndushay (approved):

ndushay:
before and after screenshots would be nice.  approved, tho.

PR #2515, opened 2022-08-26
Fix state filter in item admin report.

### LINKED ISSUE #2513: Item reports are sometimes empty
Item reports generated in production are sometimes empty, but this is not because there are no matches. 

I consistently get reports without any rows when I use any of the H2 state filters, even though I know the states chosen, like "Deposited," should produce results. Likewise, "PURL Reserved" should produce a report with 3 rows.

@hannahfrost and @andrewjbtw have also reported issues with the other filters. I have had more trouble reproducing those problems, but do sometimes get blank reports with those filters as well.

I think the full report without any filters works fine every time.

justinlittman:
closes #2513

Users were getting incorrect report results.

QA, Amy

### DIFF
--- a/app/views/admin/work_reports/new.html.erb
+++ b/app/views/admin/work_reports/new.html.erb
@@ -89,6 +89,8 @@
           <%= form.hidden_field :date_created_end %>
           <%= form.hidden_field :date_modified_start %>
           <%= form.hidden_field :date_modified_end %>
+          <%= form.hidden_field :date_deposited_start %>
+          <%= form.hidden_field :date_deposited_end %>
           <% @report.state.each do |state|%>
             <%= form.hidden_field :state, multiple: true, value: state %>
           <% end %>          
--- a/app/views/admin/work_reports/new.html.erb
+++ b/app/views/admin/work_reports/new.html.erb
@@ -71,7 +71,7 @@
                                   deposited
                                   version_draft
                                   decommissioned ]%>
-          <%= form.collection_check_boxes(:state, states_in_order.map { |state| [state.to_s, I18n.t(state, scope: 'work.state')] }, :first, :last) do |b| %>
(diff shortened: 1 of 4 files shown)


app/views/admin/work_reports/new.html.erb
```diff
@@ -71,7 +71,7 @@
                                   deposited
                                   version_draft
                                   decommissioned ]%>
-          <%= form.collection_check_boxes(:state, states_in_order.map { |state| [state.to_s, I18n.t(state, scope: 'work.state')] }, :first, :last) do |b| %>
+          <%= form.collection_check_boxes(:state, states_in_order.map { |state| [state.to_s, I18n.t(state, scope: 'work.state')] }, :first, :last, include_hidden: false) do |b| %>
```
jcoyne, line 74: If you make this change, then https://github.com/sul-dlss/happy-heron/blob/main/app/services/admin/work_report_query.rb#L35 should also be changed. that line was stripping out the empty one.

justinlittman:
@amyehodge approved

jcoyne (approved):

jcoyne (changes requested):
Can you add a test that shows how this fixed the situation?

justinlittman:
@jcoyne I'm a bit flumoxed on how to test this. My first inclination was to add it to the `work_report_spec` feature test. However, while I can test the headers, I can't seem to access the correct response body. (Based on googling, this seems driver-specific.)

I don't think that a request spec would add any value because the fix is in what is being sent from the browser.

Open to other suggestions.

jcoyne:
@justinlittman I can't see how this is different functionally from what was there before. 🤷

justinlittman:
It fixes two problems:
* While hidden empty state was being handled in the query service, it was causing problems in other: https://github.com/sul-dlss/happy-heron/issues/2513#issuecomment-1228675864
* The multiple state fields were being treated as a string, not an array, by param parsing.

justinlittman:
Replaced by #2530

PR #2516, opened 2022-08-26
Upgrade to font-awesome 6

jcoyne:
Keep up to date with dependencies.  `fa-solid` is more understandable than `fas`

locally


ndushay (approved):
I assume you looked at some of these to make sure they looked the same ...

jcoyne:
Yep, there the same.

Issue #2524, opened 2022-08-29
SUNet lookup problems


amyehodge:
SUNet ID searches are sometimes looking for "".
<img width="1436" alt="Screen Shot 2022-08-29 at 10 02 40 AM" src="https://user-images.githubusercontent.com/5558511/187237849-4ba6eee6-009a-4aba-8b63-3e8d2af94952.png">

https://user-images.githubusercontent.com/5558511/187237868-4a4cb467-bc44-4c26-a892-f8710f87438e.mov

jcoyne:
@amyehodge do you have any sense whether #2525 or #2551 have fixed this problem?

amyehodge:
@jcoyne I will need to check in with a user that has this problem consistently to see if it has fixed it. I had to postpone my testing with her because of the problems this week with dependency updates. Are both of these fixes in production? If so, I will schedule testing with her.

jcoyne:
The last deploy to production was on Tuesday (found here https://app.honeybadger.io/projects/77112/deploys?environment=prod&project_id=77112) and it was at https://github.com/sul-dlss/happy-heron/commit/3bb5414ae703d533bfdd35c99d9b942c43822549  So the former fix is deployed but not the latter.

amyehodge:
@jcoyne One of our users sent me this screen capture of her problem with SUNet ID additions, with the developer tools on. It's definitely logging something more than mine does, but with it on the right, it's hard to read. Let me know if you'd like more on this and I can talk with her further.

https://user-images.githubusercontent.com/5558511/188215020-a76a10a9-faf7-4d28-b45f-50e30b2e7491.mov

amyehodge:
@jcoyne User update from today on this issue.

I just tried this again. After logging out and logging back in again.

On Chrome, I get the same "Error looking up sunet:"SyntaxError:Unexpected token'<', ""-try again later"

On Safari, I have a new error message (See screenshot)
![Maryann.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/1f903114-2cec-41b9-abc7-5cb6957cb16a)

PR #2525, opened 2022-08-29
Tweaks sunet lookups and adds error handling.

justinlittman:
refs #2524

This functionality is working erratically.

Amy.


peetucket:
So the theory is that the problem is that the remote API goes down and this will at least alert the user in some way?

jcoyne (approved):

justinlittman:
@peetucket That is "a" theory. However, it is not entirely clear whether the user reports match that theory.

PR #2531, opened 2022-08-29
Allows selecting globus upload when creating deposit.

### LINKED ISSUE #2349: Globus: Allow user to designate that files have been uploaded to Globus
**Please release this to production with #2351**

1. Add an option to the file upload section where instead of uploading files, depositor can select an option that says "I have uploaded my files to a Globus endpoint provided to me by SDR staff." Subtext here should read: "Do not check this box unless you have already discussed this with SDR staff."
2. Depositor will then be prevented from uploading files.
3. A placeholder file will be attached to the deposit (does not need to be displayed to user, and probably shouldn't be to avoid confusion)
4. Depositor clicks "Deposit" button to complete the deposit.
5. Notification is sent to Admins that a Globus-related deposit has been completed. (Notification to be ticketed separately.)

**Background**
We have some users depositing large files who misunderstand what they need to do regarding the files in H2. The current workflow is that they provide the metadata, do not upload files, and leave the item in a draft state. Andrew then attaches a placeholder file, deposits the item via H2, and completes accessioning the files outside of H2.
 
Some users don’t understand they should not upload any of the files and still try to upload the smaller ones, some users get confused that they can’t click deposit (even though we told them not to), and we would prefer if they could do the deposit themselves, as this is more appropriate than having Andrew do it.
 
**The new option**
I’d like to add an option in the file upload section of the item deposit form that says something like “I have uploaded my files to a Globus endpoint provided to me by SDR staff.” If this option is selected the user should NOT be allowed to upload any files to the item. Instead, a placeholder file will be attached to the item. I actually think the user doesn’t need to know about this file because I think it will confuse them, but we could discuss that point. The user would then be able to deposit the item. After which, Andrew can complete the accessioning using the files on Globus and lock the item for future editing (assuming we get that feature built in this work cycle).
 
Benefits of this are that 1) the user can’t upload files when they shouldn’t but 2) they can still deposit the item, eliminating both of those points of confusion. And 3) Andrew does not need to deposit the item or manually attach a placeholder file.

justinlittman:
closes #2349

A little bit of automation is better than no automation.

Unit, Amy.

### DIFF
--- a/app/javascript/controllers/complex_radio_controller.js
+++ b/app/javascript/controllers/complex_radio_controller.js
@@ -11,7 +11,7 @@ export default class extends Controller {
     this.selectionTargets.forEach((target) => {
       const checked = target.querySelector("input[type='radio']").checked
       // If radio is checked, enable its child select elements; if unchecked, disable.
-      target.querySelectorAll('select,input[type="number"],input[type="checkbox"]').forEach((element) => element.disabled = !checked)
+      target.querySelectorAll('select,input[type="number"],input[type="checkbox"],fieldset').forEach((element) => element.disabled = !checked)
     })
   }
 }
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -1,29 +1,47 @@
-<section data-controller="dropzone" data-dropzone-max-file-size="10000" data-dropzone-max-files="1000" id="file">
+<section data-controller="complex-radio" id="file">
   <header>Add your files *</header>
-  <p>All file types are accepted. If you have a deposit that is over 10GB, or
-    have any trouble with adding your files,
-    <%= link_to('contact us.', '#contactUsModal', data: {
-                              bs_toggle: 'modal',
-                              bs_target: '#contactUsModal' }) %>
-  </p>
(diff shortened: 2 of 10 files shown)


justinlittman:
image

app/javascript/controllers/complex_radio_controller.js
```diff
@@ -11,7 +11,7 @@ export default class extends Controller {
     this.selectionTargets.forEach((target) => {
       const checked = target.querySelector("input[type='radio']").checked
       // If radio is checked, enable its child select elements; if unchecked, disable.
-      target.querySelectorAll('select,input[type="number"],input[type="checkbox"]').forEach((element) => element.disabled = !checked)
+      target.querySelectorAll('select,input[type="number"],input[type="checkbox"],fieldset').forEach((element) => element.disabled = !checked)
```
jcoyne, line 14: Cool. I didn't know a fieldset could be disabled: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset#attr-disabled

jcoyne (approved):
Looks good, but I believe Amy asked to review this before merging.

PR #2536, opened 2022-08-29
make spec less flaky

peetucket:

Noticed this spec was occasionally flaky based on the ordering of the array.  Since that doesn't seem to matter for functionality, can just sort in the spec and then compare with sorted version instead.

CI

### DIFF
--- a/spec/requests/create_collection_spec.rb
+++ b/spec/requests/create_collection_spec.rb
@@ -193,9 +193,10 @@ RSpec.describe 'Create a collection' do
             collection = Collection.last
             expect(response).to have_http_status(:found)
             expect(response).to redirect_to(collection_path(collection))
-            expect(collection.managed_by.map(&:sunetid)).to eq ['maya.aguirre', 'jcairns']
-            expect(collection.reviewed_by.map(&:email)).to eq %w[maya.aguirre@stanford.edu
-                                                                 jcairns@stanford.edu faridz@stanford.edu]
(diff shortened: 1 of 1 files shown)


spec/requests/create_collection_spec.rb
```diff
@@ -193,9 +193,9 @@
             collection = Collection.last
             expect(response).to have_http_status(:found)
             expect(response).to redirect_to(collection_path(collection))
-            expect(collection.managed_by.map(&:sunetid)).to eq ['maya.aguirre', 'jcairns']
-            expect(collection.reviewed_by.map(&:email)).to eq %w[maya.aguirre@stanford.edu
-                                                                 jcairns@stanford.edu faridz@stanford.edu]
+            expect(collection.managed_by.map(&:sunetid).sort).to eq ['jcairns', 'maya.aguirre']
```
jcoyne, line 196: Rspec has built in function, `contain_exactly` that you can use for this case.

jcoyne (changes requested):

spec/requests/create_collection_spec.rb
peetucket: changed

jcoyne (approved):

jcoyne:
@peetucket I wonder if rebasing this would help with the flakyness. It's about 10 commits behind main.

peetucket:
rebased

PR #2538, opened 2022-08-30
tweak duplicate keyword javascript error message

### LINKED ISSUE #2518: Update error text for duplicate keyword entry
The error text here should read: 
Each keyword must be unique. Enter a different keyword.

It currently says something else.

ndushay:

Fixes #2518

### AFTER

image

### BEFORE

image


jcoyne (changes requested):
Does this still give an appropriate error message when the keyword is missing?

ndushay:
@jcoyne I didn't change the behavior at all;  I just changed the error text.   The app works the same as before for empty keywords -- it removes the field if it is empty.

jcoyne (approved):

PR #2547, opened 2022-08-30
moreTypes no longer includes exact matches to work type ('Data', 'Image', 'Text')

### LINKED ISSUE #2281: Update work subtype lists in work type modal
Entering a work subtype that is the same as the work type causes duplications in the metadata displays. This is confusing to users and not helpful or needed in the metadata, so we should prevent this from happening.

See https://purl.stanford.edu/kn066mv7984.

Screen Shot 2022-06-08 at 2 18 23 PM

For each work type, do not display the subtype option that is the same as the work type, both in the work type modal and in the item edit form.

For example, for work type "data," the term "data" should not be in the subtype list in the modal or the form:

Screen Shot 2022-06-08 at 2 19 42 PM

Screen Shot 2022-06-08 at 2 20 07 PM

ndushay:

Fixes #2281

### AFTER

image

image

### BEFORE

Screen Shot 2022-06-08 at 2 19 42 PM

Screen Shot 2022-06-08 at 2 20 07 PM

### DIFF
--- a/app/components/works/subtypes_component.rb
+++ b/app/components/works/subtypes_component.rb
@@ -34,7 +34,7 @@ module Works
     end
 
     def more_types
-      WorkType.more_types - subtypes
+      WorkType.more_types - subtypes - [work_type.titlecase]
     end
 
     def tooltip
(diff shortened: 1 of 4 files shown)


ndushay:
@jcoyne you reviewed this before I was able to fix it.  Now the PR addresses both.

jcoyne (approved):

PR #2549, opened 2022-08-30
Different Cancel link for first draft

### LINKED ISSUE #2290: For initial collection form, cancel button should go to Details view page
When in the full collection editing form for an initial version of a collection (one that has never been deposited), if the user clicks the "Cancel" button at the bottom of the page, they should be taken to the Collection Details view page. Currently they are taken back to the dashboard.

lwrubel:
Resolves #2290. 

Unit.


lwrubel:
Before I get too far along with this, checking for some feedback on this approach. Is the duplication of the create button_component code too much?

mjgiarlo (approved):

PR #2551, opened 2022-08-30
add debounce to avoid too many sunet lookups

peetucket:

Could help with #2524 - sunet lookup issues for collection partipants/owner changes.

I saw this issue too when running the integration tests, and one theory is that we are running too many queries too quickly against the sunet lookup API.  This adds a debounce, so that we won't execute more than one query every 700 milliseconds.  That seems like a reasonable delay that is not so long as to make you think something is not working, but long enough to keep it from firing with each keypress.

Localhost browser

### DIFF
--- a/app/javascript/controllers/collection_participant_form_controller.js
+++ b/app/javascript/controllers/collection_participant_form_controller.js
@@ -1,4 +1,5 @@
 import { Controller } from "stimulus"
+import debounce from 'lodash.debounce'
 
 export default class extends Controller {
   static targets = ["addItem", "template", "control", "lookup",
@@ -8,6 +9,7 @@ export default class extends Controller {
   static values = { selector: String }
 
   connect() {
+    this.search = debounce(this.search, 700)
     this.controlTarget.hidden = true
     this.hideResult()
   }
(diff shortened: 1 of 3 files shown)


app/javascript/controllers/sunet_input_controller.js
```diff
@@ -1,10 +1,12 @@
 import { Controller } from "stimulus"
+import debounce from 'lodash.debounce'
 
 export default class extends Controller {
-  static targets = ["result", "resultNone", "resultOne", "resultName", "resultDescription", "queryValue", 
+  static targets = ["result", "resultNone", "resultOne", "resultName", "resultDescription", "queryValue",
   "resultError", "errorValue", "submit"]
-  
+
```
peetucket, line 7: this is just whitespace cleanup

app/javascript/controllers/collection_participant_form_controller.js
```diff
@@ -1,4 +1,5 @@
 import { Controller } from "stimulus"
+import debounce from 'lodash.debounce'
```
justinlittman, line 2: Should this be added to `package.json`?

peetucket: It worke fine for me on localhost without explicitly adding it, but I just added it now in case we need to deploy.

justinlittman: Yeah, it is already included transitively. This just makes it explicit.

justinlittman (approved):

PR #2552, opened 2022-08-30
adjust home page text

### LINKED ISSUE #2273: Update text in center of home page
On the home page at sdr.stanford.edu, update the text as follows:

From: Make a deposit today!
To: Go to the SDR Dashboard
Add subtext: to create or manage deposits

From [on button]: Start here [right arrow] 
To: Enter here [right arrow]

image

peetucket:

Fixes #2273 - update home page text

Localhost and updated test

### DIFF
--- a/app/views/welcome/show.html.erb
+++ b/app/views/welcome/show.html.erb
@@ -8,9 +8,10 @@
     </div>
     <p class="fs-4">Long term preservation of scholarly works at Stanford</p>
     <div class="text-center text-md-left way-down">
-      <h2 class="mb-5">Make a deposit today!</h2>
+      <h2>Go to the SDR Dashboard</h2>
+      <h3 class="mb-4">to create or manage deposits</h3>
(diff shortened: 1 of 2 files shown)


app/views/welcome/show.html.erb
```diff
@@ -8,9 +8,10 @@
     </div>
     <p class="fs-4">Long term preservation of scholarly works at Stanford</p>
     <div class="text-center text-md-left way-down">
-      <h2 class="mb-5">Make a deposit today!</h2>
+      <h2>Go to the SDR Dashboard</h2>
+      <h4 class="mb-4">to create or manage deposits</h4>
```
jcoyne, line 12: where is h3? For accessibility reasons you shouldn't skip header levels

peetucket: Didn't realize this was an accessibility problem - I had made it an h4 since it better matched the sizing shown in the design, but I switched it to an h3 (it's not that far off).

justinlittman (approved):

PR #2554, opened 2022-08-30
split authors/contributors into two tables on details page

### LINKED ISSUE #2283: Make clear on item view form which entries are "Authors" and which are "Contributors"
These are all currently displayed in the order they were entered, but the user can't tell when it switches from "Authors to be included in the citation" to "Additional contributors."

In the case below, the first two were authors and the second two contributors:
Screen Shot 2022-06-08 at 2 25 44 PM

The original mockup showed that this should look like this:
Screen Shot 2022-06-08 at 2 26 05 PM
_However, this wouldn't be accessible, so instead split this into two tables, one for authors and one for contributors. (edit by jcoyne)_

peetucket:

Fixes #2283 - split authors and contributors into two tables on detail page

Opted to create a new component since each table is basically the same.

Localhost and CI

### DIFF
--- a/app/components/works/show/authors_contributors_component.html.erb
+++ b/app/components/works/show/authors_contributors_component.html.erb
@@ -1,0 +1,20 @@
+  <table id="<%= participant_type %>s" class="table table-sm caption-header">
+    <caption><%= participant_type.titlecase %>
+      <%= helpers.turbo_frame_tag dom_id(work, 'edit_author'), src: edit_button_work_path(work, tag: 'author'), target: '_top' %>
+    </caption>
+    <tbody>
+    <% participants.each do |participant| %>
+      <tr>
+        <% if participant.person? %>
+          <td style="width: 50%">
+            <%= participant.first_name %> <%= participant.last_name %>
+          </td>
+        <% else %>
+          <td style="width: 50%"><%= participant.full_name %></td>
+        <% end %>
+        <td style="width: 25%"><%= render Works::Show::OrcidComponent.new(orcid: participant.orcid) %></td>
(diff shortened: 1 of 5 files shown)


app/components/works/show/authors_contributors_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        <% if participant.person? %>
+          <td style="width: 50%">
+            <%= participant.first_name %> <%= participant.last_name %>
+          </td>
+        <% else %>
+          <td width="50%"><%= participant.full_name %></td>
+        <% end %>
+        <td style="width: 25%"><%= render Works::Show::OrcidComponent.new(orcid: participant.orcid) %></td>
+        <td style="width: 25%"><%= participant.role %></td>
+      </tr>
```
peetucket, line 17: made columns fixed with or else they may line up the same between the two tables (e.g the "contributor type" column will align differently between the two tables)

app/components/works/show/authors_contributors_component.html.erb
```diff
@@ -0,0 +1,20 @@
...
+    </caption>
+    <tbody>
+    <% participants.each do |participant| %>
+      <tr>
+        <% if participant.person? %>
+          <td style="width: 50%">
+            <%= participant.first_name %> <%= participant.last_name %>
+          </td>
+        <% else %>
+          <td width="50%"><%= participant.full_name %></td>
```
jcoyne, line 13: Should this be `style="width: 50%"`?

peetucket: Whoops, indeed.  Fixed.

justinlittman (approved):

PR #2560, opened 2022-08-31
Adjust event descriptions for collections.

### LINKED ISSUE #2303: Add additional descriptions to collections history table for some events and update one description
Some user actions are not generating descriptions in the right-most column of the collections history table on the view page. Some of these are specific to when a user saves a draft or deposits, and sometimes it needs to be added for only one of these (because it works fine now for the other button). I would like to have the following added in:

- when contact is removed and item is saved as a draft: "contact email modified"
- when related link/link text are removed and collection is saved as draft OR deposited: "related links modified"
- when release setting is changed: "release settings modified"
- when download (visibility) setting is changed: "download setting modified"
- when DOI assignment setting is changed: DOI setting modified"
- when license setting is changed: "license setting modified"
- when either or both of the boxes for notifications to managers or collection participants are checked or unchecked: "notification settings mnodified"
- when review workflow is turned on or off: "review workflow setting modified". 

Note that currently if there are reviewers and the review workflow is turned off, the description says "Removed reviewers: [SUNet ID]". In this case, this text should remain and the above "review workflow setting modified" text should appear as well.

Also, when a related link is added, for some reason the descriptive text says both "related links modified" and "contact email modified". The comment about the contact email should be removed.

justinlittman:
closes #2303

Better event descriptions.

Unit

### DIFF
--- a/app/models/collection_change_set.rb
+++ b/app/models/collection_change_set.rb
@@ -44,6 +44,22 @@ class CollectionChangeSet
       removed_reviewers).uniq
   end
 
+  def email_depositors_status_changed_changed?
+    point1.email_depositors_status_changed != point2.email_depositors_status_changed
+  end
+
+  def email_when_participants_changed_changed?
+    point1.email_when_participants_changed != point2.email_when_participants_changed
+  end
+
+  def review_enabled_changed?
+    point1.review_enabled != point2.review_enabled
+  end
+
+  def reviewers_changed?
+    removed_reviewers.present? || added_reviewers.present?
+  end
+
   def participant_change_description
     %i[added_managers added_depositors added_reviewers
        removed_managers removed_depositors removed_reviewers].filter_map do |field_name|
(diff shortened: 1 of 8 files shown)


spec/services/collection_change_set_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+require 'rails_helper'
+
+RSpec.describe CollectionChangeSet do
+  let(:collection1) { create(:collection) }
+  let(:collection2) do
+    create(:collection, email_when_participants_changed: true, email_depositors_status_changed: true,
+                        review_enabled: true)
+  end
+
+  describe '#email_depositors_status_changed_changed?' do
```
ndushay, line 12: changed_changed?

app/models/collection_change_set.rb
```diff
@@ -44,6 +44,22 @@ def changed_participants
       removed_reviewers).uniq
   end
 
+  def email_depositors_status_changed_changed?
```
ndushay, line 47: changed_changed?

ndushay (commented):
I would think Amy wants to review this, if she hasn't already.   A couple of concerns for your consideration.

spec/services/collection_change_set_spec.rb
justinlittman: Unfortunately. The field is called `email_depositors_status_changed` so checking whether this field has changed appends `_changed?`.

app/models/collection_change_set.rb
justinlittman: Unfortunately. The field is called `email_depositors_status_changed` so checking whether this field has changed appends `_changed?`.

justinlittman:
@amyehodge did not request to review this.

jcoyne (approved):

PR #2568, opened 2022-09-01
author should be first in the list

peetucket:

The author should be the default in the contributor role drop-down select list and this is the easy way, by just making it first in the list, even though it is slightly out of alphabetical order.

Verified as ok with PO
Screen Shot 2022-09-01 at 9 36 03 AM

Localhost and CI

### DIFF
--- a/app/models/abstract_contributor.rb
+++ b/app/models/abstract_contributor.rb
@@ -6,9 +6,10 @@
 class AbstractContributor < ApplicationRecord
   SEPARATOR = '|'
 
+  # NOTE: "Author" is deliberately set to first (out of alpha order), as it should be the default
   PERSON_ROLES = [
-    'Advisor',
     'Author',
+    'Advisor',
     'Composer',
     'Contributing author',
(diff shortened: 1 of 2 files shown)


app/models/abstract_contributor.rb
```diff
@@ -7,8 +7,8 @@ class AbstractContributor < ApplicationRecord
   SEPARATOR = '|'
 
   PERSON_ROLES = [
-    'Advisor',
     'Author',
```
mjgiarlo, line 10: What would you think about dropping a comment here that we expect `Author` to **remain** the first element in this array?

mjgiarlo (approved):
Approve pending consideration of comment

app/models/abstract_contributor.rb
peetucket: Sounds good - comment added.

mjgiarlo: @peetucket Thx! Feel free to merge when 💚

PR #2569, opened 2022-09-01
Render work subtypes on edit form like the new form

### LINKED ISSUE #2282: Subtypes list in item edit form should be alphabetized down, then across
This list displays alphabetically across first, and then down.
Screen Shot 2022-06-08 at 2 23 21 PM

It should display like it does in the work types modal, which is down first, then across:

Screen Shot 2022-06-08 at 2 24 02 PM

mjgiarlo:

Fixes #2282

This commit tweaks CSS classes and HTML structure in the work subtypes component so that subtypes (and "more types") in the work edit form render as vertically oriented checkboxes as opposed to horizontally oriented checkboxes.

CI, local testing, and tested successfully by PO in stage

No tests added since this affects only visual rendering of page.

### DIFF
--- a/app/components/works/subtypes_component.html.erb
+++ b/app/components/works/subtypes_component.html.erb
@@ -5,10 +5,10 @@
   </div>
 </div>
 
-<fieldset>
+<fieldset class="mb-4 work-types"  data-controller="more-types">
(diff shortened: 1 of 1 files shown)


ndushay (approved):
before and after screenshots are superhelpful.

PR #2571, opened 2022-09-01
Add appropriate collection link to dashboard.

### LINKED ISSUE #2271: Need another collection workflow: add link to dashboard
**If you pick up this ticket, you may also want to pick up https://github.com/sul-dlss/happy-heron/issues/2272.**

**Please release this to production with #2272**

The "Need another collection" workflow is intended to help existing SDR users understand that there may be another collection than the ones they currently see that would be the right place for them to deposit their work. In particular, we want to direct them to a small set of general purpose collections. See also #2272.

At the bottom of the list of "Your collections" on the dashboard, add text that says "Don't see an appropriate collection?" This should be a link that opens up the help modal with the new "Request access to another collection" option already selected in the "How can we help you" drop-down and displays the list of options with check boxes (see #2272). All users should see this link every time they are on the dashboard.

Screen Shot 2022-06-08 at 12 45 18 PM

justinlittman:
closes #2271

Unit, Amy

### DIFF
--- a/app/controllers/helps_controller.rb
+++ b/app/controllers/helps_controller.rb
@@ -4,7 +4,11 @@
 class HelpsController < ApplicationController
   def new
     @email = current_user&.email
-    @help_how_value = 'I want to become an SDR depositor' unless current_user
+    @help_how_value = if params[:show_collections] == 'true'
+                        'Request access to another collection'
+                      elsif !current_user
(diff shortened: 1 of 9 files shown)


jcoyne (approved):
consider switching data attributes, but otherwise looks good.

justinlittman:
@amyehodge approved/

PR #2582, opened 2022-09-01
Use HTML label from `WorkType` class for deposit type field

### LINKED ISSUE #2479: On item show page, change "Deposit type" field to use the label in 'WorkType'

mjgiarlo:

Fixes #2479

This ensures the work type label that is rendered in the UI matches the value the user would expect, the same as they see in the work type selection modal. E.g., `Software/Code` instead of `software, multimedia`, and `Text` instead of `text`.

CI + local testing


ndushay (approved):
before and after screenshots are super helpful.

PR #2584, opened 2022-09-01
Add version description to event history tables

### LINKED ISSUE #2346: Add version comments to history table on item view page
Version comments entered by the user should display on the relevant "Deposit completed" line of the history table, prefaced by the text "What changed: "

### LINKED ISSUE #2347: Add version comments to history table on collection view page
Version comments entered by the user should display on the relevant "Deposit completed" line of the history table, prefaced by the text "What changed: "

See #2346 for mockup.

mjgiarlo:

Fixes #2346
Fixes #2347

This commit injects the version description into the event context for both works and collections, so that the user-provided description is surfaced in work and collection event history.

Thanks for working on this and using the more idiomatic approach, @lwrubel!

Also includes:
* Align collection event history component with work event history component

### Before

Screenshot from 2022-09-01 15-53-21

### After

Screenshot from 2022-09-01 15-48-25

CI, local dev, QA

### DIFF
--- a/lib/tasks/deposit.rake
+++ b/lib/tasks/deposit.rake
@@ -1,20 +1,28 @@
 # frozen_string_literal: true
 
+# NOTE: This task is largely copypasta from the DepositStatusJob, with the logic
+#       divorced from any messaging or background job frameworks. We could
+#       consider extracting the common bits to a service?
 desc 'Complete deposit of works and collections (only for development)'
 task complete_deposits: :environment do
   abort 'ERROR: This task only runs in the development environment!' unless Rails.env.development?
 
-  objects_awaiting_deposit.each do |object|
+  objects_awaiting_deposit.each do |object_version|
     druid = random_druid
-    parent = case object
+    parent = case object_version
              when CollectionVersion
-               object.collection
+               object_version.collection
              when WorkVersion
-               object.work
+               object_version.work
              end
     parent.update(druid: druid)
-    object.deposit_complete!
-    puts "Marked #{object.class} id=#{object.id} as deposited with #{druid}"
+
+    what_changed = object_version.version_description.presence || 'not specified'
+    # NOTE: This user is created when seeding the database, a common practice running in dev.
+    parent.event_context = { user: User.find_by(name: 'SDR'), description: "What changed: #{what_changed}" }
+
+    object_version.deposit_complete!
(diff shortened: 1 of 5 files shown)


app/models/collection_version.rb
```diff
@@ -17,6 +17,10 @@ def accessioned?
   state_machine initial: :new do
     before_transition do |collection_version, transition|
       event_params = collection_version.collection.event_context
+      if transition.event == :deposit_complete
+        what_changed = collection_version.version_description.presence || 'not specified'
+        event_params[:description] = "What changed: #{what_changed}"
+      end
```
mjgiarlo, line 23: @jcoyne Is there a better way, or a better place, to do this? I'm rusty.

spec/models/collection_version_spec.rb
```diff
@@ -144,6 +144,7 @@
           .to change(collection_version, :state)
           .to('deposited')
           .and change(Event, :count).by(1)
+        expect(collection_version.collection.events.first.description).to eq('What changed: not specified')
```
mjgiarlo, line 147: Better way to write this? 🤷🏻‍♂️

lib/tasks/deposit.rake
```diff
@@ -1,20 +1,28 @@
 # frozen_string_literal: true
 
+# NOTE: This task is largely copypasta from the DepositStatusJob, with the logic
+#       divorced from any messaging or background job frameworks. We could
+#       consider extracting the common bits to a service?
```
mjgiarlo, line 5: Please note this note. I'd be happy to try my hand at this in a follow-up PR. Also happy to let it sit a while.

jcoyne (approved):

lib/tasks/deposit.rake
ndushay: 2 repetitions is too few to justify extracting out a class.  Sandi Metz suggested even 3 is too few.

ndushay (approved):

lib/tasks/deposit.rake
mjgiarlo: I strongly agree with this guideline generally. In this case, I am suspicious of logic that lives *only* in jobs or in tasks because they are so coupled to how they are invoked (`ActiveJob` and `rake` respectively), which is why I'd be happy to move that logic to a service class that can be used in both contexts.

PR #2585, opened 2022-09-02
Metadata only deposits.

### LINKED ISSUE #2570: Metadata only deposit
When a new version is made in h2, but no files are changed, then do the deposit to the sdr-api using a metadata only deposit.

This will enable #2370

justinlittman:
closes #2570

Always having to deposit files is expensive.

Unit, QA

### DIFF
--- a/app/jobs/deposit_job.rb
+++ b/app/jobs/deposit_job.rb
@@ -4,19 +4,15 @@
 class DepositJob < BaseDepositJob
   queue_as :default
 
-  # rubocop:disable Metrics/AbcSize:
   def perform(work_version)
     Honeybadger.context({ work_version_id: work_version.id, druid: work_version.work.druid,
                           work_id: work_version.work.id, depositor_sunet: work_version.work.depositor.sunetid })
 
     request_dro = CocinaGenerator::DROGenerator.generate_model(work_version: work_version)
-    blobs = work_version.attached_files.map { |af| af.file.attachment.blob }
 
-    login_result = login
-    raise login_result.failure unless login_result.success?
+    perform_login
 
-    new_request_dro = SdrClient::Deposit::UpdateDroWithFileIdentifiers.update(request_dro: request_dro,
-                                                                              upload_responses: upload_responses(blobs))
+    new_request_dro = update_dro_with_file_identifiers(request_dro, work_version)
 
     case new_request_dro
     when Cocina::Models::RequestDRO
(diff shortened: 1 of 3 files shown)


app/jobs/deposit_job.rb
```diff
@@ -76,4 +76,32 @@ def build_file_metadata(blobs)
...
+
+  def metadata_only?(work_version)
+    previous_version = previous_version_for(work_version)
+    return false unless previous_version
+
+    blob_map_for(work_version) == blob_map_for(previous_version)
+  end
+
+  def previous_version_for(work_version)
+    work_version.work.work_versions.find { |check_work_version| check_work_version.version = work_version.version - 1 }
```
jcoyne, line 92: How about moving this to the WorkVersion class?

spec/jobs/deposit_job_spec.rb
```diff
@@ -26,57 +42,133 @@
...
     before do
       allow(SdrClient::Deposit::UpdateResource).to receive(:run).and_return(1234)
     end
 
-    it 'calls UpdateResource.run' do
-      described_class.perform_now(work_version)
-      expect(SdrClient::Deposit::UpdateResource).to have_received(:run)
-        .with(a_hash_including(version_description: 'Updated metadata'))
+    context 'when files have not changed' do
+      # All that matters in this cocina is structural > contains.
```
jcoyne, line 87: I don't understand this. Can you make it clearer?

spec/jobs/deposit_job_spec.rb
```diff
@@ -13,11 +13,27 @@
...
+  end
   let(:attached_file) { build(:attached_file) }
+  let(:attached_file2) { build(:attached_file) }
   let(:work) { build(:work, collection: collection, assign_doi: false) }
-  let(:work_version) do
-    build(:work_version, id: 8, work: work, attached_files: [attached_file], version_description: 'Updated metadata')
+  let(:first_work_version) do
+    build(:work_version, work: work, attached_files: [attached_file], version: 1)
   end
+  let(:second_work_version_metadata_only) do
```
jcoyne, line 29: Can this be moved down into the context "when files have not changed"?

jcoyne (changes requested):

app/jobs/deposit_job.rb
justinlittman: Done.

spec/jobs/deposit_job_spec.rb
justinlittman: Done.

spec/jobs/deposit_job_spec.rb
justinlittman: Additional comments added.

PR #2593, opened 2022-09-02
simplify recent activity on admin page

### LINKED ISSUE #2567: Simplify recent activity tables for items and collections
After seeing what these look like with the actual production data, @hannahfrost, @andrewjbtw and I have agreed that they are a bit too busy to be as useful as they could be. We are proposing the following simplification:

1. For both tables, remove the columns with the event information and the date/time information. 
-- For the collection table, this will leave just a list of collection titles (linked)
-- For the item table, this will leave a list of item titles (linked) and the titles of the collections they belong to (linked)
2. Display a *deduplicated* list of items or collections, depending on the table, for which an event has been logged in the time period selected.

This will allow us to easily scan a short list of items or collections that have been touched in some way without missing something that has only a single entry amongst the ones that 2 or 3 or 7 entries. All the detailed information about the events and dates are easily accessible with one click from the item or collection view pages.

peetucket:

Fixes #2567 - simplify recent activity tables in admin interface

Note: we are also simplifying the query and ignoring events, and just using the "last_updated" column on the work/collection versions, which means we will consider even a change to a title as a recent activity.  This required adding some hooks to ensure the updated_at column is touched when any bits of metadata in associated tables are changed.

Also, use title presenters for works and collections so that those in draft states and/or without entered titles will display correctly in the tables (e.g. they will show a linked "No title" instead of a URL or blank row).

Localhost

### DIFF
--- a/app/controllers/admins_controller.rb
+++ b/app/controllers/admins_controller.rb
@@ -13,12 +13,13 @@ class AdminsController < ApplicationController
   def items_recent_activity
     authorize! :admin_dashboard
     @days_limit = params.fetch(:days, 7).to_i
-    @events = Event.work_events.includes(:eventable).where('created_at > ?', @days_limit.days.ago)
+    @items = Work.joins(:head).where('work_versions.updated_at > ?',
+                                     @days_limit.days.ago).order('work_versions.updated_at DESC')
   end
 
   def collections_recent_activity
     authorize! :admin_dashboard
     @days_limit = params.fetch(:days, 7).to_i
-    @events = Event.collection_events.includes(:eventable).where('created_at > ?', @days_limit.days.ago)
+    @collections = Collection.where('updated_at > ?', @days_limit.days.ago).order('updated_at DESC')
   end
 end
--- a/app/components/admin/collection_activity_row_component.html.erb
+++ b/app/components/admin/collection_activity_row_component.html.erb
@@ -1,5 +1,3 @@
 <tr>
   <td><%= link_to collection_title, collection, target: '_top' %></td>
-  <td><%= action %></td>
-  <td><%= render LocalTimeComponent.new(datetime: event_date) %></td>
 </tr>
--- a/app/components/admin/collection_activity_row_component.rb
+++ b/app/components/admin/collection_activity_row_component.rb
@@ -3,28 +3,16 @@
 module Admin
   # Renders a list of all collection events
   class CollectionActivityRowComponent < ApplicationComponent
-    with_collection_parameter :event
+    with_collection_parameter :collection
 
-    def initialize(event:)
-      @event = event
+    def initialize(collection:)
+      @collection = collection
     end
 
-    attr_reader :event
-
-    def collection
-      event.eventable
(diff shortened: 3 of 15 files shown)


jcoyne (changes requested):
This seems to only take care of the first bullet point on #2567. Are you planning on doing the second part here too?

peetucket:
Oops, missed the second bullet point.  I can have a look at that too.

Gemfile
```diff
@@ -64,6 +64,7 @@ gem 'lograge', '~> 0.11.2'
 gem 'okcomputer'
 gem 'pg'
 gem 'propshaft'
+gem 'pry-rails'
 gem 'redis', '~> 4.0'
```
peetucket, line 68: This is preferable to irb when debugging -- I was having some odd problems when testing queries.

app/controllers/admins_controller.rb
```diff
@@ -13,12 +13,12 @@ def show
...
-    @events = Event.work_events.includes(:eventable).where('created_at > ?', @days_limit.days.ago)
+    @items = Event.work_events.where('created_at > ?', @days_limit.days.ago).pluck(:eventable_id).uniq
   end
 
   def collections_recent_activity
     authorize! :admin_dashboard
     @days_limit = params.fetch(:days, 7).to_i
-    @events = Event.collection_events.includes(:eventable).where('created_at > ?', @days_limit.days.ago)
+    @collections = Event.collection_events.where('created_at > ?', @days_limit.days.ago).pluck(:eventable_id).uniq
   end
```
peetucket, line 23: I tried for a while and couldn't get a working SQL query that did a distinct on the `eventable_id` and returned what we needed.  If it's possible, this would be preferable to what I am doing here, which is just fetching the `eventable_ids` and then uniquing the array, since this requires loading the items/collections later in the compontent (which triggered a bunch of changes as seen above).

jcoyne: Is it possible to just look at updated_at on the collection version? I don't think we have any  events that don't update the object? That would certainly make for a simpler query.

Gemfile
jcoyne: Should this go in the development group?  Should we consider this change in a separate PR?

jcoyne (changes requested):

jcoyne:
Please do a rebase on this PR.

Gemfile
peetucket: Yeah, I'll pull into a separate PR.  My first inclination was also to put in the development group, though other apps seem to not do this, e.g.

https://github.com/sul-dlss/workflow-server-rails/blob/main/Gemfile#L17
https://github.com/sul-dlss/pre-assembly/blob/main/Gemfile#L13

and others: https://github.com/search?q=org%3Asul-dlss+pry-rails&type=code

peetucket:
Rebased from main and squashed all commits in this PR.  Split out adding of pry-rails gem into new PR #2607

jcoyne (approved):

PR #2594, opened 2022-09-02
Adding version_description to controller

lwrubel:

Fixes bug introduced in PR #2576 resulting in version_description not being saved for works.

Needs integration testing and unit test check. 


ndushay:
I tried to write a test for this and failed.  If needed, a test can be added later.   (I tried to add it to features/edit_draft_work_spec.rb, which has a click_button 'Deposit' but it seems to only save the draft.  Clearly I'm missing something.  I will see if I can add it to H2 integration test

peetucket (approved):

PR #2606, opened 2022-09-06
Prevent orphaning embargoed items when switching collection release o…

### LINKED ISSUE #2542: Do not allow manager to change release setting to "immediately" if the collection contains embargoed items.
**See the comment below that describes how this should work.**

An item with an embargo should never be automatically switched to immediate release. The embargo should remain intact. 

This situation happens in the following scenario:
1. An item is created and assigned an embargo. The collection must be set for either requiring an embargo or allowing the user to choose to get an embargo.
2. The collection setting is updated so that immediate release is required.
3. The user versions their item and the item is immediately released. 

In this case the user is shown this in the UI but is not alerted that a change is being made and is not allowed to choose to maintain their embargo:

The user would not expect this to happen and is unlikely to notice this text in the UI.

justinlittman:
…ption to immediate.

closes #2542

Avoid user surprise.

Unit, Amy

### DIFF
--- a/app/components/collections/settings_component.html.erb
+++ b/app/components/collections/settings_component.html.erb
@@ -4,12 +4,12 @@
 
     <fieldset class="mb-5 release-option" data-controller="complex-radio">
       <legend class="h5">When will files on deposits to this collection be downloadable? * <%= render PopoverComponent.new key: 'collection.release' %></legend>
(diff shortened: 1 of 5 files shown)


justinlittman:
image

justinlittman:
@amyehodge approved.

jcoyne (approved):

PR #2607, opened 2022-09-06
add pry to gemfile for debugging

peetucket:

Add pry to Gemfile for debugging purposes

Split from #2593

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,7 @@ gem 'lograge', '~> 0.11.2'
 gem 'okcomputer'
 gem 'pg'
 gem 'propshaft'
+gem 'pry'
 gem 'redis', '~> 4.0'
 # TODO: Deal with this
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -64,6 +64,7 @@ gem 'lograge', '~> 0.11.2'
 gem 'okcomputer'
 gem 'pg'
 gem 'propshaft'
+gem 'pry-rails'
```
jcoyne, line 67: This project is not actively maintained. It doesn't have tests for Rails 7 and it has no CI builds. See https://github.com/pry/pry-rails#readme  Are we sure this is a good dependency to add?

peetucket: Well, that's not good.  I definitely ran into some issues with irb when doing debugging work on the rails console for #2593 that worked fine with pry, but perhaps team discussion is warranted (or at minimum putting in the development group here).

jcoyne (approved):

PR #2609, opened 2022-09-06
Extract deposit completer class from task and job

mjgiarlo:

This commit better aligns how local dev handles completing deposits and how it is done in deployed environments. It reduces potential drift.

CI, local, QA

### DIFF
--- a/lib/tasks/deposit.rake
+++ b/lib/tasks/deposit.rake
@@ -1,27 +1,14 @@
 # frozen_string_literal: true
 
-# NOTE: This task is largely copypasta from the DepositStatusJob, with the logic
-#       divorced from any messaging or background job frameworks. We could
-#       consider extracting the common bits to a service?
 desc 'Complete deposit of works and collections (only for development)'
 task complete_deposits: :environment do
   abort 'ERROR: This task only runs in the development environment!' unless Rails.env.development?
 
   objects_awaiting_deposit.each do |object_version|
-    druid = random_druid
-    parent = case object_version
-             when CollectionVersion
-               object_version.collection
-             when WorkVersion
-               object_version.work
-             end
-    parent.update(druid: druid)
-
-    what_changed = object_version.version_description.presence || 'not specified'
-    # NOTE: This user is created when seeding the database, a common practice running in dev.
-    parent.event_context = { user: User.find_by(name: 'SDR'), description: "What changed: #{what_changed}" }
-
-    object_version.deposit_complete!
+    deposit_completer = DepositCompleter.new(object_version: object_version)
+    druid = deposit_completer.parent.druid.presence || random_druid
+    deposit_completer.parent.update(druid: druid)
+    deposit_completer.complete
     puts "Marked #{object_version.class} id=#{object_version.id} as deposited with #{druid}"
   end
 end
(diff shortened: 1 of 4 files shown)


app/services/depositor.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+# Completes the deposit of an object
+class Depositor
```
justinlittman, line 4: Can we call this `DepositCompleteService` or smiliar? When I first read this, I thought it was referring to the person who was performing the deposit (like Contributor or Creator). And then I thought it was performing the deposit. Neither of which are correct.

app/services/depositor.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  def what_changed
+    object_version.version_description.presence || 'not specified'
+  end
+
+  def verify_depositing!
+    return if object_version&.depositing?
+
+    raise NotDepositingError, "Cannot deposit because state is not `depositing` (`#{object_version.state}`)"
```
justinlittman, line 47: This error message is confusing. The object is already deposited.

justinlittman (changes requested):

lib/tasks/deposit.rake
```diff
@@ -8,20 +8,10 @@ task complete_deposits: :environment do
   abort 'ERROR: This task only runs in the development environment!' unless Rails.env.development?
```
lwrubel, line 8: Since there is now a service (and less copypasta) the comment could be updated.

app/services/depositor.rb
mjgiarlo: Yes, good call.

app/services/depositor.rb
mjgiarlo: @justinlittman I'm open to suggestions. FWIW, as currently written, this exception message should never be surfaced in the app, as it's caught by the containing context.

lib/tasks/deposit.rake
mjgiarlo: Will do, thx!

mjgiarlo: I took a look at this, and I think we probably still want this task to only run in the development environment, since it forces depositing objects into the deposited state and jams in a fake druid. Does that make sense?

app/services/depositor.rb
mjgiarlo: @justinlittman I'm inclined to go with `DepositCompleter` if that sounds better to you.

app/services/depositor.rb
mjgiarlo: I'm dumping the exception altogether.

justinlittman (approved):

lib/tasks/deposit.rake
lwrubel: Yes, that makes sense. I was commenting more about the part that says "We could consider extracting the common bits to a service?" since that is what this PR does. Unless you think there is some other further work to do here. Your call.

mjgiarlo: Ah! GitHub UI confused my eyes. Yes, thank you, that defunct comment is now removed. Good eye. :)

PR #2616, opened 2022-09-07
Extract depositor class from task and job

mjgiarlo:
Reverts sul-dlss/happy-heron#2614

### DIFF
--- a/app/jobs/deposit_status_job.rb
+++ b/app/jobs/deposit_status_job.rb
@@ -21,6 +21,8 @@ class DepositStatusJob
     ActiveRecord::Base.connection_pool.with_connection do
       object = Work.find_by(druid: druid) || Collection.find_by(druid: druid)
 
+      return ack! unless object
+
       Honeybadger.context(object: object.to_global_id.to_s)
 
       DepositCompleter.complete(object_version: object.head)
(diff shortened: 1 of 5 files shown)


app/jobs/deposit_status_job.rb
```diff
@@ -22,22 +21,15 @@ def work(msg)
     ActiveRecord::Base.connection_pool.with_connection do
       object = Work.find_by(druid: druid) || Collection.find_by(druid: druid)
 
-      unless object && object.head.depositing? # rubocop:disable Style/SafeNavigation
-        # This guards against objects from a different project and prevents
-        # invalid transitions where the workflow was kicked off outside of h2.
-        return ack!
-      end
+      return ack! unless object
```
mjgiarlo, line 24: Here's the `ack!` that was missing from the prior PR.

spec/jobs/deposit_status_job_spec.rb
```diff
@@ -7,6 +7,15 @@
...
 
+  context 'with a not found druid' do
+    # An intentionally invalid druid so it does not collide with any test objects
+    let(:message) { '{"druid":"druid:aa11ii1111"}' }
+
+    it 'acks the message (and does not raise)' do
+      expect(run).to eq(:ack)
+    end
+  end
+
```
mjgiarlo, line 18: And here's a test that the above change fixes, which corresponds to the HB alert we were seeing in production

ndushay (approved):
I'm approving this but would appreciate @mjgiarlo watching honeybadger carefully for errors when it is deployed to prod.

PR #2619, opened 2022-09-07
this is a conditional check, not an assignment

### LINKED ISSUE #2618: Metadata-only deposit (globus checked, no files uploaded) throws an exception
The work: https://sdr.stanford.edu/works/5959

The problem: Amy submitted item without any files, but with the "globus" checkbox instead, throws HB alert.

## Backtrace

[happy-heron/prod] Cocina::Models::ValidationError: Type field not found

    line 100 of [PROJECT_ROOT]/app/jobs/deposit_job.rb: existing_structural_for
    line 81 of [PROJECT_ROOT]/app/jobs/deposit_job.rb: update_dro_with_existing_file_identifiers
    line 34 of [PROJECT_ROOT]/app/jobs/deposit_job.rb: update_dro_with_file_identifiers

[View full backtrace and more info at honeybadger.io](https://app.honeybadger.io/projects/77112/faults/88002665)

peetucket:

Will probably fix #2618, and may avert other bad stuff.

~~We should add a test for this too in a follow on PR.~~ Test added

New tests

### DIFF
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -193,6 +193,8 @@ class WorkVersion < ApplicationRecord
   end
 
   def previous_version
-    work.work_versions.find { |check_work_version| check_work_version.version = version - 1 }
+    return nil if version == 1 # shortcut any query checks if we are on the first version
+
+    work.work_versions.find { |check_work_version| check_work_version.version == version - 1 }
   end
 end
--- a/spec/models/work_version_spec.rb
+++ b/spec/models/work_version_spec.rb
@@ -60,6 +60,27 @@ RSpec.describe WorkVersion do
     end
   end
 
+  describe '#previous_version' do
+    let(:work_version) { build(:work_version, :with_work) }
+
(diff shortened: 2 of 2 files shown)


app/models/work_version.rb
```diff
@@ -193,6 +193,8 @@ def purl_reservation?
   end
 
   def previous_version
-    work.work_versions.find { |check_work_version| check_work_version.version = version - 1 }
+    return nil if version == 1 # shortcut any query checks if we are on the first version
+
+    work.work_versions.find { |check_work_version| check_work_version.version == version - 1 }
   end
```
peetucket, line 199: we were not doing a conditional check before, but instead just subtracting one each time, so always returning something even if we were on version 1

ndushay (approved):

justinlittman (approved):

jcoyne:
@peetucket please rebase this so the cypress tests run.

PR #2620, opened 2022-09-08
add a test for the #previous_version method

peetucket:

Follow on from #2619, adds a test for this method to catch that previous bug

CI, and it will fail until we merge this into #2619

(but it passes if run on that branch)


peetucket:
Just adding the test to #2619

PR #2630, opened 2022-09-08
Do not run SDR find in deposit job when no druid

mjgiarlo:

This commit allows the deposit job to proceed for metadata-only updates for objects that do not yet have a druid. If the object has a druid and the SDR find operation fails, an exception will be raised and reported to Honeybadger.

CI

### DIFF
--- a/app/jobs/deposit_job.rb
+++ b/app/jobs/deposit_job.rb
@@ -4,6 +4,7 @@
 class DepositJob < BaseDepositJob
   queue_as :default
 
+  # @raise [SdrClient::Find::Failed] if the (non-nil) druid cannot be found in SDR
   def perform(work_version)
     Honeybadger.context({ work_version_id: work_version.id, druid: work_version.work.druid,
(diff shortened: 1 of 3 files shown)


app/jobs/deposit_job.rb
```diff
@@ -99,5 +99,11 @@ def existing_structural_for(druid)
     cocina_json = JSON.parse(cocina_str)
     cocina = Cocina::Models.build(cocina_json)
     cocina.structural.contains
+  rescue SdrClient::Find::Failed
+    # This exception is rescued when the find operation fails.
+    #
+    # This is known to happen in H2 when dealing with an object version that
+    # does not yet have a druid. When this happens, return an empty list.
+    []
```
mjgiarlo, line 107: This may or may not be the right thing to do here. Reviews welcome!

jcoyne: Can we check for the druid before trying the find?  Then we can let this exception raise so that it shows up in HB?

jcoyne (changes requested):

jcoyne (approved):

PR #2634, opened 2022-09-08
do not say a file description changed if it is blank

### LINKED ISSUE #2633: Fix description changed message when no description.
When a file is added but no file description is included, the text reads "files added/removed, file description changed." This should just read "files added/removed."

peetucket:

Fixes #2633 - if a user uploads a file, but adds no description, we do not want to record the file description as changed (it currently says it's change even though it is blank, even though it technically "changed", so we just ignore if blank)

More importantly, I also noticed the the reform `changed?` method we use in `WorkVersionEventDescriptionBuilder` seems to indicate some attributes aren't changed, when they actually have.  I was having trouble getting this to work until I did some debugging and saw this.   I suspect that's because we are passing the form to this builder *after* we have persisted to the database, and perhaps this is altering the logic (update: but only for the update action for existing objects)

Event description updates for existing objects has been broken until now, and there didn't appear to be any tests to catch it.

This PR addresses this by building the context before the model is persisted and then passing it along to the method that needs it (for updates).

Localhost, a new spec to verify that event descriptions are recorded for updating existing works; existing CI

### DIFF
--- a/app/services/work_version_event_description_builder.rb
+++ b/app/services/work_version_event_description_builder.rb
@@ -22,7 +22,7 @@ class WorkVersionEventDescriptionBuilder
   end
 
   def title
-    'title of deposit modified' if form.changed?(:title)
+    'title of deposit modified' if form.changed?('title')
   end
 
   def abstract
@@ -109,6 +109,6 @@ class WorkVersionEventDescriptionBuilder
   end
 
   def assign_doi
-    'assign DOI modified' if form.changed?(:assign_doi)
+    'assign DOI modified' if form.changed?('assign_doi')
   end
 end
--- a/app/services/work_version_event_description_builder.rb
+++ b/app/services/work_version_event_description_builder.rb
@@ -80,6 +80,8 @@ class WorkVersionEventDescriptionBuilder
   end
 
   def files
+    return unless form.input_params
+
     attributes = form.input_params[:attached_files_attributes]
     return unless attributes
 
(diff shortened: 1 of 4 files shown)


app/services/work_version_event_description_builder.rb
```diff
@@ -80,6 +80,8 @@ def license
   end
 
   def files
+    return unless form.input_params
+
     attributes = form.input_params[:attached_files_attributes]
```
peetucket, line 85: `form.input_params` seems to be nil on unpersisted form with no files added (ie. new deposit where you just click "save as draft") -- this guards against it

app/services/work_version_event_description_builder.rb
```diff
@@ -99,7 +101,7 @@ def file_visibility
   end
 
   def file_description
-    'file description changed' if form.attached_files.any? { |af| af.changed?('label') }
+    'file description changed' if form.attached_files.any? { |af| af.changed?('label') && af.label.present? }
   end
```
peetucket, line 105: the actual change for this particular issue

app/services/work_version_event_description_builder.rb
```diff
@@ -22,7 +22,7 @@ def build # rubocop:disable Metrics/AbcSize
...
 
   def title
-    'title of deposit modified' if form.changed?(:title)
+    'title of deposit modified' if form.changed?('title')
   end
```
peetucket, line 26: for consistency

app/services/work_version_event_description_builder.rb
```diff
@@ -99,14 +101,14 @@ def file_visibility
...
 
   def related_works
     'related works modified' if form.changed?('related_works')
   end
 
   def assign_doi
-    'assign DOI modified' if form.changed?(:assign_doi)
+    'assign DOI modified' if form.changed?('assign_doi')
   end
```
peetucket, line 113: for consistency

jcoyne (approved):

PR #2636, opened 2022-09-09
Fix description message for new item without values.

### LINKED ISSUE #2632: Fix description message for new item without values
When a brand new item is saved without doing anything at all to the form, there shouldn't be anything displayed in the description. Right now this says "citation modified, visibility modified, license modified". Or at least it did for a new item in the collection I was using for testing. This may vary depending on the collection settings.

justinlittman:
closes #2632

Unit

### DIFF
--- a/spec/services/work_version_event_description_builder_spec.rb
+++ b/spec/services/work_version_event_description_builder_spec.rb
@@ -179,4 +179,85 @@ RSpec.describe WorkVersionEventDescriptionBuilder do
                            'files added/removed, file description changed, assign DOI modified'
     end
   end
+
+  context 'when a new work version and nothing has changed' do
+    let(:user) { build(:user) }
+    let(:collection) { build(:collection, required_license: 'CC0-1.0', release_option: 'immediate', access: 'world') }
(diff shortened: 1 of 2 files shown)


spec/services/work_version_event_description_builder_spec.rb
```diff
@@ -179,4 +179,95 @@
...
   end
+
+  context 'when a new work version and nothing has changed' do
+    let(:user) { build(:user) }
+    let(:collection) { build(:collection, required_license: 'CC0-1.0', release_option: 'immediate', access: 'world') }
+    let(:work) { Work.new(collection: collection, depositor: user, owner: user) }
+    let(:work_version) { WorkVersion.new(work: work) }
+
+    context 'when nothing has changed' do
+      # let(:params) do
```
jcoyne, line 190: Can this be removed?

jcoyne (changes requested):

spec/services/work_version_event_description_builder_spec.rb
justinlittman: Yes. Fixed.

jcoyne (approved):

PR #2640, opened 2022-09-09
Use search type input where appropriate

jcoyne:

This is the correct semantic markup


ndushay (approved):
approving.

Curious:  Is there any difference in behavior for "search_field" vs "text_field" or is it an accessibility thing or ...

jcoyne:
@ndushay mostly it's an accessibility thing, but certain plugins like "1Password" won't offer to fill in search fields, so it's an aesthetic benefit too.

PR #2645, opened 2022-09-09
Add duplicate keywords to the javascript validation

### LINKED ISSUE #2604: Deposit error for duplicate keywords should take user to keyword entry part of form with error message
When trying to deposit an item with duplicate keywords, the error message should be displaying inline in the form next to the keyword entry box and the user should be taken to that part of the page.

Currently if I open a draft item that was saved with duplicate keywords and try to deposit, I'm taking to the top of the form -- not the keyword section -- where I see 

And there is no error message next to the keywords:

This is what I see when I enter another duplicate keyword and what it should look like when I try to deposit with a duplicate already entered:

jcoyne:

This will prevent the user from submitting a Deposit that has duplicate keywords

Fixes #2604 

Manually

### DIFF
--- a/app/javascript/controllers/autocomplete_edit_controller.js
+++ b/app/javascript/controllers/autocomplete_edit_controller.js
@@ -9,7 +9,6 @@ export default class extends Controller {
 
   change() {
     if(this.valueTarget.value) {
-      this.inputTarget.readOnly = true
       const [uri, cocina_type] = this.valueTarget.value.split('::')
       this.uriTarget.value = uri
       this.typeTarget.value = cocina_type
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/autocomplete_edit_controller.js
```diff
@@ -9,7 +9,6 @@ export default class extends Controller {
 
   change() {
     if(this.valueTarget.value) {
-      this.inputTarget.readOnly = true
```
jcoyne, line 12: Fields that are read only will not validate (see [willValidate](https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/willValidate)).

PR #2649, opened 2022-09-09
New works and collections should say "Created" in the description of changes

peetucket:

Currently when you create a new work, the description of changes in the "Updated by user" history event lists a bunch of fields.  When you create a new collection, the description of changes in the "Updated by user" history event is blank.

This makes both say "Created", which seems to make more sense for a brand new work or collection.

Note: a new version of an existing work will continue to list fields that were changed, as will a new version of an existing collection.  This change only applies to new works and collections.

~~Verify this change with PO.~~ Verified

Localhost and updated tests.

### DIFF
--- a/app/controllers/first_draft_collections_controller.rb
+++ b/app/controllers/first_draft_collections_controller.rb
@@ -25,9 +25,12 @@ class FirstDraftCollectionsController < ObjectsController
     @form = collection_form(collection_version)
 
     if @form.validate(create_params) && @form.save
-      collection_version.collection.event_context = { user: current_user }
+      collection_version.collection.event_context = { user: current_user, description: 'Created' }
       collection_version.update_metadata!
-      collection_version.begin_deposit! if deposit_button_pushed?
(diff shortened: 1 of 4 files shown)


app/controllers/first_draft_collections_controller.rb
```diff
@@ -25,9 +25,12 @@ def create
...
 
     if @form.validate(create_params) && @form.save
-      collection_version.collection.event_context = { user: current_user }
+      collection_version.collection.event_context = { user: current_user, description: 'Created' }
       collection_version.update_metadata!
-      collection_version.begin_deposit! if deposit_button_pushed?
+      if deposit_button_pushed?
+        collection_version.collection.event_context = { user: current_user }
+        collection_version.begin_deposit!
+      end
```
peetucket, line 33: this ensures we don't also say "Created" for the "Deposited" event description that this will generate

jcoyne (approved):

PR #2665, opened 2022-09-12
[HOLD] remove date last deposited filter in admin work report

### LINKED ISSUE #2661: Remove "date last deposited range" from filter options for item reports
Remove this option completely from the page. 

The reason for this removal is that we are unable to fill in this field for the majority of the items currently in H2, so using this filter producing results that the user might not expect. The data we do have will still be present in the report and can be reviewed there.

peetucket:

Fixes #2661 - remove last deposited range filters for admin work reports (field still present in downloaded report when available).

Localhost and CI


jcoyne (changes requested):
Please hold this. I'm conferring with Amy about the need for this.

peetucket:
Not needed anymore due to data fixes.

PR #2669, opened 2022-09-13
Create event context before after_save called

### LINKED ISSUE #2617: Collection history table lists "contact email modified" when it shouldn't
See #2303 -- this ticket is to capture one piece of that ticket that isn't working properly, which is this:

When a related link is added, the descriptive text says both "related links modified" and "contact email modified". The comment about the contact email should be removed.

lwrubel:
Resolves #2617. 

Added unit test and ran integration test on QA. 

### DIFF
--- a/app/controllers/collection_versions_controller.rb
+++ b/app/controllers/collection_versions_controller.rb
@@ -37,9 +37,14 @@ class CollectionVersionsController < ObjectsController
     @form.prepopulate!
   end
 
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
   def update
     collection_version = CollectionVersion.find(params[:id])
-    clean_params = collection_params
+    orig_collection_version = collection_version
+    orig_clean_params = collection_params
+    clean_params = orig_clean_params.deep_dup
+
     if collection_version.deposited?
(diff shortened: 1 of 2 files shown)


app/controllers/collection_versions_controller.rb
```diff
@@ -37,23 +37,31 @@ def edit
     @form.prepopulate!
   end
 
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/AbcSize
   def update
     collection_version = CollectionVersion.find(params[:id])
-    clean_params = collection_params
+    orig_collection_version = collection_version
```
jcoyne, line 44: Why is this assignment necessary? Both variables point to the sma object?

jcoyne (changes requested):

app/controllers/collection_versions_controller.rb
lwrubel: `collection_version` eventually gets re-assigned and we need to send the initial version to `context_form` on line 55. Maybe there is a better way to do this?

jcoyne (changes requested):

jcoyne (approved):

PR #2678, opened 2022-09-13
remove attached files when globus option is selected

### LINKED ISSUE #2662: Unable to remove a file uploaded while item was in draft state so I can select the Globus option
Once an item has been saved as a draft with a file, it is not possible to choose the Globus option and make a deposit. This means the user is not able to change their mind about which option to use in this scenario.

If you attempt to open the saved draft and change to Globus and deposit, you are prompted that you must remove the file:

The user then has to select the file upload option again in order to remove the file, and then select the Globus option again. This is an awkward series of steps. At which point it looks like this:

Note this error says I need both attach a file and remove all files.

And when the user tries to deposit from this state, they see this error message again and are caught in an endless loop:

peetucket:

Fixes #2662 (see comments for discussion on the ticket that led to this outcome): if the user selects the "globus" option and saves the form  (draft or deposit), remove any attached files instead of throwing a validation error.

Localhost and added new requests specs

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -157,6 +157,8 @@ class WorksController < ObjectsController
   def after_save(form:, event_context:)
     work_version = form.model[:work_version]
     work = form.model[:work]
+    # if the user selects globus uploads, we cannot have any attached files
+    work_version.attached_files.destroy_all if work_version.globus
     work.event_context = event_context
     work_version.update_metadata!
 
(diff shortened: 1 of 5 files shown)


app/controllers/works_controller.rb
```diff
@@ -157,6 +157,8 @@ def work_form(work_version)
   def after_save(form:, event_context:)
     work_version = form.model[:work_version]
     work = form.model[:work]
+    # if the user selects globus uploads, we cannot have any attached files
+    work_version.attached_files.destroy_all if work_version.globus
```
jcoyne, line 161: Can this be done in the form itself? Possibly the attached_files populator?

jcoyne (changes requested):

jcoyne (approved):

PR #2680, opened 2022-09-13
show/hide file uploads panel depending on globus radio selection

### LINKED ISSUE #2627: Revise Globus interaction
Current Globus interaction is confusing because 1) the two radio buttons are really far apart, making it hard to see that there are two options to choose from and 2) when the Globus option is selected after a file is uploaded, the file is still there and may make users think the file is going to be saved, which it isn't.

We want instead 
- for the file upload section to collapse when the Globus option is selected
- to update the formatting and language of the text for the Globus option (see mockups for new text)
- to bold the header text for the two options.

By default, the file upload option should be selected and the user should see this:

Then if the user selects the Globus option, the file upload section collapses so that the user sees only this:

peetucket:

Fixes #2627 - show or hide the uploaded file panel depending on which radio button (globus or uploads) the user has chosen.  It now matches the design shown in the ticket.

Localhost

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -1,47 +1,49 @@
-<section data-controller="complex-radio" id="file">
+<section data-controller="complex-radio file-uploads" id="file">
   <header>Add your files *</header>
 
   <div class="form-check" data-complex-radio-target="selection">
     <%= form.radio_button :globus, false, class: 'form-check-input',
-      data: { action: 'complex-radio#disableUnselectedInputs' }
+      data: { action: 'complex-radio#disableUnselectedInputs file-uploads#updatePanelVisibility' }
     %>
-    <%= form.label :globus, 'Upload files', class: 'form-check-label' %>
-    <p>All file types are accepted. If you have a deposit that is over 10GB, or
-      have any trouble with adding your files,
-      <%= link_to('contact us.', '#contactUsModal', data: {
-                                bs_toggle: 'modal',
-                                bs_target: '#contactUsModal' }) %>
-    </p>
-      <div data-controller="dropzone" data-dropzone-max-file-size="10000" data-dropzone-max-files="1000">
-      <fieldset>
-        <div class="dropzone dropzone-default" data-dropzone-target="container">
-          <%= form.file_field :files, multiple: true, direct_upload: true,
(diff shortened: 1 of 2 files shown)


app/components/works/add_files_component.html.erb
```diff
@@ -1,47 +1,49 @@
...
-      data: { action: 'complex-radio#disableUnselectedInputs' }
+    <%= form.radio_button :globus, true, class: 'form-check-input', 'data-file-uploads-target': 'globusRadioButton',
+      data: { action: 'complex-radio#disableUnselectedInputs file-uploads#updatePanelVisibility' }
     %>
-    <%= form.label :globus, 'I have uploaded my files to a Globus endpoint provided to me by SDR staff.', class: 'form-check-label' %>
-    <p>Do not check this box unless you have already discussed this with SDR staff.</p>
+    <%= form.label :globus, 'I have uploaded my files to a Globus endpoint provided to me by SDR staff.', class: 'form-check-label', style: 'font-weight: bold' %>
+    <p><em>Only select this option if you have already discussed Globus with SDR staff.</em></p>
   </div>
 </section>
```
peetucket, line 49: there are fewer changes than this diff suggests...basically added some stimulus targets and actions to the radio button, rewording text and bolding labels, and then wrapping the main file uploads panel in a new div so it can be hidden or shown as needed (which caused spacing  issues, making it appears there are number of other changes).

app/components/works/add_files_component.html.erb
```diff
@@ -1,47 +1,49 @@
...
-    </p>
-      <div data-controller="dropzone" data-dropzone-max-file-size="10000" data-dropzone-max-files="1000">
-      <fieldset>
-        <div class="dropzone dropzone-default" data-dropzone-target="container">
-          <%= form.file_field :files, multiple: true, direct_upload: true,
-                              data: { dropzone_target: 'input' } %>
-          <div class="dz-message needsclick text-secondary">
-            <p>Drop files here</p>
+    <%= form.label :globus, 'Upload files', class: 'form-check-label', style: 'font-weight: bold' %>
+    <div id="uploaded-files-panel" data-file-uploads-target="fileUploads">
```
peetucket, line 9: the new `<div>` i added is here

jcoyne (approved):

Issue #2626, opened 2022-09-08
Experimental docker based deploy


jcoyne:
Check to see that we can deploy in a docker based environment with SAML working

justinlittman:
This is the request to provision a server: https://github.com/sul-dlss/operations-tasks/issues/3148

amyehodge:
@justinlittman  Tested deployment at https://h2-docker-qa.stanford.edu/ on Oct. 3.

Good news:

- web site loaded
- I was able to log in
- I was able to fill in and submit the form to create a new collection
- when I created a new collection, I received the correct email notification about being added as a manager
- the help link created a JIRA ticket
- I appear to have been able to sign up for the newsletter from the home page
- save and print functions for the Terms of Deposit appear to work like they do in production (though the save option seems borked)
- I see the admin features and am able to use them, but it’s hard to tell what’s working properly without any data in this system. When I can create some data I’ll poke at this further.

Less good news:

- My collection deposit appears to be stuck in the “depositing….” state. Creation of the collection appears to be the source of this HB error: https://app.honeybadger.io/projects/77112/faults/88287751
- the email notification was addressed to “Dear (null),”
- since I am unable to get the collection to complete depositing and there is no other data in the system, I’m currently unable to test anything else.

justinlittman:
Not proceeding with this work.

PR #2688, opened 2022-09-14
Adds docker QA deploy.

justinlittman:
refs #2626

Do you have an hour to discuss?


docker/app/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # This Dockerfile is optimized for running in development. That means it trades
```
jcoyne, line 1: This is no longer true.

justinlittman:
TODO: Consider how to re-integrate with local development Dockerfile since not using `invoke.sh`.

PR #2718, opened 2022-09-16
Add doi to auto-generated citation

### LINKED ISSUE #2628: Add DOI to autogenerated citation
We should already know what the DOI is as soon as we have a druid.

Like this: https://github.com/sul-dlss/dor-services-app/blob/2f3817c9a07d918f923b8a1a74e04b2c54759763/app/services/create_object_service.rb#L115-L116

lwrubel:
Resolves #2628. Adds DOI or placeholder to the auto-generated citation. Goes with https://github.com/sul-dlss/dor-services-app/pull/4234. 

Unit, integration tests on QA. 

### DIFF
--- a/app/components/works/detail_component.html.erb
+++ b/app/components/works/detail_component.html.erb
@@ -150,7 +150,7 @@
     </tr>
     <tr>
       <th scope="row">Preferred citation</th>
-      <td data-work-updates-target="citation">
+      <td>
         <%= citation %>
       </td>
     </tr>
--- a/app/components/works/form_component.rb
+++ b/app/components/works/form_component.rb
@@ -21,6 +21,9 @@ module Works
 
     delegate :persisted?, to: :work_form
     delegate :purl, to: :work
+    delegate :druid, to: :work
+    delegate :doi, to: :work
+    delegate :will_assign_doi?, to: :work_form
 
     def work
       work_form.model.fetch(:work)
@@ -28,6 +31,17 @@ module Works
 
     def work_version
       work_form.model.fetch(:work_version)
+    end
+
+    def doi_field
+      return "https://doi.org/#{doi}." if doi
+
+      # create DOI URL or placeholder for works in collections that allow DOI assignment
+      return unless will_assign_doi?
+
+      return "https://doi.org/#{Doi.for(druid)}." if druid
+
+      WorkVersion::DOI_TEXT
     end
   end
 end
--- a/app/services/cocina_generator/description/generator.rb
+++ b/app/services/cocina_generator/description/generator.rb
@@ -72,9 +72,10 @@ module CocinaGenerator
       def citation
         return if work_version.citation.blank?
 
-        # :link: is a special placeholder in dor-services-app.
+        # :link: and :doi: are special placeholders in dor-services-app.
         # See https://github.com/sul-dlss/dor-services-app/pull/1566/files#diff-30396654f0ad00ad1daa7292fd8327759d7ff7f3b92f98f40a2e25b6839807e2R13
-        exportable_citation = work_version.citation.gsub(WorkVersion::LINK_TEXT, ':link:')
+        exportable_citation = work_version.citation.gsub(WorkVersion::LINK_TEXT, ':link:').gsub(WorkVersion::DOI_TEXT,
+                                                                                                ':doi:')
 
         Cocina::Models::DescriptiveValue.new(
           value: exportable_citation,
--- a/spec/features/create_new_draft_work_spec.rb
+++ b/spec/features/create_new_draft_work_spec.rb
@@ -31,5 +31,32 @@ RSpec.describe 'Create a new work in a deposited collection', js: true do
 
     expect(page).to have_content 'My Draft'
     expect(page).to have_content 'Draft - Not deposited'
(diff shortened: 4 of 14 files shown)


app/components/works/detail_component.html.erb
```diff
@@ -150,7 +150,7 @@
     </tr>
     <tr>
       <th scope="row">Preferred citation</th>
-      <td data-work-updates-target="citation">
+      <td>
```
lwrubel, line 153: Removing vestige of previously deleted stimulus controller.

lwrubel:
@jcoyne Should DOIs be added into the exported citation? I started down that path and got halfway before wondering if we should then be add [something like was done for PURL links](https://github.com/sul-dlss/dor-services-app/pull/1566/files#diff-30396654f0ad00ad1daa7292fd8327759d7ff7f3b92f98f40a2e25b6839807e2R13) to DSA for DOIs as well?

jcoyne:
@lwrubel yes, we're trying to do just what we do for the purl links

app/components/works/form_component.rb
```diff
@@ -29,5 +31,17 @@ def work
     def work_version
       work_form.model.fetch(:work_version)
     end
+
+    def doi_field
+      return "https://doi.org/#{Doi.for(druid)}." if doi
```
jcoyne, line 36: Shouldn't we just return the DOI if it is present?  Some DOIs do not match the druid patter as they were issued outside of H2.

app/components/works/form_component.rb
```diff
@@ -29,5 +31,17 @@ def work
...
+
+      # create DOI URL or placeholder for works in collections that allow DOI assignment
+      collection = work.collection
+      return unless (collection.doi_option == 'depositor-selects' && work.assign_doi) || collection.doi_option == 'yes'
+
+      return "https://doi.org/#{Doi.for(druid)}." if druid
+
+      ':DOI will be inserted here automatically when available:'
```
jcoyne, line 44: Should this use `WorkVersion::DOI_TEXT`?

app/components/works/form_component.rb
```diff
@@ -29,5 +31,17 @@ def work
     def work_version
       work_form.model.fetch(:work_version)
     end
+
+    def doi_field
+      return "https://doi.org/#{Doi.for(druid)}." if doi
+
+      # create DOI URL or placeholder for works in collections that allow DOI assignment
+      collection = work.collection
+      return unless (collection.doi_option == 'depositor-selects' && work.assign_doi) || collection.doi_option == 'yes'
```
jcoyne, line 40: It looks like this duplicates some logic in `work.assign_doi?` Can you extract that into a separate method (`will_assign_doi?` perhaps) that we can use here?

app/services/cocina_generator/description/generator.rb
```diff
@@ -72,9 +72,10 @@ def abstract
       def citation
         return if work_version.citation.blank?
 
-        # :link: is a special placeholder in dor-services-app.
+        # :link: and :doi: are special placeholders in dor-services-app.
         # See https://github.com/sul-dlss/dor-services-app/pull/1566/files#diff-30396654f0ad00ad1daa7292fd8327759d7ff7f3b92f98f40a2e25b6839807e2R13
         exportable_citation = work_version.citation.gsub(WorkVersion::LINK_TEXT, ':link:')
```
jcoyne, line 77: These can be chained into one line. (e.g. `.gsub(...).gsub(...)`)

jcoyne (changes requested):

spec/features/create_new_draft_work_spec.rb
```diff
@@ -31,5 +31,32 @@
...
+
+      click_button 'Continue'
+
+      fill_in 'Title of deposit', with: 'My Draft'
+
+      click_button 'Save as draft'
+
+      expect(page).to have_content 'My Draft'
+      expect(page).to have_content 'Draft - Not deposited'
+      expect(page).to have_content WorkVersion::LINK_TEXT.to_s
```
jcoyne, line 58: Is `.to_s` necessary on these  two lines?

jcoyne (changes requested):

app/services/cocina_generator/description/generator.rb
lwrubel: Done.

app/components/works/form_component.rb
lwrubel: I moved this logic into a new method as you suggested and put it in `draft_work_form.rb`. Does that make sense?

app/components/works/form_component.rb
lwrubel: Right, that's better. Fixed.

app/components/works/form_component.rb
lwrubel: Oops hadn't been thinking about other DOIs. Fixed.

spec/features/create_new_draft_work_spec.rb
lwrubel: Looks like it isn't (must have been leftover from some previous version of the test I was working on). Fixed.

lwrubel:
@jcoyne this is ready to be reviewed again.

jcoyne (approved):

PR #2725, opened 2022-09-19
Require subtypes for music and mixed materials.

### LINKED ISSUE #2687: Provide clearer error message to user for mixed materials
When a user fails to choose two subtypes in the work type modal as required for the Mixed Materials work type, and then clicks "Continue," they are taken back to the dashboard where an error message displays at the top of the page.

Ideally, the error message would display on the modal.

Wherever it displays, it should say "You must select two subtype options for work type "Mixed Materials."

Here's what it currently says if you only choose one option (dependent on which option you chose, of course):

And here's what it says if you don't select any subtypes:

justinlittman:
closes #2687

Per user feedback for errors.

Unit, cypress

### DIFF
--- a/app/components/works/work_type_modal_component.html.erb
+++ b/app/components/works/work_type_modal_component.html.erb
@@ -65,7 +65,10 @@
 
           <div class="mb-4 subtype-container" data-work-type-target="moreTypes" hidden></div>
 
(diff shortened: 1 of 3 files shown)


mjgiarlo:
@justinlittman Approved. Not sure if @amyehodge wanted to test before merging. If not, merge away!

justinlittman:
She didn't mark it as requiring review, so I'm going to merge.

PR #2732, opened 2022-09-19
Allows selecting all/none email preferences.

### LINKED ISSUE #2707: Add "select all" checkbox to email preferences
See designs on #2510

justinlittman:
closes #2707

cypress

Note that deviation from mock-ups was @amyehodge approved.

### DIFF
--- a/app/javascript/controllers/checkboxes_controller.js
+++ b/app/javascript/controllers/checkboxes_controller.js
@@ -1,0 +1,19 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = ["checkbox"]
+
+  checkAll(event) {
+    event.preventDefault()
+    this.checkboxTargets.forEach((checkbox) => {
+      checkbox.checked = true
+    })
+  }
+
+  checkNone(event) {
+    event.preventDefault()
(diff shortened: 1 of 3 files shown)


justinlittman:
image

app/javascript/controllers/checkboxes_controller.js
```diff
@@ -0,0 +1,23 @@
+import { Controller } from "stimulus"
+
+export default class extends Controller {
+  static targets = ["checkbox"]
+
+  connect() {
+    console.log("Connected")
```
jcoyne, line 7: Cruft.

jcoyne (changes requested):
This doesn't seem to follow the UI as sketched out by @astridu.  Is there a reason to deviate from those designs?

amyehodge:
@jcoyne @justinlittman and I talked about how there isn't a clear path for the user here when only some of the boxes are checked. Should we be displaying the "select all" or the "deselect all" option? This allows the user to use either of those options at any time and seemed a good solution.

app/javascript/controllers/checkboxes_controller.js
justinlittman: Removed.

jcoyne (approved):

PR #2738, opened 2022-09-20
add some additional checks to work modal window

### LINKED ISSUE #2624: Write cypress test for type of content modal
image

This is for both the overall behavior of the modal and the dynamic behavior when particular content types are selected.

peetucket:

Fixes #2624 - Add some additional cypress tests for the work modal

### DIFF
--- a/cypress/spec/work_type_modal.cy.js
+++ b/cypress/spec/work_type_modal.cy.js
@@ -15,7 +15,16 @@ describe('Work type modal', () => {
       cy.wait('@editLink')
       cy.get(`button[data-destination="/collections/${collection_id}/works/new"]`).click()
       // This wait for the async turbo call to finish.
-      cy.wait('@editLink')  
+      cy.wait('@editLink')
+    })
+  })
+
+  // clicking submit without select a work type leaves user on the work type options modal and not the new work form
+  it('requires a work type', () => {
+    cy.get('#workTypeModal').within(() => {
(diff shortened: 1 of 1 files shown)


ndushay (changes requested):
I see that what you added is perfectly analogous, but I don't see how this description:

```it('requires 0 subtypes for text type'``

is satisfied by 

```
cy.url().should('include', `/collections/${collection_id}/works/new?work_type=text`)
```

Is it a poor description, or a poor test, or is the test missing a comment explaining why?

justinlittman (approved):

peetucket:
> I see that what you added is perfectly analogous, but I don't see how this description:
> 
> ```it('requires 0 subtypes for text type'``
> 
> is satisfied by
> 
> ```
> cy.url().should('include', `/collections/${collection_id}/works/new?work_type=text`)
> ```
> 
> Is it a poor description, or a poor test, or is the test missing a comment explaining why?

I'll add a code comment.  Basically the explanation is that you cannot proceed to that URL (which is the form where you actually create the new work) unless you satisfy the requirements of selecting the type.  This test shows that do not need to select any sub types in order to proceed.  A previous test (for a different work type, e.g. mixed materials) shows that when you click submit you will instread see a validation message and does not proceed to the form) until you select subtypes.

peetucket:
Added a code comment and one more test verifying that you must select a work type to proceed.

ndushay (approved):

PR #2741, opened 2022-09-21
Adds admin function for moving a collection.

### LINKED ISSUE #2685: Add option to move an item to a new collection
Admins should be able to move an item to a different collection in H2. This feature should appear in the admin function drop-down menu on the item page.

* Items can only be moved to collections that have been deposited (so, not anything in a "first draft" state, but later draft states are fine)
* Items can't be moved if certain conflicts exist:
-- the item is embargoed but the collection is set for immediate release only
-- the depositor of the item chose not to get a DOI but the collection requires DOI assignment
-- the item has a license that is not allowed by the collection setting
-- the item is set for Stanford visibility but the collection requires world visibility
* this move needs to trigger a versioning of the item so that the new collection information is sent to SDR

Item view page with "move to another collection" selected.

The drop-down menu would exclude collections that are in first draft state. Collection titles should be prefaced with their druid in this list.

Once a collection has been selected, the user will click the "Move" button.

At this point, the check for conflicts would run and if there are none, the move would happen and the item would be redeposited and a success message similar to the others for the admin actions on items would be displayed.

If conflicts are found, the user would see a message that would include whichever errors are appropriate for this item:

justinlittman:
closes #2685

Unit, Amy

### DIFF
--- a/app/controllers/work_move_controller.rb
+++ b/app/controllers/work_move_controller.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+# a controller for changing the collection of work.
+class WorkMoveController < ApplicationController
+  before_action :authenticate_user!
+  verify_authorized
+
+  def search
+    authorize! :work_move
+
+    work = Work.find(params[:id])
+    collections = Collection.where(druid: "druid:#{params[:druid]}")
+
+    render json: collections.map { |collection| collection_json(collection, work) }
+  end
+
+  def edit
+    work = Work.find(params[:id])
+    authorize! work, to: :move_collection?
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def update
+    work = Work.find(params[:id])
+    authorize! work, to: :move_collection?
+
+    collection = Collection.find(params[:collection])
+
+    if CheckMoveWorkService.check(work:, collection:).any?
+      flash[:error] = I18n.t('work.flash.work_not_moved')
+    else
+      work.update!(collection:)
+      flash[:success] = "Moved #{work.head.title} to #{collection.head.name}"
+    end
+
+    redirect_to work_path(work), status: :see_other
+  end
+  # rubocop:enable Metrics/AbcSize
+
+  private
+
+  def collection_json(collection, work)
+    {
+      id: collection.id,
+      name: collection.head.name,
+      druid: collection.druid,
+      errors: CheckMoveWorkService.check(work:, collection:)
+    }
+  end
+end
--- a/app/models/work.rb
+++ b/app/models/work.rb
@@ -53,6 +53,10 @@ class Work < ApplicationRecord
     head&.deposited? && head.embargo_date.present? && head.embargo_date <= Time.zone.today
   end
 
+  def embargoed?
+    head.deposited? && head.embargo_date.present? && head.embargo_date > Time.zone.today
+  end
+
   delegate :name, to: :depositor, prefix: true
   delegate :name, to: :owner, prefix: true
   delegate :purl_reservation?, to: :head
--- a/app/policies/work_policy.rb
+++ b/app/policies/work_policy.rb
@@ -25,4 +25,8 @@ class WorkPolicy < ApplicationPolicy
   def owner_of_the_work?
     record.owner == user
   end
+
+  def move_collection?
+    administrator?
+  end
 end
--- a/cypress/app_commands/clean.rb
+++ b/cypress/app_commands/clean.rb
@@ -1,13 +1,9 @@
 # frozen_string_literal: true
 
-if defined?(DatabaseCleaner)
-  # cleaning the database using database_cleaner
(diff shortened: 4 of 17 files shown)


justinlittman:
@amyehodge approved.

app/controllers/work_move_controller.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# a controller for changing the collection of work.
+class WorkMoveController < ApplicationController
+  before_action :authenticate_user!
+  verify_authorized except: %i[search]
```
jcoyne, line 6: Can you add a comment here about this decision. "We don't authorize on the search action because ..."

app/controllers/work_move_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    render json: collections.map { |collection| collection_json(collection, work) }
+  end
+
+  def edit
+    authorize! :work_owner
+  end
+
+  def update
+    authorize! :work_owner
```
jcoyne, line 22: I believe this should be `authorize! work, to: :move_collection?`  you could possibly add:
`authorize! work, to: :move_collection?, context: {collection: collection}`

app/controllers/work_move_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    work = Work.find(params[:id])
+    collection = Collection.find(params[:collection])
+
+    work.update!(collection:)
```
jcoyne, line 27: Should we re-run `CheckMoveWorkService` here? Is would be possible something changes between when they search for the item and when they update it, right?

app/controllers/work_move_controller.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+# a controller for changing the collection of work.
+class WorkMoveController < ApplicationController
+  before_action :authenticate_user!
+  verify_authorized except: %i[search]
+
+  def search
+    druid = params[:druid]
+    druid = "druid:#{druid}" unless druid.start_with?('druid:')
```
jcoyne, line 10: Doesn't the javascript controller always strip off the `druid:` prefix?  Why do we need this?

app/models/work.rb
```diff
@@ -53,6 +53,10 @@ def already_embargo_released?
     head&.deposited? && head.embargo_date.present? && head.embargo_date <= Time.zone.today
   end
 
+  def embargoed?
+    head&.deposited? && head.embargo_date.present? && head.embargo_date > Time.zone.today
```
jcoyne, line 57: Why do we need the null test on head here? There should be no way to have created a record in that state.

cypress/app_commands/clean.rb
```diff
@@ -6,7 +6,7 @@
   DatabaseCleaner.clean
 else
   # Manual cleaning
-  true
+  Collection.update_all(druid: nil)
```
jcoyne, line 9: Can we remove the clause above? We don't have DatabaseCleaner. I'm also curious why you are only updating the druid and not doing `Collection.destroy_all`?

jcoyne (changes requested):

app/controllers/work_move_controller.rb
justinlittman: I just removed the exception. There was no good reason for it.

app/controllers/work_move_controller.rb
justinlittman: Added.

app/controllers/work_move_controller.rb
justinlittman: Removed.

app/models/work.rb
justinlittman: I don't know, but the existing `already_immediately_released?` and `already_embargo_released?` have the null test. I was just copying.

cypress/app_commands/clean.rb
justinlittman: `Collection.destroy_all` requires the collection head be set to nil first. This is the least bit of clean-up to allow collections to be created with identical druids. I've added a comment to better explain.

app/policies/work_policy.rb
```diff
@@ -25,4 +25,10 @@ def destroy?
   def owner_of_the_work?
     record.owner == user
   end
+
+  def move_collection?
+    # debugger
```
jcoyne, line 30: Cruft?

app/models/work.rb
jcoyne, line 57: Please remove this unnecessary nil check

jcoyne (changes requested):

app/models/work.rb
jcoyne: Yes, I believe this is required for those because they are used on the works/new form.  If this is not used in that scenario, it isn't required.

jcoyne (approved):

PR #2743, opened 2022-09-21
[HOLD] remove unused component; properly check for duped keywords for validation

### LINKED ISSUE #2734: Problem depositing after item is saved as draft with duplicate keywords
See the video below. 

Before the video starts I had opened a version of this item and saved it as a draft with duplicate keywords, both chosen from the FAST vocabulary options.

Then in the video you'll see me do this:
1. Open the item to edit
2. Delete the second keyword -- note the other keyword still shows as red/invalid
3. Click the deposit button. I'm taken back up to the keywords. 
4. Click inside the existing keyword box and then click outside the box.
5. Click the deposit button. Now it moves me a tiny bit up the page. Every time I click deposit it moves up the page a tiny bit.

https://images.zenhubusercontent.com/272543337/b4d992bb-5588-4954-8389-1881e51b1d2c/screen_recording_2022_09_19_at_4_49_11_pm.mov

If I scrolled back up you would see there aren't any errors showing on the page any more. From here I was able to save it as a draft. When I edited it again, the second keyword was still gone and I was able to click deposit and it worked.

Ideally, once the second keyword is deleted the validations would understand that there is no longer a problem and I should be able to deposit the item.

peetucket:

See #2755 as an alternative - which does de-duplication on the server instead of validation
----------------

Fixes #2734 - were were not properly clearing the invalid state when users were removing duplicate keywords, and thus the entire form was incorrectly being marked as invalid and thus unable to submit.

Fixed by ensuring we call the JS method responsible for validating keywords and marking the form as valid when there are no dupes.

Also removed a related keyword component that I don't think is used anywhere (I cannot find any references to it in the code).

HOLD For testing by PO on a server

Note: build is passing, just not updated in github due to an issue with circeci on 9/21: https://app.circleci.com/pipelines/github/sul-dlss/happy-heron/4987/workflows/bde3bd03-1cd9-40d7-a858-e5914551f3bd

Localhost browser and CI

Maybe useful to have a cypress test for this since it's all javascript client side validation at this level, the server side validation is already tested elsewhere)?


app/javascript/controllers/keywords_controller.js
```diff
@@ -18,36 +18,17 @@ export default class extends Controller {
...
-      .replace(/TEMPLATE_LABEL/g, event.target.value)
-
-    this.searchTarget.value = ''
-    this.addItemTarget.insertAdjacentHTML('beforeend', content)
-    // Remove error indications when keyword(s) added
-    this.containerTarget.classList.remove('is-invalid')
-    this.checkForDuplicates()
-  }
-
   checkForDuplicates() {
```
peetucket, line 21: this methods isn't called anywhere from any components in use as far as I can tell, instead we will just call `checkForDuplicates` directly from the existing component in use

app/javascript/controllers/keywords_controller.js
```diff
@@ -18,36 +18,17 @@ export default class extends Controller {
...
   }
 
-  removeAssociation(event) {
-    const item = event.target.closest(".selection-choice")
-    item.querySelector("input[name*='_destroy']").value = 1
-    item.style.display = 'none'
-    this.checkForDuplicates() // clear out the invalid selection class if there are no duplicates anymore
-  }
-
   hasDuplicateKeywords() {
```
peetucket, line 32: this methods isn't called anywhere from any components in use as far as I can tell, instead we will just call `checkForDuplicates` directly from the existing component in use

peetucket:
Closed in favor of #2755

PR #2752, opened 2022-09-26
Update dependencies

sul-devops-team:


edsu:
We are seeing this build error in CI and I see it in my dev environment. But @lwrubel and @ndushay don't see it. So in order not to hold up dep updates further I'm going to manually merge.

PR #2754, opened 2022-09-26
try sleeping for 1 for failing test

ndushay:

dependency updates, for some reason, caused a test to fail.


ndushay:
FAIL.

edsu:
Thanks for trying it!

PR #2755, opened 2022-09-26
De-dupe keywords server side

### LINKED ISSUE #2734: Problem depositing after item is saved as draft with duplicate keywords

peetucket:

Fixes #2734 - Alternative to #2743 after discussion in standup with Amy.  Instead of trying to do complex validation client and servers side, we will instead do auto de-duplication keywords on the server side (preferring keywords with URIs that come from auto-complete selection) and then remove keyword validation (server/client side).

Also removes an unused keywords component and it's related test.

Localhost browser, and updated tests.


app/forms/work_form.rb
```diff
@@ -6,7 +6,6 @@
 class WorkForm < DraftWorkForm
   validates :abstract, :access, :title, presence: true, allow_nil: false
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
-  validate :keywords_unique
```
peetucket, line 9: remove server side keyword validation

app/javascript/controllers/keywords_controller.js
```diff
@@ -17,42 +11,4 @@ export default class extends Controller {
...
-    this.checkForDuplicates() // clear out the invalid selection class if there are no duplicates anymore
-  }
-
-  hasDuplicateKeywords() {
-    // determine if the user has entered duplicate keywords
-    const nodes = document.querySelectorAll('.keyword-row:not([style*="display: none"]) input[type="text"]')
-    const keywords = Array.from(nodes, target => target.value)
-    // a set cannot have duplicates but an array can, so if the set and array are different in size, we have dupes
-    return (new Set(keywords)).size !== keywords.length
-  }
```
peetucket, line 57: remove all of the client side javascript that was specific to keyword validation (there are generic methods for adding/removing associations that are used to add/remove keywords)

app/forms/draft_work_form.rb
```diff
@@ -175,4 +176,18 @@ def model_name
   def will_assign_doi?
     (collection.doi_option == 'depositor-selects' && assign_doi) || collection.doi_option == 'yes'
   end
+
+  # de-dupe keywords, preferring to throwing away the free text entry duplicate(s) without a URI
+  def dedupe_keywords
+    return if work_version.keywords.size.zero?
+
+    grouped_keywords = Keyword.where(work_version:).group_by(&:label) # group keywords by the label
+    grouped_keywords.each do |_label, group|
```
justinlittman, line 185: Slight suggestion: `grouped_keywords.values.each`

justinlittman (approved):

app/forms/draft_work_form.rb
peetucket: sounds good, in fact rubocop suggested `grouped_keywords.each_value` which I didn't know existed, and will use.

PR #2756, opened 2022-09-27
Docker2

justinlittman:
Docker cures all ills.


app/jobs/assign_pid_job.rb
```diff
@@ -6,7 +6,7 @@ class AssignPidJob
   # This worker will connect to "h2.druid_assigned" queue
   # env is set to nil since by default the actual queue name would be
   # "h2.druid_assigned_development"
-  from_queue 'h2.druid_assigned', env: nil
+  from_queue Settings.rabbitmq.queues.druid_assigned, env: nil
```
justinlittman, line 9: All of these queue changes are so that we can run multiple instances of H2 in a single environment and each has its own queue. This will allow us to run a dockerized H2 and a plain-old H2 together.

config/puma.rb
```diff
@@ -6,8 +6,8 @@
 # the maximum value specified for Puma. Default is set to 5 threads for minimum
 # and maximum; this matches the default thread size of Active Record.
 #
-max_threads_count = ENV.fetch('RAILS_MAX_THREADS', 5)
-min_threads_count = ENV.fetch('RAILS_MIN_THREADS') { max_threads_count }
+max_threads_count = ENV.fetch('PUMA_MAX_THREADS', 5)
```
justinlittman, line 9: Changing these variables so that their naming is clearer elsewhere (e.g., in puppet files).

db/seeds.rb
```diff
@@ -9,4 +9,4 @@
 #   Character.create(name: 'Luke', movie: movies.first)
 
 # The SDR user is used in Events performed by SDR.
-User.create!(name: 'SDR', email: 'sdr@stanford.edu')
+User.find_or_create_by!(name: 'SDR', email: 'sdr@stanford.edu')
```
justinlittman, line 12: This allows seeding as part of deploy. Currently, it was performed as a manual operation.

justinlittman (commented):
For an explanation of some of these changes, see https://github.com/sul-dlss/dlss-capistrano-docker/blob/0d7c1c12e14f02cf0a9e27569098377402796eee/README.md

docker-compose.prod.yml
```diff
@@ -0,0 +1,66 @@
...
+      - DATABASE_USERNAME
+      - DATABASE_PASSWORD
+      - DATABASE_HOSTNAME=host.docker.internal
+      # for ActionCable
+      - REDIS_URL=redis://host.docker.internal:6379/
+      - SMTP_HOST=host.docker.internal
+      - HOST
+      - HONEYBADGER_API_KEY
+      - SETTINGS__RABBITMQ__PASSWORD
+      - SETTINGS__RABBITMQ__QUEUES__DEPOSIT_COMPLETE
```
jcoyne, line 23: Do we need these `*_QUEUES_*` variables in here?  Aren't they basically constants in settings.yml?

justinlittman: No, they vary by server so that we can run multiple instances of H2 (e.g., dockerized and non-dockerized) and each has its own queue. This is set in puppet.

PR #2757, opened 2022-09-27
add a cypress test for creating a work with uploaded files

### LINKED ISSUE #2623: Write cypress test for adding files
image

Developer discretion as to what needs to be tested.

peetucket:

Fixes #2623 - add a test for uploading a file in the browser and test that is not required for a globus upload

Localhost CI

### DIFF
--- a/cypress/spec/create_work.cy.js
+++ b/cypress/spec/create_work.cy.js
@@ -1,0 +1,57 @@
+describe('Create work', () => {
+    let work_id
+
+    beforeEach(() => {
+      const results = cy.appFactories([
+        ['create', 'work_version_with_work_and_collection', 'with_required_associations', {} ]
+      ]).then((results) => {
+        cy.log(results[0])
+        work_id = results[0].work_id
+
+        cy.visit(`/works/${work_id}/edit`)
+    })
+    })
+
+    it('deposits a work correctly after uploading a file', () => {
+
+      // try to deposit
+      cy.get('input.btn[value="Deposit"]').click()
+
+      // there is a message telling us we need to upload a file
+      cy.get('div.invalid-feedback').should('contain', 'You must attach a file')
+
+      // now upload a file
+      cy.get('div.dropzone').selectFile('cypress/fixtures/test.txt', {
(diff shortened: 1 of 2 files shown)


cypress/spec/create_work.cy.js
```diff
@@ -0,0 +1,57 @@
+describe('Create work', () => {
+    let work_id
+
+    beforeEach(() => {
+      const results = cy.appFactories([
+        ['create', 'work_version_with_work_and_collection', 'with_required_associations', {} ]
+      ]).then((results) => {
+        cy.log(results[0])
```
justinlittman, line 8: cruft

cypress/spec/create_work.cy.js
```diff
@@ -0,0 +1,57 @@
...
+      // there is a message telling us we need to upload a file
+      cy.get('div.invalid-feedback').should('contain', 'You must attach a file')
+
+      // now upload a file
+      cy.get('div.dropzone').selectFile('cypress/fixtures/test.txt', {
+        action: 'drag-drop'
+      })
+
+      // wait for the upload to finish
+      cy.wait(1000)
```
justinlittman, line 29: Is there something you can wait for to appear rather than just a wait?

justinlittman (approved):
Approve pending consideration of comments.

cypress/spec/create_work.cy.js
peetucket: Actually this is part of the test -- the fixture sets up a draft work with all of the required metadata filled in, so all that is left to do to complete the deposit and make it valid is upload a file.  If we didn't do this, we'd have to tediously fill in a big form.

cypress/spec/create_work.cy.js
peetucket: I tried waiting for the filename and checkmark to appear after uploading, and while those assertions do work, the test will still always fail.  It looks like the click of the "Deposit" button doesn't work unless you pause for just a bit after uploading the file.  I'm not sure why, perhaps some interaction with the file upload component.

PR #2760, opened 2022-09-28
Change modeling of EDTF approximate.

### LINKED ISSUE #2739: model approximate dates with ~ rather than ?
? means uncertain, but we want to use `~` which indicated approximate. See https://www.loc.gov/standards/datetime/

justinlittman:
closes #2739

To align with EDTF spec.

Unit


jcoyne (approved):
This looks good, but I was imagining, we could just clean all the data rather than support both ways. Can you suggest a strategy for that?

PR #2761, opened 2022-09-28
Rename deposit complete job

jcoyne:
DepositCompleteJob seems easier to understand than DepositStatusJob

CI


jmartin-sul (approved):
linter's complaining about the new name for the spec file (`Spec path should end with deposit_complete_job*_spec.rb`), but otherwise LGTM.

jmartin-sul (approved):

PR #2762, opened 2022-09-28
Remove locally cached files after deposit

### LINKED ISSUE #2639: Remove local files after deposit
This change means we are no longer required to hold a copy of every file uploaded to H2. 

How do to this:

Find the file on disk 
```
blob = Work.first.head.attached_files.first.file.blob
disk_service = blob.service
blob.service.path_for(blob.key)
 => "/data/h2-files/xk/ci/xkci6tpbmb33v5n9jp2hxuwuckl8"
```
Then delete that file.

Next update the ActiveStorage::Blob record https://github.com/sul-dlss/happy-heron/blob/main/app/models/attached_file.rb#L19-L20

jcoyne:

Once the files have moved to preservation, we don't need to keep an extra copy around

Fixes #2639

Test suite, tested on stage.

### DIFF
--- a/app/models/attached_file.rb
+++ b/app/models/attached_file.rb
@@ -5,19 +5,23 @@ class AttachedFile < ApplicationRecord
   belongs_to :work_version
   has_one_attached :file
 
-  def blob
-    file.attachment.blob
-  end
+  # def blob
+  #   file.blob
+  # end
 
+  delegate :blob, to: :file
   delegate :filename, :content_type, :byte_size, to: :blob
 
-  # This is a temporary method that makes the blobs appear to be from preservation,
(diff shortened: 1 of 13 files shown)


jcoyne:
Seems to cause this problem: https://app.honeybadger.io/projects/77112/faults/88742974

app/models/attached_file.rb
```diff
@@ -5,19 +5,23 @@ class AttachedFile < ApplicationRecord
   belongs_to :work_version
   has_one_attached :file
 
-  def blob
-    file.attachment.blob
-  end
+  # def blob
+  #   file.blob
+  # end
```
mjgiarlo, line 10: Cruft?

mjgiarlo (approved):

PR #2766, opened 2022-09-28
Replace uncertain with approximate for EDTF dates.

### LINKED ISSUE #2739: model approximate dates with ~ rather than ?

justinlittman:
closes #2739

To align with EDTF spec.

### DIFF
--- a/app/components/works/dates_component.rb
+++ b/app/components/works/dates_component.rb
@@ -56,7 +56,7 @@ module Works
     def created_approximate?
       return false unless created_edtf
 
-      created_edtf.uncertain?
+      created_edtf.approximate?
     end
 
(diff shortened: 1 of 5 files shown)


justinlittman:
To clean existing data:
```
WorkVersion.where("created_edtf like '%?%'").each do |wv|
  wv.update!(created_edtf: Date.edtf(wv.created_edtf.edtf.to_s.gsub('?', '~')))
end
```

jcoyne (approved):
This is much simpler!

mjgiarlo (approved):

PR #2771, opened 2022-09-28
Remove unnecessary conditionals

jcoyne:

No works should be in this state where this needs to be checked.

CI


mjgiarlo (approved):

ndushay (commented):
tests are failing and it's probably the tests that caused these conditionals to appear ...

jcoyne:
The conditional is necessary for rendering the new work form.

PR #2776, opened 2022-09-29
Extract the state machine for WorkVersion to its own file

jcoyne:

The file was getting too long. This makes it easier to find what you are looking for.

CI

### DIFF
--- a/app/models/collection_version.rb
+++ b/app/models/collection_version.rb
@@ -10,67 +10,10 @@ class CollectionVersion < ApplicationRecord
 
   after_update_commit -> { collection.broadcast_update }
 
+  include CollectionVersionStateMachine
+
   def accessioned?
     %w[first_draft depositing].exclude?(state)
-  end
-
-  state_machine initial: :new do
-    before_transition do |collection_version, transition|
-      event_params = collection_version.collection.event_context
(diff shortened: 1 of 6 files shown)


ndushay (commented):
Here, I feel like splitting out the state machine into a class ... that is only used by one class ... is just making me have to look in more files to find something.   If I'm in a minority on this point, so be it;  we can go with the majority.

But maybe it's better to just put the state machine at the bottom of the file.

Another argument against is that we're doing this for only ONE state machine?

More files doesn't always mean more readable -- having to jump between files to read code has a cost, too.

ndushay (approved):
approving, but wondering - when do we use concerns vs using a service?

ndushay:
oh no - test failures!

app/models/collection_version.rb
```diff
@@ -14,65 +14,6 @@ def accessioned?
     %w[first_draft depositing].exclude?(state)
   end
```
ndushay, line 16: is this missing include statement?

mjgiarlo: Sure looks like it.

jcoyne: Oops, I forgot to add that change to the commit. Got it now.

PR #2789, opened 2022-10-04
Delete on Preservation Object storage should be a no-op

### LINKED ISSUE #2785: Possible issue decommissioning an item
See https://app.honeybadger.io/projects/77112/faults/88770594.

I triggered this when decommissioning https://sdr.stanford.edu/works/4659. I don't see any problems with this from the H2 side of things.

peetucket:

Fixes #2785 ~~we aren't sure why a `PurgeJob` was called -- expectation is that you should not be able to purge something in preservation already.  This change will hopefully alert HB in a way that gives us a more useful backtrace than what is in~~ This makes the call to delete the Preservation file object a noop, which is what we intend.  Will prevent this HB alert from happening in the future: https://app.honeybadger.io/projects/77112/faults/88770594

CI

### DIFF
--- a/app/services/active_storage/service/sdr_service.rb
+++ b/app/services/active_storage/service/sdr_service.rb
@@ -15,6 +15,14 @@ module ActiveStorage
         end
       end
 
+      def delete(key)
+        # This is called by ActiveSupport when #destroy is called on AttachedFile due to our use
+        # of ActiveStorage for file storage.  This can happen during a decommission of a work.
+        # We define this as a noop, because we explictly do not want the related preservation
+        # file to be destroyed in this case and we do not want the NotImplementedError exception
+        # to be raised to HB either: https://app.honeybadger.io/projects/77112/faults/88770594
+      end
+
       def self.encode_key(druid, version, filepath)
         [druid, version, filepath].join('/')
       end


app/services/active_storage/service/sdr_service.rb
```diff
@@ -15,6 +15,11 @@ def download(key, &block)
         end
       end
 
+      def delete(key)
+        Honeybadger.context(key)
+        Honeybadger.notify('SdrService#delete called unexpectedly - possible attempt to purge work in preservation')
```
ndushay, line 20: can this include the key in the message?  What is the key, a druid?  Or, if it is a hash, can the druid be in the message?

ndushay (changes requested):

app/services/active_storage/service/sdr_service.rb
```diff
@@ -15,6 +15,11 @@ def download(key, &block)
...
+      end
```
mjgiarlo, line 21: Since we're manually notifying HB here, we can include the context in the `.notify` call:

```ruby
Honeybadger.notify('...', { key: })
```

app/services/active_storage/service/sdr_service.rb
```diff
@@ -15,6 +15,11 @@ def download(key, &block)
         end
       end
 
+      def delete(key)
+        Honeybadger.context(key)
```
mjgiarlo, line 19: Should we use `#split_key` below to make it easier to figure out what we're seeing in the HB context? Or maybe the `key` is good enough?

peetucket:
Note: PR was updated after further discussion to remove any notification of HB.

app/services/active_storage/service/sdr_service.rb
mjgiarlo: I think in the context might be preferable, that way these errors stack on one another in the UI

mjgiarlo (approved):

ndushay (approved):

ndushay:
merge when green, peter!

PR #2800, opened 2022-10-12
prevent user from saving a work with duplicate filenames

### LINKED ISSUE #2798: Validate files so we don't allow users to deposit works with two or more files with the same filename
To prevent issues like this from happening again: https://github.com/sul-dlss/happy-heron/issues/2795

We should validate that users are unable to deposit a work that has two files with the same filename.  Perhaps add a validation to either `AttachedFile` or `WorkForm` or `DraftWorkForm`

peetucket:

Fixes #2798 - we do not want to allow users to save a work that has two files with the same filename, as this will cause problems in various parts of SDR when depositing.

~~I added a new cypress test to verify this and had to add some extra mocking to an unrelated test because of the new validation.~~ Remove: always fails on circleci, reason unknown

I also had to make some changes to the draft work form to allow a user to remove an uploaded file that had not yet been saved.

Localhost

### DIFF
--- a/app/components/works/file_row_component.html.erb
+++ b/app/components/works/file_row_component.html.erb
@@ -3,10 +3,15 @@
     <img data-dz-thumbnail />
   </div>
   <div class="col-5">
-    <% if uploaded? %>
+    <% if saved? %>
       <div data-dropzone-target="fileName">
         <span class="fa-regular fa-file" aria-hidden="true"></span>
         <span><%= filename %>: <wbr><%= number_to_human_size(filesize) %>, <wbr><%= render LocalTimeComponent.new(datetime: date_uploaded, show_time: false) %></span>
+      </div>
+    <% elsif uploaded? %>
+      <div data-dropzone-target="fileName">
+        <span class="fa-regular fa-file" aria-hidden="true"></span>
+        <span><%= filename %>: <wbr><%= number_to_human_size(filesize) %>, <wbr>Now</span>
       </div>
     <% else %>
       <div class="dz-filename" data-dropzone-target="fileName"><span data-dz-name></span></div>
--- a/app/components/works/file_row_component.rb
+++ b/app/components/works/file_row_component.rb
@@ -22,12 +22,18 @@ module Works
     end
 
     def date_uploaded
-      return unless uploaded?
+      return unless saved?
 
       form.object.model.created_at
     end
 
+    # file has been uploaded but may not yet have been saved to the database model (and may not validate)
     def uploaded?
+      !form.object.try(:model)&.blob.nil?
+    end
+
+    # file has been uploaded and saved to the model
+    def saved?
       form.object.persisted?
     end
   end
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -41,6 +41,7 @@ class DraftWorkForm < Reform::Form
 
   validates :subtype, work_subtype: true
   validates :work_type, presence: true, work_type: true
+  validate :unique_filenames
 
   delegate :user_can_set_availability?, to: :collection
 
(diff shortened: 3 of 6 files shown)


app/components/works/file_row_component.html.erb
```diff
@@ -3,11 +3,16 @@
...
       <div data-dropzone-target="fileName">
         <span class="fa-regular fa-file" aria-hidden="true"></span>
         <span><%= filename %>: <wbr><%= number_to_human_size(filesize) %>, <wbr><%= render LocalTimeComponent.new(datetime: date_uploaded, show_time: false) %></span>
       </div>
+    <% elsif uploaded? %>
+      <div data-dropzone-target="fileName">
+        <span class="fa-regular fa-file" aria-hidden="true"></span>
+        <span><%= filename %>: <wbr><%= number_to_human_size(filesize) %>, <wbr>Now</span>
+      </div>
     <% else %>
```
peetucket, line 16: this scenario captures when a user has uploaded a file and clicked "Save" or "Deposit" but it's not valid.  In this case, we have not yet saved to the database, but we will need the UI to render the row correctly so the user can delete that file

app/components/works/file_row_component.rb
```diff
@@ -22,12 +22,18 @@ def filesize
...
 
+    # file has been uploaded but may not yet have been saved to the database model (and may not validate)
     def uploaded?
+      !form.object.try(:model)&.blob.nil?
+    end
+
+    # file has been uploaded and saved to the model
+    def saved?
       form.object.persisted?
     end
```
peetucket, line 38: distinguishing between a persisted file and one that has been uploaded (and has a blob) but is not yet saved to the database... that test for `uploaded?` is a bit unusual -- but without digging into the depths of Reform, was the best I could come up with that worked in all cases

app/forms/draft_work_form.rb
```diff
@@ -71,6 +72,13 @@ def deserialize_embargo(params)
...
   # rubocop:enable Metrics/AbcSize
 
+  # ensure all attached files have a unique filename
+  def unique_filenames
+    filenames = attached_files.map { |file| file.model.filename.to_s }
+
+    errors.add(:attached_files, 'must all have a unique filename.') unless filenames.size == filenames.uniq.size
+  end
+
   def access_from_collection(params)
```
peetucket, line 82: the actual new validation

app/models/attached_file.rb
```diff
@@ -5,10 +5,6 @@ class AttachedFile < ApplicationRecord
   belongs_to :work_version
   has_one_attached :file
 
-  # def blob
-  #   file.blob
-  # end
-
   delegate :blob, to: :file
```
peetucket, line 8: remove cruft

cypress/spec/create_work.cy.js
```diff
@@ -36,6 +36,43 @@ describe('Create work', () => {
...
+      // try to save the deposit
+      cy.get('input.btn[value="Save as draft"]').click()
+
+      // there is a message telling us we cannot have duplicate filenames
+      cy.get('#error_explanation').should('contain', 'Attached files must all have a unique filename.')
+
+      // and we are still on the edit page
+      cy.url().should('include', `/works/${work_id}/edit`)
+    })
+
```
peetucket, line 75: new cypress test for this change

spec/services/work_version_event_description_builder_spec.rb
```diff
@@ -12,6 +12,9 @@
   let(:work) { work_version.work }
   let(:form) { DraftWorkForm.new(work_version:, work:) }
   let(:attached_files) { [] }
+  let(:filename) { 'xml.svg' }
+  let(:blob) { ActiveStorage::Blob.new(filename:) }
+  let(:attached_file) { AttachedFile.new }
```
peetucket, line 18: some new mocking since we need the `AttachedFile` objects to respond to blob and filename for validation purposes

justinlittman (approved):
@peetucket Would you mind rebasing and re-testing locally just to make sure nothing else has changed that affects this?

peetucket:
Rebased, re-tested locally in browser, and tests run locally.  All seems fine. 

Going offline for the week soon, but will leave this unmerged. Feel free to merge when ready.

PR #2803, opened 2022-10-13
Skip check_version in dev

lwrubel:
To get tests to pass in local development 

Local testing


ndushay (approved):
Assuming it's okay to do the mocking for work_version_spec, this looks fine.

PR #2804, opened 2022-10-13
Submit full path as label.

justinlittman:
Support hierarchical file structures.


mjgiarlo:
Thanks for this! I've got this in my branch now. 🙏🏻

PR #2583, opened 2022-09-01
Add file size and date uploaded to upload file display

peetucket:
## Why was this change made? 🤔

Fixes #1359 - add file size and date uploaded to the work edit page for previously uploaded files.

NOTE: See comment trail for updated screenshots

What I have now: 

Screen Shot 2022-09-02 at 12 40 24 PM

What is should look like: https://github.com/sul-dlss/happy-heron/issues/1359#issuecomment-900828723

## How was this change tested? 🤨

Localhost browser

PR #2805, opened 2022-10-13
[HOLD] Support recording and displaying paths of uploaded files

mjgiarlo:

[HOLD] keeping around in case there is useful code ... likely will close when work is complete

Connects to sul-dlss/dor-services-app#4256

This commit adds a new attribute to the AttachedFile model in order to record, and later display, the full source path of an uploaded file (if supplied). It modifies the Dropzone controller to display paths (if present) instead of base filenames, and makes the duplicate filename checker smart about this. It also fixes a bug with duplicate filename checking caused by https://github.com/sul-dlss/happy-heron/pull/2583/files#diff-81e6b6427ea21d10a73dc41dd6b55cbdccd6002a200ddf5cce9b5e26daee19f6R9.

CI, and tested in a deployed env

### DIFF
--- a/app/javascript/controllers/dropzone_controller.js
+++ b/app/javascript/controllers/dropzone_controller.js
@@ -32,7 +32,7 @@ export default class extends Controller {
   }
 
   checkForDuplicates(file) {
-    const fileName = file.fullPath || file.name
+    const fileName = file.fullPath || file.webkitRelativePath || file.name
     // Extract all filenames that are visible
     const fileNames = this.fileNameTargets.map(target => {
       if (target.offsetParent !== null) return target.innerText.trim().split(": ")[0]
@@ -84,7 +84,12 @@ export default class extends Controller {
 
   bindEvents() {
     this.dropZone.on("addedfile", file => {
-      setTimeout(() => {
+      // No hidden files
+      if(file.name.startsWith('.')) {
+        this.dropZone.removeFile(file)
+        return
+      }
+      setTimeout(() => {        
         file.accepted && createDirectUploadController(this, file).start();
         this.fileCount++
         this.enableSubmission()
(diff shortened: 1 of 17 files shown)


app/javascript/controllers/dropzone_controller.js
```diff
@@ -31,13 +31,15 @@ export default class extends Controller {
     }
   }
 
-  checkForDuplicates(fileName) {
+  checkForDuplicates(file) {
+    const fileName = file.fullPath || file.name
     // Extract all filenames that are visible
-    const fileNames = this.fileNameTargets.map(target => { if (target.offsetParent !== null) return target.innerText.trim() })
+    const fileNames = this.fileNameTargets.map(target => {
+      if (target.offsetParent !== null) return target.innerText.trim().split(": ")[0]
```
mjgiarlo, line 38: The `.split(": ")[0]` is a bug fix that is currently breaking duplicate filename checking.

app/javascript/controllers/dropzone_controller.js
```diff
@@ -175,7 +177,9 @@ class DirectUploadController {
 
   start() {
     this.file.controller = this;
-    this.hiddenInput = this.createHiddenInput();
```
mjgiarlo, line 178: Rename this function now that we're attaching *two* hidden inputs to each uploaded file.

peetucket:
Possibly has some overlap with #2800 -- though I don't think there will be any conflicts

peetucket:
Is this PR still relevant?

mjgiarlo:
@peetucket 💬 
> Is this PR still relevant?

I don't know. @justinlittman is this now defunct? (Fine by me if so.)

justinlittman:
I don't think so, but let's hold onto it until H2 HFS work is done. This was hard-fought knowledge.

mjgiarlo:
@justinlittman Time to kill this? Or leave it be?

PR #2822, opened 2022-10-21
readme should indicate you need to clear sessions when changing roles

peetucket:

~~To me it looks like the `bin/dev` script sets a user and role by default and does not allow them to be overridden, as is currently suggested by the README.  This was confusing me during recent debugging until I noticed it.~~

Add note about clearing session when changing roles

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,9 @@ Then run the asset pipeline and webserver:
 bin/dev
 ```
 
-You can change user or roles by setting environment variables:
+You can change user or roles by setting environment variables.  Note that you may need to clear your browser
+session cookies to pick up the new roles set here, as they may be cached from a previous session.
+
 ```shell
 REMOTE_USER=auser@stanford.edu ROLES=dlss:hydrus-app-administrators bin/dev
 ```


README.md
```diff
@@ -27,21 +27,21 @@ To run the H2 application in development mode, set `REMOTE_USER` because we aren
...
+Then run the asset pipeline and webserver (uses a default user that is an admin):
 ```shell
 bin/dev
 ```
 
-You can change user or roles by setting environment variables:
+You can override the user or roles by setting environment variables:
 ```shell
-REMOTE_USER=auser@stanford.edu ROLES=dlss:hydrus-app-administrators bin/dev
+REMOTE_USER=auser@stanford.edu ROLES=dlss:hydrus-app-administrators foreman start -f Procfile.dev
```
mjgiarlo, line 37: I'm curious about changing `bin/dev` to `foreman` here.

peetucket: bin/dev sets default REMOTE_USER and ROLES and cannot overridden (and then calls out to foreman like this).  so if you use bin/dev it won't matter what you put... perhaps a better approach is to make bin/dev smarter to allow overrides

mjgiarlo: Does it? My understanding of the bash in lines https://github.com/sul-dlss/happy-heron/blob/main/bin/dev#L5-L6 is that `"${REMOTE_USER:=auser@stanford.edu}"` means if REMOTE_USER is unset, use auser@stanford.edu, else use the value it's already set to. If this isn't true, we may want to fix `bin/dev`.

peetucket: ahh, i see - i didn't understand that syntax

 I think the issue I am seeing is that the roles are cached in a session variable, and so you need to clear the session cookies in your browser to pick up changes to roles set in the script

so I don't think this change is needed but will instead add a statement noting this finding

mjgiarlo: I've found the same re: the sticky session variables. Thanks for adding that!

mjgiarlo (approved):

PR #2831, opened 2022-10-25
Do not redirect to the root when a depositor views a collection

### LINKED ISSUE #2825: Depositor should be able to view collection details, settings, deposits
Users with depositor permissions for a collection should be able to see all the information for a collection. Currently when trying to click on the collection title on the dashboard, as shown in the video below, they are taken back to the home page. The link in the breadcrumbs does the same thing. It looks like the right page loads briefly in between the clicking and the home page.

I had two users with this same problem. When I gave one of them manager permissions on the collection, the links worked for her immediately, so it seems to have something to do with the depositor permissions.

https://user-images.githubusercontent.com/5558511/197587452-43d4d91d-c251-4f65-935f-6a24cf95336a.mov

aaron-collier:

Fixes #2825 by not enforcing authorization before rendering the admin component which causes a redirect to root.

### DIFF
--- a/app/controllers/collections_controller.rb
+++ b/app/controllers/collections_controller.rb
@@ -4,7 +4,7 @@
 class CollectionsController < ObjectsController
   before_action :authenticate_user!
   before_action :ensure_sdr_updatable
-  verify_authorized except: %i[deposit_button delete_button edit_link dashboard]
+  verify_authorized except: %i[admin dashboard delete_button deposit_button edit_link]
 
   def edit
     collection = Collection.find(params[:id])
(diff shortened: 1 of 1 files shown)


mjgiarlo:
And it's OK not to authorize in the controller because this controller action renders a template that renders a component within a turbo-frame, and that component **only renders** for administrators: https://github.com/sul-dlss/happy-heron/blob/main/app/components/collections/admin_component.rb#L13

So the authorization is being done in a different way.

mjgiarlo (approved):

PR #2854, opened 2022-11-14
Unpin sdr-client and make deposit fixes.

justinlittman:
refs https://github.com/sul-dlss/sdr-client/issues/242

Unbreak things

Infra integration test.


mjgiarlo:
@justinlittman Is this a PR we want to take forward?

Issue #2843, opened 2022-11-10
Add new option for single ZIP file upload


peetucket:
As part of hierarchical file upload, we are removing the ability to upload a folder and instead will ask users to upload a single ZIP, which will be expanded.

This new option is in addition to the current options of uploading files, or using globus, here:

https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005606

The new radio button option language is shown in the UI above.

1. If this option selected, users will are able to select and upload a single ZIP file.  We need to validate that only a single ZIP is provided with this option.

2. Once uploaded, the content of the ZIP will be shown to the depositor in the H2 work edit/show page, something like this (but note we will only allow a single ZIP to be uploaded)

https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005875
 
3. Once deposited, the single ZIP file will be expanded and accessioned as files with hierarchy (the same way the current folder upload works)

- only the zip will be downloadable in H2; individual files will not be downloadable from anywhere in H2
- users will not be able to rename, move, or delete individual files 
- the "hide file" option and the file description entry field will not be available for this upload option

astridu:
3 radio button choices on the deposit page:
<img width="884" alt="image" src="https://user-images.githubusercontent.com/42216982/202080042-200082af-05a2-4680-8e89-9865dee3bd36.png">

astridu:
@amyehodge This is what it might look like if user selects to upload a zip file and sees the expanded folders.

<img width="669" alt="image" src="https://user-images.githubusercontent.com/42216982/202082607-e83b1239-b530-49b3-9bde-887932c58ec2.png">

amyehodge:
> Questions for PO:
>
> Do we want users to be able to drag and drop the single ZIP or click the browse button and select a single file?

Yes, we want users to be able to either drag and drop the single ZIP or click the browse button and select a single file. Drag and drop is a must have. Browse button is slightly less important.

amyehodge:
> Questions for PO:
>
> Updated verbiage for options and any instructions

For the first option, I'm concerned that saying "individual files" will make some users think they can only upload one file.
For the last option, I'm not sure if the rest of the language that @astridu suggested for this one is accurate until I see how that workflow is going to work out. It may need to also indicate that they may need to install software.


My suggested text for the radio buttons:

- **Upload one or more files (< 10GB) to be displayed as a flat list.**
[...]

PR #2857, opened 2022-11-15
prepare for multiple selectable upload types

peetucket:

In this workcycle, we will need to expand the options available for users to select from when adding files to an object from two choices (browser upload and globus) to three choices: 

- browser upload 
- globus
- single ZIP file to be expanded

Our current database schema uses a `globus` boolean type column on the `work_version` to designate if an object uses globus uploads or not.  This PR switches the name of the column, makes it an enum to allow for more than two options, and fixes up the code to use the new schema.

This is pre-requisite work for #2843 (and possibly others).

Exiting test suite (updated as needed) and localhost

### DIFF
--- a/app/forms/draft_work_form.rb
+++ b/app/forms/draft_work_form.rb
@@ -96,7 +96,7 @@ class DraftWorkForm < Reform::Form
     property :file, virtual: true
     property :_destroy, virtual: true, type: Dry::Types['params.nil'] | Dry::Types['params.bool']
   end
-  property :globus, on: :work_version
+  property :upload_type, on: :work_version
 
   collection :contact_emails, populator: ContactEmailsPopulator.new(:contact_emails, ContactEmail),
                               prepopulator: ->(*) { contact_emails << ContactEmail.new if contact_emails.blank? },
@@ -163,7 +163,7 @@ class DraftWorkForm < Reform::Form
   def save_model
     super
     # if the user selects globus uploads, we cannot have any attached files
-    work_version.attached_files.destroy_all if work_version.globus
+    work_version.attached_files.destroy_all if work_version.globus?
     dedupe_keywords
     work.update(head: work_version)
   end
--- a/app/forms/work_form.rb
+++ b/app/forms/work_form.rb
@@ -7,9 +7,7 @@ class WorkForm < DraftWorkForm
   validates :abstract, :access, :title, presence: true, allow_nil: false
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
   validates :attached_files, length: { minimum: 1, message: 'Please add at least one file.' },
-                             unless: lambda {
-                                       ::ActiveRecord::Type::Boolean.new.cast(globus)
-                                     }
+                             if: -> { upload_type == 'browser' }
   validates :contact_emails, length: { minimum: 1, message: 'Please add at least contact email.' }
   validates :license, presence: true, inclusion: { in: License.license_list }
   validates :authors, length: { minimum: 1, message: 'Please add at least one author.' }
--- a/app/models/work_version.rb
+++ b/app/models/work_version.rb
@@ -41,6 +41,13 @@ class WorkVersion < ApplicationRecord
     world: 'world'
   }
 
+  # provides helper method to infer upload type... e.g. work_version.globus?
+  enum upload_type: {
+    browser: 'browser',
+    globus: 'globus',
+    zipfile: 'zipfile'
+  }
+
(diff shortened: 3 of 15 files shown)


app/forms/work_form.rb
```diff
@@ -7,9 +7,7 @@ class WorkForm < DraftWorkForm
   validates :abstract, :access, :title, presence: true, allow_nil: false
   validates :keywords, length: { minimum: 1, message: 'Please add at least one keyword.' }
   validates :attached_files, length: { minimum: 1, message: 'Please add at least one file.' },
-                             unless: lambda {
-                                       ::ActiveRecord::Type::Boolean.new.cast(globus)
-                                     }
+                             if: -> { upload_type == 'browser' }
   validates :contact_emails, length: { minimum: 1, message: 'Please add at least contact email.' }
```
peetucket, line 11: note: this validation will get a bit more complicated once we add option 3 (single ZIP file) in a future PR

app/models/work_version.rb
```diff
@@ -41,6 +41,12 @@ class WorkVersion < ApplicationRecord
     world: 'world'
   }
 
+  enum upload_type: {
+    browser: 'browser',
+    globus: 'globus',
+    hfszip: 'hfszip'
+  }
+
```
peetucket, line 49: not sure if we can come up with better names for these options (not shown to user, but stored in the database)

app/models/work_version.rb
```diff
@@ -60,6 +66,10 @@ def clean_abstract
     Rails.logger.info { "Null character stripped from abstract for work_version id #{id}." }
   end
 
+  def globus
+    upload_type == 'globus'
+  end
+
```
peetucket, line 72: helper method so that various other parts of the existing codebase that check to see if a work is using globus or not will work without further change

we can add additional methods as needed for the other types in future PRs

db/migrate/20221115215744_work_upload_type.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def down
+    add_column :work_versions, :globus, :boolean, default: false, null: false
+
+    WorkVersion.where(upload_type: 'globus').update_all(globus: true)
+    WorkVersion.where(upload_type: 'browser').update_all(globus: false)
+
+    remove_column :work_versions, :upload_type
+  end
+end
```
peetucket, line 22: this migration cleanly allows us to move back and forth if needed without destroying any information

app/services/work_observer.rb
```diff
@@ -42,7 +42,7 @@ def self.after_deposit_complete(work_version, _transition)
             mailer.deposited_email
           end
     job.deliver_later
-    mailer.globus_deposited_email.deliver_later if work_version.globus && Settings.notify_admin_list
+    mailer.globus_deposited_email.deliver_later if work_version.globus? && Settings.notify_admin_list
   end
```
peetucket, line 46: make use of our new enum helper method instead of the previous model attribute

app/forms/draft_work_form.rb
```diff
@@ -163,7 +163,7 @@ def save!(*)
   def save_model
     super
     # if the user selects globus uploads, we cannot have any attached files
-    work_version.attached_files.destroy_all if work_version.globus
+    work_version.attached_files.destroy_all if work_version.globus?
     dedupe_keywords
```
peetucket, line 167: make use of our new enum helper method instead of the previous model attribute

db/migrate/20221115215744_work_upload_type.rb
```diff
@@ -0,0 +1,22 @@
+class WorkUploadType < ActiveRecord::Migration[7.0]
+  def up
+    # new column to store which type of upload mechanism the user wants for their files
+    add_column :work_versions, :upload_type, :string, default: 'browser', null: false
```
justinlittman, line 4: I'm wondering whether this migration is going to work on a db with a existing work versions? My suspicion is that you'll need to set `null: true` and then do the update below and then change it to `false`.

Have you tested this on a db with data?

peetucket: I tested it on localhost which had a few rows of data and it worked fine.  I believe the default value is what allows null to be false in the migration.  Perhaps I can deploy to qa as well as see what happens?

peetucket: Deployed to QA and migration ran cleanly with 385 work versions across 253 works (5 were globus type, the rest were file upload).  No upload_types were left as nil or empty.

justinlittman (approved):

PR #2863, opened 2022-11-18
do not have a default upload option selected

### LINKED ISSUE #2859: When creating a new work, no default option for uploading files should be selected
See the design in this ticket: https://github.com/sul-dlss/happy-heron/issues/2843#issuecomment-1316292714

Currently when a new work page is opened, the browser file upload radio button is selected by default and shows the browser upload drag/drop area.  There is a concern this will obscure the other two upload options (globus/ZIP).  

So when the page is first opened, no radio button should be selected, and no controls for the options are shown.  Once a user selects an option, the controls for that option are shown/hidden as needed if they select other options (as is currently done).  Saving the draft will select the current option, which will be re-selected the next time they open the page.

peetucket:

Fixes #2859 -- do not select any radio button by default when creating a new work, which forces the user to select one

We still need to validate that something is selected to save as draft, but we cannot enforce at the model level because PURL reservations still need to be created without a selection.

Localhost, added a test, modified existing tests to pre-select the previous "browser upload" default.

### DIFF
--- a/app/javascript/controllers/file_uploads_controller.js
+++ b/app/javascript/controllers/file_uploads_controller.js
@@ -1,13 +1,13 @@
 import { Controller } from "stimulus"
 
 export default class extends Controller {
-  static targets = ["fileUploads", "globusRadioButton"]
+  static targets = ["fileUploads", "globusRadioButton", "browserRadioButton"]
 
   connect() {
     this.updatePanelVisibility()
   }
 
   updatePanelVisibility(_event) {
-    this.fileUploadsTarget.hidden = this.globusRadioButtonTarget.checked
+    this.fileUploadsTarget.hidden = !this.browserRadioButtonTarget.checked
   }
 }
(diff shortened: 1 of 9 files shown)


app/javascript/controllers/file_uploads_controller.js
```diff
@@ -1,13 +1,13 @@
...
+  static targets = ["fileUploads", "globusRadioButton", "browserRadioButton"]
 
   connect() {
     this.updatePanelVisibility()
   }
 
   updatePanelVisibility(_event) {
-    this.fileUploadsTarget.hidden = this.globusRadioButtonTarget.checked
+    this.fileUploadsTarget.hidden = !this.browserRadioButtonTarget.checked
   }
```
peetucket, line 12: the filesUpload panel will always be hidden unless it is explicitly selected

db/migrate/20221117233130_no_upload_type_default.rb
```diff
@@ -0,0 +1,11 @@
...
+  def up
+    change_column_default(:work_versions, :upload_type, nil)
+    change_column_null(:work_versions, :upload_type, true)
+  end
+
+  def down
+    change_column_default(:work_versions, :upload_type, 'browser')
+    change_column_null(:work_versions, :upload_type, false)
+  end
+end
```
peetucket, line 11: we actually need to allow the column to be nil and have no default...since otherwise PURL reservations will not work correctly (unless we default them to "browser" upload type as was happening until now)

we will force the user to make a choice by a validation on the work form

mjgiarlo (approved):

PR #2868, opened 2022-11-21
add new rubocops

ndushay:


ndushay (commented):
whoa - rubocop autofix moved methods around in the controllers!

lwrubel (approved):

Issue #2871, opened 2022-11-22
Handle non-existent Globus user when saving new work as draft


mjgiarlo:
When the user clicks "save as draft" (for the first time only) on a deposit where they selected "use globus", see if the user exists in Globus via `globus_client` gem.

If the user does not exist in globus:

* Redirect to work detail page displaying this message to the user on the top of the screen:
  * > Please check your email for further instructions on how to proceed with your deposit of files into Globus.
* Email user with instructions explaining how to setup Globus and ask them to return to work edit page once done and click "Save As Draft" again when complete.

(If the user does exist, that's covered by a separate ticket: #2847)

Issue #2847, opened 2022-11-11
When globus option is selcted and the deposit is saved as draft, trigger new job


peetucket:
The new job should:

- See if the user exists in globus via globus client gem

If the user exists in globus:
- Create new endpoint that user will put files into using globus client gem and set permissions correctly
- Email user with this endpoint URL and other instructions (need text from PO)
- Email sdr-support indicating a user has begun a globus deposit with info about the work

If the user does not exist in globus:
- Email user with instructions on how to get started with globus (need text from PO)

astridu:
@peetucket @amyehodge Should the 'Save as draft' button be disabled if the user does not have a Globus account?

amyehodge:
> @peetucket @amyehodge Should the 'Save as draft' button be disabled if the user does not have a Globus account?

@astridu I would not want to disable that button, because the user should be able to save metadata entries that they may have spent a lot of time on.

amyehodge:
Seems like we need to figure out what to do when they aren't in globus client gem. Up to this point, this has been nearly everyone except for our repeat customers.

peetucket:
The challenge we face is that as currently scoped the "Save as draft" is what kicks off the process of setting up their Globus folder.  If they don't have an account yet, we can't do that.  So the approaches are:

1. Don't let them save as draft or do anything until they create a globus account (would need to be very well messaged at the top).  But I can definitely see Amy's point of this being undesirable.
2. Let them save as a draft, then detect they don't have an account, and then send them an alternate email giving them further instructions.  They would then need to come back to H2 when done and save as draft again to get the folders setup.  

For option 2, we may also want to keep track of the fact they already got the instruction email so if they return to save as a draft again but still haven't setup their globus, they won't get the "setup" email again (though is perhaps too much complication for a limited edge case).

amyehodge:
Users really need to be able to save their other work. If we didn't show them the rest of the form (so they wouldn't end up with unsaved data) until they create the globus account, it wouldn't be as big an issue, but that's a completely different design for this entire form.

Option 2 is sounding quite cumbersome. I don't have a great ideas about this at the moment but will keep thinking.

astridu:
This may be out of scope and is basically the same as no. 1 above, but the SSO to Globus only takes seconds so this might not be so bad. This is how the flow might be:

1. Don't show the whole deposit screen, only show this question first: Is your deposit > 10GB?
[...]

PR #2882, opened 2022-11-30
new globus setup states

### LINKED ISSUE #2865: Record status of draft globus deposit
The globus workflow for users (https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/470591811_Globus_Workflow) looks like:

1. User starts a new draft, selects globus radio button
2. Deposit button disabled 
3. User clicks Save As Draft, actions happen depending on if they are a new globus user or not (user exists: #2847, user does not exist: #2871)
4. User returns to work and clicks "Deposit", actions happen (#2848)

Thus we need to know when to enable/disable the "Deposit" button, as it depends on if we were able to create a folder for the user after they clicked "Save As Draft" and the actions in #2847 have been taken.

This means we likely need a new state for draft works.

The state for an object where the user does not exist in Globus yet and needs to complete setup could be: "Draft - pending globus setup".  In this state, the "Deposit" button would not be enabled.

Once setup in globus is complete, the state could become the usual draft state: "Draft - Not deposited", which would enable the "Deposit" button.

Current state diagram: https://github.com/sul-dlss/happy-heron/wiki/State-machine-diagrams

We already know if a user has selected "globus" as the upload option (it's in the database for a work_version), so really what we need to know is if we've be able to create a folder for them in globus (from #2847).  If we have, then next time the user comes to the edit form, the Deposit button can be enabled.  If we haven't, this means we haven't been able to create a folder for them in globus because they first need to set themselves up an account in Globus.  In that case, the Deposit button will continue to be disabled on the work edit form.

peetucket:

Fixes #2865 -- gives us two new states for work_versions we will need for the globus integration

1. When a user selects the globus upload option, and we are able to create their globus endpoint for them, no new state is needed.  See #2847
2. When a user selects the globus upload option, and we are NOT able to create their globus endpoint for them since they first need to complete their globus setup, we DO need a new state, and this is triggered via `work_version.globus_setup_pending!`.  See #2871.  This (a) shows the state to the user, (b) allows us to selectively show the new button that will let the user know their setup is complete, and (c) lets us disable the Deposit button.
3. Once the user tells us they have completed their globus setup, we can try and create the globus endpoint folder again, and if it succeeds this time, we can transition back to the draft state so they can proceed with adding their files to globus and completing the deposit, done via `work_version.globus_setup_complete!`

Note that since a user can go into the new state from either a first_draft or a version_draft, we actually have two states we can end up in, so that we know which state we will transition back into.

Localhost and existing tests


app/models/concerns/work_version_state_machine.rb
```diff
@@ -38,6 +38,18 @@ module WorkVersionStateMachine
...
+      event :globus_setup_pending do
+        transition first_draft: :globus_setup_first_draft
+        transition version_draft: :globus_setup_version_draft
+      end
+
+      # event occurs when a user has completed initial globus setup; go back to regular draft state
+      event :globus_setup_complete do
+        transition globus_setup_first_draft: :first_draft
+        transition globus_setup_version_draft: :version_draft
+      end
```
mjgiarlo, line 51: Should these two new states be added to line 78, so the `update_metadata` event can still fire off on works in these states?

peetucket: right, and this is needed because a save as draft will trigger that event, correct?  and this should still be allowed when we are pending globus setup

mjgiarlo: @peetucket I believe so

mjgiarlo (commented):
1. Would you consider adding test coverage for the new transitions and events to the [`state machine flow` block of work version tests](https://github.com/sul-dlss/happy-heron/blob/main/spec/models/work_version_spec.rb)?
2. For after the PR is merged, can you [update the state machine diagrams](https://github.com/sul-dlss/happy-heron/wiki/State-machine-diagrams)?

peetucket:
1.  Did you have specific tests in mind?  The current tests seem to look at behavior we expect to happen when a transition occurs, in our current case we just change the state itself back and forth.  So not super interesting (though I suppose could be built on when other behaviors are attached to the transition).
2. Yes, thanks for the reminder.

mjgiarlo:
@peetucket Yeah, I don't think they would be super interesting at this point but even a handful of basic tests would make it such that if someone accidentally removed or mucked with the code already in this branch (e.g., a rebase gone awry), at least one test would break.

mjgiarlo (approved):

Issue #2884, opened 2022-11-30
Deposit button enabled for globus objects when ready


peetucket:
#2875 disables the deposit button via JS when the user selects the globus upload option (and the feature flag is set).  However, we need to be able to re-enable the deposit button once the work in #2847 and #2848 is complete

Basically, for a user who is setup in globus already, once they have clicked "save as draft", and we've been able to setup their globus endpoint, we can are now read to enable the Deposit button.  This can happen once they check a new checkbox called "I've finished uploading my files to Globus".

The Deposit button is disabled is (1) the new work is not yet persisted (i.e. not yet saved as draft), but they've selected the globus upload radio button (which is handled by the work in #2875), (2) the new work is persisted and is in the new states defined in #2882 (`globus_setup_first_draft` or `globus_setup_version_draft`), and (3) they've selected the globus upload radio button, the work has been saved as draft and an endpoint was created (so work is in the "first_draft" or "version_draft" state), but they haven't yet checked the "I've finished uploading my files to Globus" checkbox and saved as draft to pull the files from globus.

peetucket:
@justinlittman wanted to check in with you before i pick up this ticket having to do with fixing how the deposit button gets enabled/disabled for globus deposits (since it may be part of what you are currently doing)

mjgiarlo:
I believe https://github.com/sul-dlss/happy-heron/pull/2913 has unblocked this issue. Confirm/deny, @justinlittman?

justinlittman:
I suspect this ticket is done. Confirm/deny @amyehodge ?

amyehodge:
I'm pretty sure this worked properly when I tested it before the break. I can't verify that it works now because of the other issues I reported, but I think we can close this.

mjgiarlo:
OK, closing. thanks!

PR #2887, opened 2022-12-02
globus upload workflows

### LINKED ISSUE #2847: When globus option is selcted and the deposit is saved as draft, trigger new job

### LINKED ISSUE #2855: Add a page/Google doc with instructions for users on how to sign up for Globus and link to it
Users who have never used globus before but want to select that as an upload option need to have logged in at least once so they will be known to the API.  We need to provide them with some instructions that can be linked to from the work edit page if they select that option.

Ask @amyehodge for some verbiage for this page.

### LINKED ISSUE #2871: Handle non-existent Globus user when saving new work as draft

peetucket:

Fixes #2871 and fixes #2847 and fixes #2855 - handle non-existing and existing globus users saving works as drafts

Adds:

- New job to run all globus setup tasks
- Emails for user needs to activate globus and for sending user globus endpoint link
- State transitions for when user selects globus upload option and back again

To do:
- [x] add actual link to globus documentation google help doc in globus account setup and endpoint emails (from Amy)
- [x] figure how/when/where to send globus endpoint created email
- [x] figure out how to use globus client gem in different environments for development (qa/stage) and troubleshoot
- [x] add globus config/keys to shared configs for all environments
- [x] add test coverage
- [x] integrate globus client gem where needed
- [x] add banner when user needs to activate globus account with button to do this
- [x] add button to dashboard for activating globus account when needed
- [x] show message next to the deposit button when globus selected

Next, we need to do #2884 so that the deposit button is re-enabled for globus uploads when an endpoint exists

NOTE: none of these features are available unless you set the `Settings.globus_upload` feature flag to true (currently set to false).  You can do this in `settings.local.yml`.  Also note that this will attempt to make actual globus API calls, so you will also need to configure the globus credentials in your `settings.local.yml`.  Current values for each environment are in h2 shared configs, e.g. https://github.com/sul-dlss/shared_configs/blob/sul-h2-qa/config/settings/production.yml for QA

Localhost
Needs tests


app/components/works/buttons_component.html.erb
```diff
@@ -1,3 +1,11 @@
+<% if Settings.globus_upload %>
+  <div data-deposit-button-target="globusMessage" class="row justify-content-end">
+    <div class="col-8">
+      <p>You can save this deposit as a draft.  Next, you will receive an email with instructions on next steps in Globus.
```
justinlittman, line 4: This sentence as "next" in it twice.

app/jobs/globus_setup_job.rb
```diff
@@ -0,0 +1,45 @@
...
+  def perform(work_version)
+    druid = work_version.work.druid # may be nil
+    user = work_version.work.owner
+    Honeybadger.context({ work_version_id: work_version.id, druid:,
+                          work_id: work_version.work.id, depositor_sunet: user.sunetid })
+
+    if globus_user_exists?(user.email) && work_version.globus_endpoint.blank?
+      # user is known to globus but doesn't have a globus endpoint yet, so create it and send the email
+      create_globus_endpoint(work_version)
+      WorkObserver.globus_endpoint_created(work_version) # send email with globus endpoint
```
justinlittman, line 17: Seems odd that this is being invoked through WorkObserver. Typically, WorkObserver is only invoked via the state machine.

app/jobs/globus_setup_job.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def create_globus_endpoint(work_version)
+    user = work_version.work.owner
+    # e.g. 'mjgiarlo/work1234/version1'
+    endpoint_path = "#{user.sunetid}/work#{work_version.work.id}/version#{work_version.version}"
+    success = GlobusClient.mkdir(user_id: user.email, path: endpoint_path)
+
+    # TODO: Error handling?
+    return unless success
```
justinlittman, line 41: raise here?

justinlittman (changes requested):
A few minor changes / questions.

app/jobs/globus_setup_job.rb
peetucket: removed `WorkObserver` method and send email directly now from the job

app/components/works/buttons_component.html.erb
peetucket: adjusted text

justinlittman (approved):

Issue #2886, opened 2022-12-02
Incorrect fileset external identifiers on update


justinlittman:
H2 is creating incorrect external identifiers when performing an update.

See, for example, https://argo.stanford.edu/view/hp267yt1382:
![image](https://user-images.githubusercontent.com/588335/205360631-17769f5d-1640-4bf1-8fb5-94a2b15b533a.png)

This is occurring here: https://github.com/sul-dlss/happy-heron/blob/main/app/services/cocina_generator/structural/generator.rb#L54-L56

Instead, H2 should be re-using the existing fileset identifier (if there is one), otherwise it should be generating one using the approach from https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/id_generator.rb. (Perhaps this should be moved to Cocina gem?)

mjgiarlo:
@justinlittman Is this for the current work cycle, or the first responder, or neither? Thanks! (Doing board grooming.)

justinlittman:
@mjgiarlo Either, but not one.

Issue #2879, opened 2022-11-30
Files should be checked for a change to hide/unhide file setting when sending updates


amyehodge:
Currently, in depositing a new work version, h2 checks whether files are previously preserved, based on filename. It uses the existing structural cocina if there's a filename match. It does not look to see if the hide/unhide file option in h2 work version has changed and no longer matches the existing cocina. The cocina generator should do this. 

The end result here is that the user checks/unchecks the box and the item is saved, but no change is actually made and the user is not informed that the change was not made.

Note that this issue is not currently affecting any user that we are aware of, but was raised in a work cycle discussion.

andrewjbtw:
I am not sure how this looks in H2's internal settings, but the hide/unhide actions correspond to the Cocina publish/shelve settings.

Hide --> publish=no, shelve=no
Unhide (the default) --> publish=yes,shelve=yes

Note that in both of those cases, preserve=yes.

I think as long as a file was previously preserved, the system will publish and shelve it when "publish=no,shelve=no" is flipped to "publish=yes,shelve=yes". This is what happens when those settings are changed via Argo (though it's the accessionWF that actually does the publishing and shelving).

mjgiarlo:
@amyehodge @andrewjbtw Is this for the current work cycle, or the first responder, or neither? Thanks! (Doing board grooming.)

andrewjbtw:
My understanding from Slack was that the fix for https://github.com/sul-dlss/happy-heron/issues/2878 should also address this. The more general case would be phrased as:

> H2 should update Cocina structural when any property that maps to Cocina structural changes, even if the file itself isn't changed. 

I would consider that an FR ticket as a regression that came from a previous WC. But it's relevant to the current workcycle in that it will affect the behavior of versioning the hierarchical file layouts as well.

PR #2889, opened 2022-12-05
Always rebuild structural cocina

### LINKED ISSUE #2878: Files should be checked for a changed description when sending updates
Currently, in depositing a new work version, h2 checks whether files are previously preserved, based on filename. It uses the existing structural cocina if there's a filename match. It does not look to see if the description of the file in h2 work version has changed and no longer matches the existing cocina file's label. The cocina generator should do this. See druid:hp267yt1382 as an example.

lwrubel:
Resolves #2878, #2879, #2886 to

- update cocina structural metadata on later versions of works, rebuilding cocina (rather than copying) for current and existing filesets. Rebuilding cocina will update the label, publish/shelve options, and access rights.
- fixes the externalIdentifier format for filesets and files

Integration, unit. 

DONE: This also needs sul-dlss/sdr-api#535 to be deployed to have the file ids match the fileSet ids. 

### DIFF
--- a/app/services/cocina_generator/structural/file_generator.rb
+++ b/app/services/cocina_generator/structural/file_generator.rb
@@ -4,16 +4,20 @@ module CocinaGenerator
   module Structural
     # This generates a Cocina File for a work
     class FileGenerator
-      def self.generate(work_version:, attached_file:)
-        new(work_version:, attached_file:).generate
+      def self.generate(work_version:, attached_file:, resource_id: nil, cocina_file: nil)
+        new(work_version:, attached_file:, resource_id:, cocina_file:).generate
       end
 
-      def initialize(work_version:, attached_file:)
+      def initialize(work_version:, attached_file:, resource_id: nil, cocina_file: nil)
         @work_version = work_version
         @attached_file = attached_file
+        @resource_id = resource_id
+        @cocina_file = cocina_file
+
+        raise 'Either resource_id or cocina_file should be provided.' if resource_id.nil? && cocina_file.nil?
       end
 
-      attr_reader :work_version, :attached_file
+      attr_reader :work_version, :attached_file, :resource_id, :cocina_file
 
       def generate
         return nil unless blob
@@ -25,6 +29,7 @@ module CocinaGenerator
         end
       end
 
+      # rubocop:disable Metrics/AbcSize
       def request_file_attributes
         {
           type: Cocina::Models::ObjectType.file,
@@ -33,22 +38,24 @@ module CocinaGenerator
           filename:,
           access:,
           administrative:,
-          hasMimeType: blob.content_type,
+          hasMimeType: cocina_file&.hasMimeType || blob.content_type,
           hasMessageDigests: message_digests,
-          size: blob.byte_size
+          size: cocina_file&.size || blob.byte_size
         }
       end
+      # rubocop:enable Metrics/AbcSize
 
       def file_attributes
         request_file_attributes.merge(externalIdentifier: external_identifier)
       end
 
       def filename
-        blob.filename.to_s # File.basename(filename(blob.key))
+        cocina_file&.filename || blob.filename.to_s # File.basename(filename(blob.key))
       end
 
       def external_identifier
-        "#{work_version.work.druid}/#{filename}" if work_version.work.druid
+        # see https://github.com/sul-dlss/dor-services-app/blob/main/app/services/cocina/id_generator.rb
+        cocina_file&.externalIdentifier || blob.signed_id
       end
 
       def administrative
@@ -60,10 +67,11 @@ module CocinaGenerator
       end
 
       def message_digests
-        [
-          { type: 'md5', digest: base64_to_hexdigest(blob.checksum) },
-          { type: 'sha1', digest: Digest::SHA1.file(file_path(blob.key)).hexdigest }
-        ]
+        cocina_file&.hasMessageDigests ||
+          [
+            { type: 'md5', digest: base64_to_hexdigest(blob.checksum) },
+            { type: 'sha1', digest: Digest::SHA1.file(file_path(blob.key)).hexdigest }
+          ]
       end
 
       delegate :blob, to: :attached_file
--- a/app/services/cocina_generator/structural/generator.rb
+++ b/app/services/cocina_generator/structural/generator.rb
@@ -24,38 +24,63 @@ module CocinaGenerator
       end
 
       # 1) Start with the existing filesets from SDR
-      # 1) Only keep the filesets that have attached_files that are preserved
-      # 2) Add new filesets that have attached_files in the local store.
+      # 2) Rebuild filesets that have attached_files that are preserved
+      # 3) Add new filesets that have attached_files in the local store.
       def build_filesets
-        filesets = find_preserved_filesets
-        filesets + work_version.staged_files.map.with_index(filesets.size + 1) { |af, n| build_fileset(af, n) }
-      end
-
-      # @return [Array<Cocina::FileSet>] the list of items where all files are already preserved.
-      def find_preserved_filesets
-        return [] unless cocina_obj
-
-        cocina_obj.structural.contains.select do |fileset|
-          existing_file_names = Set.new(fileset.structural.contains.map(&:filename))
-          preserved_file_names = Set.new(work_version.preserved_files.map(&:filename).map(&:to_s))
-          preserved_file_names.superset?(existing_file_names)
+        filesets = find_preserved_filesets.to_h do |filename, fileset|
+          [
+            filename,
+            build_fileset(attached_file: get_preserved_file(filename), fileset:)
+          ]
         end
+        filesets.values + work_version.staged_files.map { |af| build_fileset(attached_file: af) }
       end
 
-      def build_fileset(attached_file, offset)
+      # @return [Hash] map of filename to matching fileset's cocina for preserved files.
+      def find_preserved_filesets
+        return {} unless cocina_obj
+
+        preserved_file_names = work_version.preserved_files.map(&:filename).map(&:to_s)
+        create_filesmap(preserved_file_names)
+      end
+
+      def get_preserved_file(filename)
+        work_version.preserved_files.find { |attached_file| attached_file.filename.to_s == filename }
+      end
+
+      # rubocop:disable Rails/IndexBy
+      def create_filesmap(preserved_file_names)
+        # h2 only has one file per fileset
+        cocina_obj.structural.contains
+                  .select { |fileset| preserved_file_names.include?(fileset.structural.contains.first.filename) }
+                  .to_h { |fileset| [fileset.structural.contains.first.filename, fileset] }
+      end
+      # rubocop:enable Rails/IndexBy
+
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/MethodLength
+      def build_fileset(attached_file:, fileset: nil)
+        # h2 only has one file per fileset
+        cocina_file = fileset&.structural&.contains&.first
+        resource_id = SecureRandom.uuid if cocina_file.blank?
         {
           type: Cocina::Models::FileSetType.file,
           version: work_version.version,
           label: attached_file.label,
           structural: {
-            contains: [FileGenerator.generate(work_version:, attached_file:)]
+            contains: [FileGenerator.generate(work_version:, attached_file:, resource_id:, cocina_file:)]
           }
-        }.tap do |fileset|
-          if work_version.work.druid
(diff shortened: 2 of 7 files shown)


spec/jobs/deposit_job_spec.rb
```diff
@@ -167,6 +167,35 @@
...
+        let(:attached_file_updated) { build(:attached_file, label: 'My changed label') }
+        let(:second_work_version_metadata_only) do
+          build(:work_version, work:, attached_files: [attached_file_updated], version: 2,
+                               version_description: 'Updated metadata')
+        end
+
+        before do
+          work.work_versions = [first_work_version, second_work_version_metadata_only]
+          allow(SdrClient::Find).to receive(:run).and_return(cocina.to_json)
+          allow(attached_file_updated).to receive_message_chain(:file, :blob).and_return(blob) # rubocop:disable RSpec/MessageChain
```
lwrubel, line 181: I'm not sure why it should be necessary to mock getting a blob back in this test. I don't want to upload new blobs for this metadata-only second version. The goal is to get a test of changing the label on an otherwise unchanged file in a new version, and then add expectations that the size and publish/shelve match the returned cocina from line 90.

lwrubel: Worked with Justin to set up a new factory trait to generate a preserved blob instead.

app/services/cocina_generator/structural/file_generator.rb
```diff
@@ -4,16 +4,20 @@ module CocinaGenerator
...
-      def self.generate(work_version:, attached_file:)
-        new(work_version:, attached_file:).generate
+      def self.generate(work_version:, attached_file:, resource_id:, cocina_file: nil)
+        new(work_version:, attached_file:, resource_id:,  cocina_file:).generate
       end
 
-      def initialize(work_version:, attached_file:)
+      ID_NAMESPACE = 'https://cocina.sul.stanford.edu'
+
+      def initialize(work_version:, attached_file:, resource_id:, cocina_file: nil)
```
mjgiarlo, line 13: It looks like `resource_id` is a required argument even when the `cocina_file` is non-`nil`, but in that case it's never used. I'm wondering about ways to refactor this class to reflect that it has two modes---one that generates from a blob and another that generates from a Cocina instance---but I've not yet come up with anything that's better the current (small) number of branches where we need to discriminate between the two cases.

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
       end
 
       # 1) Start with the existing filesets from SDR
-      # 1) Only keep the filesets that have attached_files that are preserved
-      # 2) Add new filesets that have attached_files in the local store.
+      # 2) Rebuild filesets that have attached_files that are preserved
+      # 3) Add new filesets that have attached_files in the local store.
```
mjgiarlo, line 30: 🥇

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
       def build_filesets
         filesets = find_preserved_filesets
-        filesets + work_version.staged_files.map.with_index(filesets.size + 1) { |af, n| build_fileset(af, n) }
+        # rebuild the the ones that are still preserved
+        filesets.each do |filename, fileset|
+          rebuilt_fileset = build_fileset(attached_file: get_preserved_file(filename), fileset:)
+          filesets[filename] = rebuilt_fileset
+        end
```
mjgiarlo, line 37: It looks like this snippet is iterating over a hash and updating each value? If so, I wonder about using a different iteration strategy is worth it. E.g.:

```suggestion
        filesets = find_preserved_filesets.to_h do |filename, fileset| 
          [
            filename, 
            build_fileset(attached_file: get_preserved_file(filename), fileset:)
          ] 
        end 
```

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
-        return [] unless cocina_obj
+        return {} unless cocina_obj
+
+        preserved_file_names = work_version.preserved_files.map(&:filename).map(&:to_s)
+        create_filesmap(preserved_file_names)
+      end
+
+      def get_preserved_file(filename)
+        work_version.preserved_files.find { |attached_file| attached_file.filename.to_s == filename }
+      end
```
mjgiarlo, line 51: 👍🏻

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
 
-        cocina_obj.structural.contains.select do |fileset|
-          existing_file_names = Set.new(fileset.structural.contains.map(&:filename))
-          preserved_file_names = Set.new(work_version.preserved_files.map(&:filename).map(&:to_s))
-          preserved_file_names.superset?(existing_file_names)
+      def create_filesmap(preserved_file_names)
+        files_map = {}
+        cocina_obj.structural.contains.to_a.each do |fileset|
```
mjgiarlo, line 55: Is the `#to_a` being used here to handle a `nil` contains? (Because `contains` should already be an array, right?)

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
+          existing_file_name = fileset.structural.contains.first.filename
+          files_map[existing_file_name] = fileset if preserved_file_names.include?(existing_file_name)
         end
+        files_map
       end
 
-      def build_fileset(attached_file, offset)
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/MethodLength
+      def build_fileset(attached_file:, fileset: nil)
```
mjgiarlo, line 65: Dig the method signature change!

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
+        cocina_file = fileset&.structural&.contains&.first
+        resource_id = SecureRandom.uuid
```
mjgiarlo, line 67: Is it possible the fileset already has an identifier that we'd want to keep instead of always generating a UUID, or is that a thing we don't need to fret about in this context?

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
+          files_map[existing_file_name] = fileset if preserved_file_names.include?(existing_file_name)
         end
+        files_map
       end
 
-      def build_fileset(attached_file, offset)
+      # rubocop:disable Metrics/AbcSize
+      # rubocop:disable Metrics/MethodLength
+      def build_fileset(attached_file:, fileset: nil)
+        cocina_file = fileset&.structural&.contains&.first
```
mjgiarlo, line 66: It's always the first because that's part of the current H2 design, yeah?

spec/factories/attached_files.rb
```diff
@@ -10,4 +10,10 @@
   trait :with_file do
     file { fixture_file_upload(Rails.root.join('spec/fixtures/files/sul.svg'), 'image/svg+xml') }
   end
+
+  trait :with_preserved_file do
+    file { fixture_file_upload(Rails.root.join('spec/fixtures/files/sul.svg'), 'image/svg+xml') }
+
+    after(:build, &:transform_blob_to_preservation)
+  end
```
mjgiarlo, line 18: If you want to build on the `with_file` factory above, you can also consider:

```suggestion
  trait :with_preserved_file do
	with_file
    after(:build, &:transform_blob_to_preservation)
  end
```

app/services/cocina_generator/structural/generator.rb
lwrubel: I wasn't familiar with that iteration approach, thank you! It is more concise. Why is it that you have to add `.to_h` on an object that is already a hash for it to work?

app/services/cocina_generator/structural/generator.rb
lwrubel: Right, `contains` should be an array. I had been getting some errors about not being able to use `.each` on a cocina object, and that seemed to address it. But if I remove the `to_a` now, I'm no longer getting that error. So perhaps it was something to do an earlier version I have been working on. At any rate, I am taking this out and will test.

app/services/cocina_generator/structural/generator.rb
lwrubel: Yes, exactly. I had a comment higher up about this and will put it here too for clarity.

spec/factories/attached_files.rb
lwrubel: Oh, nice! Will do that.

app/services/cocina_generator/structural/generator.rb
mjgiarlo: This is not obvious at all: when you send `#to_h` to a hash, it returns itself (which seems pointless)... but if a block is also passed, the return value of the block for each key/value pair will be used as the new key/value pair.

So, you can think of `hash.to_h { |key, value| [key, value] }` as a short form of `hash.map { |key, value| [key, value] }.to_h`

app/services/cocina_generator/structural/generator.rb
mjgiarlo: OK, sounds good!

app/services/cocina_generator/structural/generator.rb
```diff
@@ -24,38 +26,63 @@ def generate
...
-          preserved_file_names = Set.new(work_version.preserved_files.map(&:filename).map(&:to_s))
-          preserved_file_names.superset?(existing_file_names)
+      def create_filesmap(preserved_file_names)
+        files_map = {}
+        cocina_obj.structural.contains.to_a.each do |fileset|
+          # h2 only has one file per fileset
+          existing_file_name = fileset.structural.contains.first.filename
+          files_map[existing_file_name] = fileset if preserved_file_names.include?(existing_file_name)
         end
+        files_map
```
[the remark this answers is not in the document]
lwrubel: Thanks, taking this change, too (and no longer need to make an array of the cocina).

app/services/cocina_generator/structural/file_generator.rb
lwrubel: I've made some tweaks to the `resource_id` and requiring both, as you suggested. 

I see what you're saying about there being two situations here to handle. Things already feel like they're getting pretty abstract, so I'm not sure about diving into that at this moment. But open to it, especially if this gets any more complicated.

app/services/cocina_generator/structural/file_generator.rb
```diff
@@ -33,22 +38,24 @@ def request_file_attributes
           filename:,
           access:,
           administrative:,
-          hasMimeType: blob.content_type,
+          hasMimeType: cocina_file&.hasMimeType || blob.content_type,
```
justinlittman, line 41: Don't you want `hasMimeType` and `size` to be in the other direction?

justinlittman: Actually, also for all of these attributes.

justinlittman (changes requested):
Logic question.

app/services/cocina_generator/structural/file_generator.rb
justinlittman: I suspect we're all misunderstanding because this is so confusing.

What if you are replacing an existing file? Then there will be cocina AND a blob.

lwrubel: I tested by replacing a file with one with the same name but different size / md5. The new size and md5 are reflected in the updated structural cocina.

spec/jobs/deposit_job_spec.rb
```diff
@@ -181,9 +215,94 @@
...
+          has_mime_type2 = params[:metadata].structural.contains.second.structural.contains.first.hasMimeType
+          expect(external_identifier2).to eq('9999999')
+          expect(size2).to eq(17_675)
+          expect(has_mime_type2).to eq('image/svg+xml')
+        end
+        expect(SdrClient::Deposit::UploadFiles).to have_received(:upload)
+      end
+    end
+
+    context 'when a file with the same name is replaced on a deposit' do
```
lwrubel, line 267: @justinlittman I added this test to try to check the scenario of replacing a file with one with the same name. Would you take a look?

app/services/cocina_generator/structural/file_generator.rb
lwrubel: To correct what I said above, there would be a blob on preserved_files, but the blob has a service_name of "preservation" and does not point to an actual file. So we want to use the cocina anyway on those files.

justinlittman (approved):

PR #2890, opened 2022-12-07
Adds support for unzipping.

### LINKED ISSUE #2843: Add new option for single ZIP file upload

justinlittman:
closes #2843

It's easier for users.

CI, tested locally by J-Litt, tested in QA by Amy


app/components/works/file_row_component.html.erb
```diff
@@ -28,21 +28,25 @@
   </div>
 
   <div class="upload-description col-4">
-      <div>
-        <%= form.label :label, 'Description (optional)' %>
-        <%= render PopoverComponent.new key: 'work.description' %>
-      </div>
-      <%= form.text_field :label, class: 'form-control' %>
+      <% unless zip_template? %>
```
justinlittman, line 31: Since zips are transient, they don't have labels.

justinlittman:
This is ready for dev review. I'll need to tweak some wording per Amy before merging.

app/components/works/add_files_component.html.erb
```diff
@@ -13,7 +13,7 @@
                                   bs_toggle: 'modal',
                                   bs_target: '#contactUsModal' }) %>
       </p>
-        <div data-controller="dropzone" data-dropzone-max-file-size="10000" data-dropzone-max-files="1000">
+      <div data-controller="dropzone" data-dropzone-max-file-size="10000" data-dropzone-previews-container=".dropzone-files-previews">
```
justinlittman, line 16: Here and below, dropzone controller is modified to support having multiple instances on the same page.

app/components/works/file_row_component.html.erb
```diff
@@ -28,21 +28,26 @@
...
-    <%= form.hidden_field :_destroy %>
-    <%= form.hidden_field :file %>
-    <%= form.check_box :hide, class: 'form-check-input' %>
-    <%= form.label :hide, 'Hide file', class: 'form-check-label col-form-label ' %>
-    <%= render PopoverComponent.new key: 'work.hide_file' %>
-    </div>
+      <div class="form-check">
+        <%= form.hidden_field :_destroy %>
+        <%= form.hidden_field :file %>
+        <%= form.hidden_field :path %>
```
justinlittman, line 44: Files now have "paths".

app/components/works/state_display_component.rb
```diff
@@ -11,7 +11,7 @@ def initialize(work_version:)
 
     def call
       value = I18n.t(work_version.state, scope: 'work.state')
-      return value unless work_version.depositing? || work_version.reserving_purl?
+      return value unless work_version.wait_state?
```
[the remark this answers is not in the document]
mjgiarlo: Dig it.

justinlittman:
1. User selects zip upload and adds a zip file.
image

2. User saves as draft or deposits.
3. H2 unzips the zip and attaches the individual files. The contents of the zip are now attached files just like files that had been individually uploaded.

app/jobs/unzip_job.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+require 'zip'
+
+# Change work version from a zipfile type to a browser type by unzipping the attached zip file.
+class UnzipJob < BaseDepositJob
+  queue_as :default
+
+  def perform(work_version)
+    zip_attached_file = work_version.attached_files.last
```
mjgiarlo, line 10: I assume we can reliably expect the last AF to be the zip?

app/jobs/unzip_job.rb
```diff
@@ -0,0 +1,53 @@
...
+    destroy_attached_files_except(zip_attached_file, work_version)
```
mjgiarlo, line 11: I could see making a case for this method living on the `WorkVersion` but I don't feel strongly about that.

justinlittman:
Also:
image

app/jobs/unzip_job.rb
mjgiarlo: (i.e., I could also see leaving it right here.)

app/jobs/unzip_job.rb
```diff
@@ -0,0 +1,53 @@
...
+    work_version.upload_type = 'browser'
+    work_version.unzip_complete!
+  end
+
+  private
+
+  def destroy_attached_files_except(except_attached_file, work_version)
+    work_version.attached_files.each do |attached_file|
+      work_version.attached_files.destroy(attached_file) unless attached_file == except_attached_file
+    end
```
mjgiarlo, line 31: There's likely a more database-/AR-optimized way to do this but we're invoking this in a job so 🤷🏻‍♂️ ?

app/jobs/unzip_job.rb
mjgiarlo, line 10: There are lots of related DB operations in `#perform`. Will things get in a weird state without a transaction in place?

app/jobs/unzip_job.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  def ignore?(entry)
+    !entry.file? || entry.name.start_with?('__MACOSX') || entry.name.end_with?('.DS_Store')
+  end
+
+  def filepath_for(attached_file)
+    ActiveStorage::Blob.service.path_for(attached_file.blob.key)
+  end
```
mjgiarlo, line 40: This could also be a method worth considering for the `AttachedFile` model so it's responsible for knowing its own file path.

app/models/work_version.rb
```diff
@@ -172,4 +172,14 @@ def staged_files
...
+
+  # current state might take some time, so may want to display a spinner
+  def wait_state?
+    depositing? || reserving_purl? || unzipping_state?
+  end
+
+  # in one of the unzipping states
+  def unzipping_state?
+    unzip_first_draft? || unzip_version_draft? || unzip_pending_approval? || unzip_depositing?
+  end
```
mjgiarlo, line 184: Nice.

app/services/cocina_generator/structural/file_generator.rb
```diff
@@ -34,7 +34,7 @@ def request_file_attributes
         {
           type: Cocina::Models::ObjectType.file,
           version: work_version.version,
-          label: attached_file.label,
+          label: attached_file.label || '',
```
mjgiarlo, line 37: I see at least a couple of `attached_file.label || ''` in this PR. Might it be worth tossing something like this at the `AttachedFile` model (like what we're now doing for paths above)?

```ruby
def label
  super || ''
end
```

❓

spec/jobs/deposit_job_spec.rb
```diff
@@ -25,9 +25,7 @@
     allow(SdrClient::Login).to receive(:run).and_return(Success())
     allow(SdrClient::Connection).to receive(:new).and_return(conn)
     allow(Honeybadger).to receive(:notify)
-    # rubocop:disable RSpec/MessageChain
-    allow(attached_file).to receive_message_chain(:file, :blob).and_return(blob)
-    # rubocop:enable RSpec/MessageChain
+    allow(attached_file).to receive_message_chain(:file, :blob).and_return(blob) # rubocop:disable RSpec/MessageChain
```
mjgiarlo, line 28: Thank you for this.

app/jobs/unzip_job.rb
justinlittman: Yes, at least with current implementation.

justinlittman: Possibly, but since unzipping might take several minutes I'm reluctant to proactively transactionalize.

app/jobs/unzip_job.rb
justinlittman: When I was writing this code I shared that feeling, but couldn't quite put my finger on the code.

mjgiarlo: How about something like...

```ruby
work_version.attached_files.excluding(except_attached_file).destroy_all
```

app/jobs/unzip_job.rb
justinlittman: I just went to do this. However, AttachedFile is a bit of a confusing mess because it might have different kinds of Blobs. (Some of the existing methods only work with certain kinds of Blobs, which is part of the problem.) This method will only work for ActiveStorage Blobs.

So I'm inclined to leave this closer to its context then add to AttachedFile where it can generate confusion.

mjgiarlo: Sounds good. Thanks for looking into it.

app/services/cocina_generator/structural/file_generator.rb
justinlittman: Better yet, I'll make it the DB default.

mjgiarlo: Good call!

mjgiarlo: Path too?

justinlittman: Path needs a value else we get it from the attached file. So I don't think we want to default to ''.

app/jobs/unzip_job.rb
justinlittman: Why couldn't I figure that out?

mjgiarlo:
@justinlittman Can you include this in the PR? https://github.com/sul-dlss/happy-heron/wiki/State-machine-diagrams (Or add it in a follow-on?)

justinlittman:
@mjgiarlo Updated state machine diagrams.

app/jobs/unzip_job.rb
mjgiarlo: It took a bit of digging once I convinced myself it should be possible, FWIW!

mjgiarlo (approved):

PR #2901, opened 2022-12-12
Generate filesets sorted by their filenames

mjgiarlo:
**Hold until confirmed tested** by @andrewjbtw 

# Why was this change made? 🤔

Connects to sul-dlss/sul-embed#1380

This commit supports work to render folder/directory path information in filenames within sul-embed. Currently, sul-embed respects the order of contentMetadata in public XML (which is generated from from Cocina). Meanwhile, files for a given work in H2 are ordered randomly, based partly on 1) the order in which they were uploaded, and 2) whether the files are staged or preserved. This commit establishes a consistent order for files when generating Cocina for an H2 work, and that order is ascending alpha based on filenames, meaning sul-embed can continue to respect the given order.

CI

### DIFF
--- a/app/services/cocina_generator/structural/generator.rb
+++ b/app/services/cocina_generator/structural/generator.rb
@@ -26,36 +26,30 @@ module CocinaGenerator
       # 1) Start with the existing filesets from SDR
       # 2) Rebuild filesets that have attached_files that are preserved
       # 3) Add new filesets that have attached_files in the local store.
+      # 4) Sort the filesets by filename (each fileset has a single file in H2 design) ascending and case-insensitive
       def build_filesets
-        filesets = find_preserved_filesets.to_h do |filename, fileset|
-          [
-            filename,
-            build_fileset(attached_file: get_preserved_file(filename), fileset:)
-          ]
-        end
-        filesets.values + work_version.staged_files.map { |af| build_fileset(attached_file: af) }
+        (work_version.staged_files.map { |af| build_fileset(attached_file: af) } + find_preserved_filesets)
+          .sort_by { |file_set_hash| file_set_hash.dig(:structural, :contains, 0).filename.downcase }
       end
 
       # @return [Hash] map of filename to matching fileset's cocina for preserved files.
       def find_preserved_filesets
-        return {} unless cocina_obj
+        return [] unless cocina_obj
 
-        preserved_file_names = work_version.preserved_files.map(&:filename).map(&:to_s)
-        create_filesmap(preserved_file_names)
+        # h2 only has one file per fileset
+        cocina_obj
+          .structural
+          .contains
+          .filter_map do |fileset|
+          next unless (preserved_file = get_preserved_file(fileset.structural.contains.first.filename))
+
+          build_fileset(attached_file: preserved_file, fileset:)
+        end
       end
 
       def get_preserved_file(filename)
         work_version.preserved_files.find { |attached_file| attached_file.filename.to_s == filename }
       end
-
-      # rubocop:disable Rails/IndexBy
-      def create_filesmap(preserved_file_names)
-        # h2 only has one file per fileset
-        cocina_obj.structural.contains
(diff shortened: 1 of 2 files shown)


app/services/cocina_generator/structural/generator.rb
```diff
@@ -27,36 +27,34 @@ def generate
       # 2) Rebuild filesets that have attached_files that are preserved
       # 3) Add new filesets that have attached_files in the local store.
       def build_filesets
-        filesets = find_preserved_filesets.to_h do |filename, fileset|
```
mjgiarlo, line 30: I noticed that we were generating a hash in `find_preserved_filesets` and then immediately calling `.to_h` on that again. Now we call `.to_h` on it once down within `find_preserved_filesets`. This change combines two bits of code.

app/services/cocina_generator/structural/generator.rb
```diff
@@ -27,36 +27,34 @@ def generate
...
+        (work_version.staged_files.map { |af| build_fileset(attached_file: af) } + find_preserved_filesets.values)
+          .sort_by { |file_set_hash| file_set_hash.dig(:structural, :contains, 0).filename }
```
mjgiarlo, line 31: Here's the new logic that sorts all filesets by filename.

app/services/cocina_generator/structural/generator.rb
```diff
@@ -27,36 +27,34 @@ def generate
...
 
-      # @return [Hash] map of filename to matching fileset's cocina for preserved files.
-      def find_preserved_filesets
-        return {} unless cocina_obj
-
-        preserved_file_names = work_version.preserved_files.map(&:filename).map(&:to_s)
-        create_filesmap(preserved_file_names)
+      def preserved_file_names
+        work_version.preserved_files.map(&:filename).map(&:to_s)
       end
```
mjgiarlo, line 52: Used to be a local variable, extracted to private instance method.

app/services/cocina_generator/structural/generator.rb
```diff
@@ -27,36 +27,34 @@ def generate
...
 
-      # rubocop:disable Rails/IndexBy
-      def create_filesmap(preserved_file_names)
-        # h2 only has one file per fileset
-        cocina_obj.structural.contains
-                  .select { |fileset| preserved_file_names.include?(fileset.structural.contains.first.filename) }
-                  .to_h { |fileset| [fileset.structural.contains.first.filename, fileset] }
-      end
-      # rubocop:enable Rails/IndexBy
```
mjgiarlo, line 58: I inlined this above since it was only used in one spot and added more clarity to what `find_preserved_filesets` was/is doing.

spec/services/cocina_generator/dro_generator_spec.rb
```diff
@@ -395,6 +403,27 @@
...
+                      hasMimeType: 'image/png',
+                      label: 'MyString',
+                      size: 4_045,
+                      type: Cocina::Models::ObjectType.file,
+                      version: 1
+                    }
+                  ] },
+                  type: Cocina::Models::FileSetType.file,
+                  version: 1
+                },
```
mjgiarlo, line 426: This test fails when the `sort_by` is removed in the structural generator. This shows that the order of filesets is based not on the order of uploads but on the filename.

app/services/cocina_generator/structural/generator.rb
justinlittman: If we're just calling `.values`, it is necessary to create a hash?

justinlittman (approved):

app/services/cocina_generator/structural/generator.rb
mjgiarlo: @justinlittman Good eye. That led me to a little more refactoring which I think brings added clarity. 🍻

justinlittman: Much appreciated. That area of the code is very difficult to grok.

PR #2903, opened 2022-12-12
allow globus to be put in test mode for localhost work

peetucket:

Follow on from #2887 

Allow us to test the globus workflow in localhost by simulating the response from the globus API.  This is useful as we may to verify the workflow works as expected by simulating the test H2 user not having a globus account, and then later having a globus account, so we can verify H2 behaves as expected.  You configure this in your `settings.local.yml`  as explained in the updated README.  It will automatically be ignored if running in production regardless of setting.

Localhost

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -117,7 +117,7 @@ class WorksController < ObjectsController
 
     authorize! work_version, to: :show?
 
-    if GlobusClient.user_exists?(work.depositor.sunetid)
+    if globus_user_exists?(work.depositor.email)
       GlobusSetupJob.perform_later(work_version)
       flash[:notice] = I18n.t('work.flash.globus_setup_complete')
     else
@@ -151,6 +151,12 @@ class WorksController < ObjectsController
   end
 
   private
+
+  def globus_user_exists?(user_id)
+    return Settings.globus.test_user_exists if Settings.globus.test_mode && Rails.env.development?
+
+    GlobusClient.user_exists?(user_id)
+  end
 
   # Create the next WorkVersion for this work
   def create_new_version(previous_version)
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -13,6 +13,8 @@ globus:
   uploads_directory: /uploads/
   transfer_endpoint_id: endpoint_uuid
   help_doc_url: https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk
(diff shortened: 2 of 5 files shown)


app/controllers/works_controller.rb
```diff
@@ -117,7 +117,7 @@ def complete_globus_setup
 
     authorize! work_version, to: :show?
 
-    if GlobusClient.user_exists?(work.depositor.sunetid)
+    if globus_user_exists?(work.depositor.email)
       GlobusSetupJob.perform_later(work_version)
```
peetucket, line 121: note: i noticed this is actually a small bug in the recent implementation ... we actually need to send the email address to the globus client and not just the sunet

mjgiarlo: Good catch. That was one of the breaking changes between 0.4.0 and 0.5.0.

app/controllers/works_controller.rb
```diff
@@ -152,6 +152,12 @@ def edit_button
 
   private
 
+  def globus_user_exists?(user_id)
+    return Settings.globus.test_user_exists if Settings.globus.test_mode && !Rails.env.production?
```
mjgiarlo, line 156: Do you think we will want/need this ability when running tests? If not, I wonder whether we'd want `Rails.env.development?` or `!Rails.env.production?` here.

peetucket: no i guess we won't... i can do the switch to development ...  and then get rid of the test bits that force the state of this setting too

mjgiarlo: OK, awesome!

config/settings.yml
```diff
@@ -13,6 +13,8 @@ globus:
   uploads_directory: /uploads/
   transfer_endpoint_id: endpoint_uuid
   help_doc_url: https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk
+  test_mode: false # for testing purposes in non-production only, simulates globus API calls
+  test_user_exists: true # if test_mode=true, simulates if the globus user exists
```
mjgiarlo, line 17: Should the default value be `false`?

peetucket: the default value for `test_mode` absolutely should be false, the second value is going to toggle back and forth depending on the workflow you want to test (and is irrelevant unless `test_mode = true` anyway)

mjgiarlo: OK, sounds good. :)

mjgiarlo (approved):

PR #2909, opened 2022-12-13
add hb checkins for cron

### LINKED ISSUE #2869: use Honeybadger checkins to ensure crons are running
See https://docs.honeybadger.io/guides/check-ins/

This approach would be a fine way to ensure our cron jobs are running. Note similar work in prescat: PR https://github.com/sul-dlss/preservation_catalog/pull/2075

peetucket:

Fixes #2869 -- add HB checkins for cron jobs.  Modeled on sul-dlss/preservation_catalog#2075

Config update for production in sul-dlss/shared_configs#1876

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,16 @@ sudo /usr/bin/systemctl status sneakers
 
 This is started automatically during a deploy via capistrano
 
+## Cron check-ins
+
+Some cron jobs (configured via the whenever gem) are integrated with Honeybadger check-ins. These cron jobs will check-in with HB (via a curl request to an HB endpoint) whenever run. If a cron job does not check-in as expected, HB will alert.
+
+Cron check-ins are configured in the following locations:
+1. `config/schedule.rb`: This specifies which cron jobs check-in and what setting keys to use for the checkin key. See this file for more details.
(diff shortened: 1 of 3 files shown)


mjgiarlo:
@peetucket I'm not entirely sure this is what I'm seeing, since I'm new to HB check-ins, but looking at https://app.honeybadger.io/projects/77112/check_ins it appears that H2's check-ins may not be reporting in, and it looks like only prod has been wired up in shared_configs (no stage or QA) unlike with our prescat HB setup. (I'm poking around here because I picked up the ticket to add check-ins to DSA.)

peetucket:
I believe the reason we have seen no check-ins yet for H2 is because this code was merged to main after we split the production branch off (currently running in prod).  That branch has no check-ins yet: https://github.com/sul-dlss/happy-heron/blob/production/config/schedule.rb

I tested one of the check-ins manually by curling from the H2 prod server and HB did pick it up so I think it should work when we deploy main of H2 to prod.

We can discuss if we want H2 stage and QA to check-in as well.  My recollection is that we didn't think it was that useful, but I could be wrong.

mjgiarlo:
@peetucket OK, thank you!

PR #2913, opened 2022-12-14
Add support for globus deposit.

### LINKED ISSUE #2848: User can initiate deposit after done adding files to Globus
See also #2865 for state tracking 

When the user is done uploading to globus, they will return to the H2 work edit page (for works that have selected the "upload via globus option")

Note: the deposit button on the work edit form will only be enabled if we've been able to create them a folder in globus from their first time they clicked "Save As Draft" AND they have checked the "I've finished uploading my files to Globus" checkbox. 

After clicking save as draft and after the endpoint is created, they will see the total number of files and total file size of all of the files in the globus endpoint they were told to use

When clicking deposit (note user must check the "I've finished uploading my files to Globus" checkbox to enable deposit)

- Changes the permission on the globus endpoint to prevent them from writing any more files to it
- Starts the deposit 

When clicking save as draft:

- nothing new happens if they don't check the box that says "I've finished uploading my files to Globus",
- when checking the checkbox that says "I've finished uploading my files to Globus", we will fetch the files from Globus, attach to the object, and then switch the upload type back to "browser" so the user can edit the files on the next draft edit

Summary: 

1. User select Globus.
2. User saves draft. (Deposit is not an option.)
3. User uploads via Globus.
4. User returns to draft. Draft indicates the number of files found on Globus.
5. User indicates that completed with Globus upload (via a checkbox).
6. User selects Save as Draft or Deposit.
7. H2 retrieves file info from Globus and creates the necessary DB objects. (This is a wait state for the user with a spinner, like unzipping.)
8. If user selected Deposit, H2 performs deposit after previous step is completed.
9. If user selected Save Draft, user can return to draft and edit, remove, enhance, etc. files before deposit.

UI workflow: https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/470591811_Globus_Workflow

### LINKED ISSUE #2896: Make removing existing attached files for globus async
Currently, all existing attached files for a work version are destroyed as part of form save when the user selects globus: https://github.com/sul-dlss/happy-heron/blob/main/app/forms/draft_work_form.rb#L174

However, this involves deleting all of the files from ActiveStorage disk, which is slow. For a large number of attached files, this results in an unacceptable delay.

Instead, this should be performed asynchronously. One option would be to invoke a job as part of the globus state transitions.

justinlittman:
closes #2848
closes #2896
closes #2961


lwrubel (approved):
This is pretty large, so would understand if you'd like to wait to have another person to review it before merging.

justinlittman:
This is likely to need continued refinement, so I'm good with merging now. Thanks for the review!

PR #2918, opened 2022-12-14
Add a banner to globus deposits on item show page

### LINKED ISSUE #2915: Add a banner to globus deposits on item show page
In the same way we show a banner on globus deposits on item show pages for users who need to setup their account (see https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471028472), we want to show a similar banner for users who have a globus account but still need to complete uploading files (globus item in draft states).  The banner would have some instructional text and a link to help instead of the "complete setup" button.

Text to come

peetucket:

Fixes #2915 - add a banner to the top of the item show/detail page for globus deposits in a draft state directing users to the help page and showing them the folder name we created on globus.  

This is in the same spot that the banner shows for users who still need to complete their globus account setup (i.e. that banner goes away when they complete their setup and is replaced by this one).

**Account setup complete** (new banner)

Screen Shot 2022-12-14 at 2 55 33 PM

**Account setup needed** (existing banner)

Screen Shot 2022-12-14 at 2 59 22 PM

~~HOLD for language updates by Amy~~

Localhost

### DIFF
--- a/app/components/works/globus_setup_component.html.erb
+++ b/app/components/works/globus_setup_component.html.erb
@@ -10,8 +10,8 @@
 <% else %>
    <div class="row my-5 mx-2 alert alert-info">
       <div class="col-12">
-         Your files should now be uploaded to the Globus folder "<%=work_version.globus_endpoint%>" we created for this deposit.
-         Refer to our <%= link_to 'help page', Settings.globus.help_doc_url %> if you have questions.</p>
+         <p>Upload your files to the Globus folder "<%=work_version.globus_endpoint%>" and then click the "Edit or Deposit" button on this page to complete your deposit.
+         Refer to our <%= link_to 'help page', Settings.globus.help_doc_url %> if you have questions, or contact us at the Help link above.</p>
       </div>
    </div>
 <% end %>
--- a/app/components/works/globus_setup_component.html.erb
+++ b/app/components/works/globus_setup_component.html.erb
@@ -1,8 +1,17 @@
-<div class="row my-5 mx-2 alert alert-danger">
-  <div class="col-9">
-     Click this button when you have completed setting up your Globus account and we will email you instructions on where to upload your files.
-  </div>
-  <div class="col-3">
-     <%= render Works::GlobusSetupButtonComponent.new(work_version:) %>
-  </div>
-</div>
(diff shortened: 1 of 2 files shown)


peetucket:
@amyehodge  this adds the banner for draft globus deposits that shows when users have completed their account setup ... i just made up some language for now and linked to the help page ... we can tweak as needed

amyehodge:
@peetucket I'd like to tweak this a bit to read:

"Upload your files to the Globus folder "auser/work9/version1" and then click the Edit button on this page to complete your deposit. Refer to our help page[linked] if you have questions, or contact us at the Help link above."

peetucket:
Do we want to say "click the Deposit button..." instead of "click the Edit button..." ?

amyehodge:
If this message is on the work view/show page, the button actually says "Edit or Deposit" at this point. Perhaps it would be best to say that for clarity, even if it is longer.

app/components/works/globus_setup_component.html.erb
```diff
@@ -1,8 +1,17 @@
...
+         Click this button when you have completed setting up your Globus account and we will email you instructions on where to upload your files.
+      </div>
+      <div class="col-3">
+         <%= render Works::GlobusSetupButtonComponent.new(work_version:) %>
+      </div>
+   </div>
+<% else %>
+   <div class="row my-5 mx-2 alert alert-info">
+      <div class="col-12">
+         <p>Upload your files to the Globus folder "<%=work_version.globus_endpoint%>" and then click the "Edit or Deposit" button on this page to complete your deposit.
```
mjgiarlo, line 13: When Globus emails the user, and they click the link, do they see the `/uploads/` part of the endpoint? If so, we may want to include that here. If not, I think this is 💯 🆗

peetucket: I do not believe the user sees the `/uploads/` part of the path, but that part of the path should also not be stored in the database and thus should not be shown here either.  See https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/globus_setup_job.rb#L36-L47 for where the path is created and stored in the database

mjgiarlo: Good call!

mjgiarlo (approved):

PR #2922, opened 2022-12-16
Load work files table with turbo

### LINKED ISSUE #2910: Scroll long lists of files on item view page
See mockups 1 and 2 on https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471212700_long_List_Item_View

User sees 15 lines (folder and file names) in the file section, and the rest can be seen when scrolling through that section.

If this takes a long time to load, load the rest of the page first and present a spinner/wait image over the file section so user knows it's still in progress but they can see the rest of the page while they wait.

File list view:

Wait state:

Analogous to the #2920 (which is the same thing on the item edit page)

lwrubel:
Resolves #2910 to put files in a scrollable container and lazy load with turbo.

Unit. 

### DIFF
--- a/app/assets/stylesheets/work.scss
+++ b/app/assets/stylesheets/work.scss
@@ -53,6 +53,18 @@
     }
   }
 
+  // For files table
+  .files-list {
+    max-height: 550px;
+    overflow-y: auto;
+  }
+
+  .files-list thead th {
+    position: sticky;
+    top: 0;
+    z-index: 1;
+  }
+
   // The pencil icon
   .edit {
     margin-left: 10px;
--- a/app/components/works/show/files_component.html.erb
+++ b/app/components/works/show/files_component.html.erb
@@ -1,22 +1,14 @@
-<table class="table table-sm mb-5" id="filesTable">
-  <thead class="table-light">
-  <tr>
-    <th scope="col" class="table-title">Files
-      <span class="ms-3">
-        <%= helpers.turbo_frame_tag dom_id(work, 'edit_file'), src: edit_button_work_path(work, tag: 'file'), target: '_top' %>
-      </span>
-      <% if download_all? %>
-        <%= link_to work_zip_path(work), class: 'ms-4', data: { turbo: false } do %>
-          <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download All Files</span>
-        <% end %>
-      <% end %>
-    </th>
-    <th scope="col">Description</th>
-    <th scope="col">Hide file</th>
-  </tr>
-  </thead>
-  <tbody>
-    <%= render Works::Show::AttachedFileComponent.with_collection(attached_files.sort_by { |attached_file| attached_file.path.downcase },
-                                                                  work_version: work_version) %>
-  </tbody>
-</table>
+<turbo-frame id="<%= dom_id(work, 'files_list') %>" src="<%= files_list_work_path(work) %>" loading="lazy">
+  <table class="table table-sm mb-5" id="filesTable">
+    <thead class="table-light">
+    <tr>
+      <th scope="col" class="table-title">Files</th>
+      <th scope="col">Description</th>
+      <th scope="col">Hide file</th>
+    </tr>
+    </thead>
+    <tbody>
+      <tr><td colspan="3" class="text-center"><%= spinner %></td></tr>
+    </tbody>
+  </table>
+</turbo-frame>
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -126,6 +126,16 @@ class WorksController < ObjectsController
     redirect_to work_path(work)
   end
 
+  def files_list
+    work = Work.find(params[:id])
+    work_version = work.head
+    attached_files = work_version.attached_files.sort_by { |attached_file| attached_file.path.downcase }
+
+    authorize! work_version, to: :show?
+
+    render partial: 'works/files_list', locals: { work:, work_version:, attached_files: }
+  end
+
   # We render this button lazily because it requires doing a query to see if the user has access.
   # The access can vary depending on the user and the state of the work.
   def delete_button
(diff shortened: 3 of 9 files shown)


app/components/works/show/files_component.html.erb
```diff
@@ -1,19 +1,21 @@
...
-    <%= render Works::Show::AttachedFileComponent.with_collection(attached_files, work_version: work_version) %>
-  </tbody>
-</table>
+<turbo-frame id="<%= dom_id(work, 'files_list') %>" src="<%= files_list_work_path(work) %>" loading="lazy">
+  <table class="table table-sm mb-5" id="filesTable">
+    <thead class="table-light">
+    <tr>
+      <th scope="col" class="table-title">Files
+        <span class="ms-3">
+          <%= helpers.turbo_frame_tag dom_id(work, 'edit_file'), src: edit_button_work_path(work, tag: 'file'), target: '_top' %>
```
lwrubel, line 7: It seemed OK to keep these buttons available and operable, since they should work while the file list loading is running.

spec/components/works/show/attached_file_component_spec.rb
```diff
@@ -3,13 +3,14 @@
...
+  let(:attached_file) { build(:attached_file, :with_file) }
 
-  # it "renders something useful" do
-  #   expect(
-  #     render_inline(described_class.new(attr: "value")) { "Hello, components!" }.css("p").to_html
-  #   ).to include(
-  #     "Hello, components!"
-  #   )
-  # end
+  context 'with an attached file' do
```
lwrubel, line 10: Moved spec from `file_component_spec.rb` to here since file_component.rb is not initially displaying the attached file info anymore.

ndushay (approved):
marking approved but suspect someone else who hasn't had their head in prescat for this whole work cycle shoul review.

lwrubel:
@amyehodge here are screenshots of how this is implemented. Let me know if you'd like to see it live and I can deploy. 

Loading with spinner: 
Screen Shot 2022-12-16 at 3 49 53 PM (2)

Scrollable files table
Screen Shot 2022-12-16 at 3 50 01 PM

amyehodge:
That looks great @lwrubel ! I would kinda like to see it in action, but maybe Monday? If you aren't around, perhaps after the break. Thanks!

app/assets/stylesheets/work.scss
```diff
@@ -53,6 +53,18 @@
...
+  .files-list {
+    max-height: 550px;
+    overflow-y: scroll;
+  }
+
+  .files-list thead th {
+    position: sticky;
+    top: 0;
+    z-index: 1;
+  }
```
lwrubel, line 66: To keep header fixed and not scrolling.

spec/features/create_new_work_spec.rb
```diff
@@ -163,6 +168,11 @@
           expect(page).to have_content 'Everyone'
           expect(page).to have_content 'CC0-1.0'
 
+          within('#filesTable') do
```
lwrubel, line 171: Moved test to feature spec since the table is now loaded via a turbo frame.

app/controllers/works_controller.rb
```diff
@@ -244,7 +254,7 @@ def work_params
                      :abstract, :citation_auto, :citation, :default_citation,
                      :access, :license, :version_description,
                      :release, 'embargo_date(1i)', 'embargo_date(2i)', 'embargo_date(3i)',
-                     :agree_to_terms, :assign_doi, :upload_type, :globus, :fetch_globus_files,
+                     :agree_to_terms, :assign_doi, :upload_type, :globus, :fetch_globus_files, :attached_files_list,
```
justinlittman, line 257: Why is this getting added?

app/components/works/show/files_component.html.erb
```diff
@@ -1,22 +1,23 @@
...
-      <% if download_all? %>
-        <%= link_to work_zip_path(work), class: 'ms-4', data: { turbo: false } do %>
-          <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download All Files</span>
+<turbo-frame id="<%= dom_id(work, 'files_list') %>" src="<%= files_list_work_path(work) %>" loading="lazy">
+  <table class="table table-sm mb-5" id="filesTable">
+    <thead class="table-light">
+    <tr>
+      <th scope="col" class="table-title">Files
+        <span class="ms-3">
+          <%= helpers.turbo_frame_tag dom_id(work, 'edit_file'), src: edit_button_work_path(work, tag: 'file'), target: '_top' %>
```
justinlittman, line 7: I'd be inclined to omit edit and download all until the frame is fully rendered.

app/views/works/_files_list.html.erb
```diff
@@ -0,0 +1,24 @@
...
+        <%= link_to work_zip_path(work), class: 'ms-4', data: { turbo: false } do %>
+          <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download All Files</span>
+        <% end %>
+      </th>
+      <th scope="col">Description</th>
+      <th scope="col">Hide file</th>
+    </tr>
+    </thead>
+    <tbody>
+      <%= render Works::Show::AttachedFileComponent.with_collection(attached_files.sort_by { |attached_file| attached_file.path.downcase },
```
justinlittman, line 19: Can these be sorted in the controller?

lwrubel: Yes. Moved the sorting to `works_controller.rb`.

app/components/works/show/files_component.html.erb
lwrubel: I removed the icons. I'm seeing it does cause the "Description" and "Hide?" headers to shift once the icons appear.

app/controllers/works_controller.rb
lwrubel: Argh, left over from a previous approach. Thanks for catching.

justinlittman (approved):

PR #2925, opened 2022-12-19
Order files by filename

### LINKED ISSUE #2907: Sort files in H2 to match sorting in sul-embed
Priority is to match case-insensitive sorting on the item view page.
Also desired but lower priority is to match the sorting on the new work form and edit work form.

lwrubel:
Resolves #2907 to sort files case-insensitive. 

Unit and deployed to QA.

### DIFF
--- a/app/components/works/show/files_component.html.erb
+++ b/app/components/works/show/files_component.html.erb
@@ -14,6 +14,7 @@
   </tr>
   </thead>
   <tbody>
-    <%= render Works::Show::AttachedFileComponent.with_collection(attached_files, work_version: work_version) %>
+    <%= render Works::Show::AttachedFileComponent.with_collection(attached_files.sort_by { |attached_file| attached_file.path.downcase },
+                                                                  work_version: work_version) %>
   </tbody>
 </table>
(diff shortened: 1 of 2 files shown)


app/components/works/show/files_component.html.erb
```diff
@@ -14,6 +14,7 @@
   </tr>
   </thead>
   <tbody>
-    <%= render Works::Show::AttachedFileComponent.with_collection(attached_files, work_version: work_version) %>
+    <%= render Works::Show::AttachedFileComponent.with_collection(attached_files.sort_by { |af| af.path.to_s.downcase },
```
justinlittman, line 17: 1. `.path` is already a string, so `.to_s` isn't necessary.
2. Slight preference for `attached_file` instead of `af`.

justinlittman (changes requested):

app/components/works/show/files_component.html.erb
lwrubel: Got it. Have made both of those changes.

justinlittman (approved):

Issue #2929, opened 2023-01-05
H2 does not recognize that I have Globus account set up (on stage)


amyehodge:
When I click on the button on the view page to indicate that I'm done setting up Globus, it tells me that I haven't set up an account yet, even though I was actively logged into Globus at the time and definitely have an account.

I’m now stuck and unable to finish the deposit. We need this to work for user testing.

lwrubel:
This is likely the issue: the access token from Globus includes `"expires_in":172800,` which is a 48-hour expiration. The token is created in the [globus_client initializer](https://github.com/sul-dlss/happy-heron/blob/6ebb79bfd23b10d341216e5dfdebaa11b332a709/config/initializers/globus_client.rb#L3).

lwrubel:
See #2937 for work to request a new access token.

lwrubel:
See #2937 for work to request a new access token.

lwrubel:
Closing since the work moved to #2937.

PR #2936, opened 2023-01-06
[HOLD] Catch Globus token error and raise

lwrubel:
First step in resolving #2929 to trap error first raised in Globus interaction if there is a problem with looking up a user's identity. This is temporary to gather info until we complete work to refresh the token. (See #2937, for requesting a new access token as needed.)

HOLD since we think #2937 will resolve this.

Unit


lwrubel:
Given #2937, do we even want to do this temporary change?

mjgiarlo (approved):

mjgiarlo:
@lwrubel I could go either way! Approving for now.

lwrubel:
Closing this since the globus client gem was updated and will handle requesting a new token upon these errors.

PR #2938, opened 2023-01-06
[HOLD] Displays info on uploaded globus files.

### LINKED ISSUE #2912: Display number and total size of all files in globus folder on item edit page
For a globus deposit in the draft state (either first_draft or version_draft), when the user is on the item edit form, we should display the total number of files and the total size of all of the files in their globus endpoint.  This allows the user to have a visual clue that they have correctly placed files in their globus folder and they when they click deposit, the expected number of files will be deposited.

see https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005611 
see sul-dlss/globus_client#34

Note: we expect this may cause performance issues as is dependent on the globus API.  It may also need to be async.

justinlittman:
closes #2912

Allows users to verify Globus upload.

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -82,6 +82,15 @@
         <%= form.label :fetch_globus_files,
          "All files have been uploaded to Globus.",
          class: 'form-check-label' %>
+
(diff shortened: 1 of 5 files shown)


justinlittman:
Per discussion with @amyehodge this is on hold while we try to determine if users are encountering a problem with Globus uploads that this solves (or if they are encountering a problem, what would be the best solution).

PR #2939, opened 2023-01-06
Ignore MacOS-specific bogus file paths in Globus uploads

### LINKED ISSUE #2935: Globus upload should not include .DS_Store files
Globus file upload option on QA picked up a .DS_Store file that we don't want it to pick up. 
Particularly odd is that even when I display hidden files on my machine, I don't see this file, so I'm not sure where it's coming from.

mjgiarlo:
# Why was this change made? 🤔

Fixes #2935

This commit mirrors logic already applied to zip uploads to make sure bogus MacOS files are not included in H2 items.

CI + tested successfully by Amy in QA

### DIFF
--- a/app/jobs/fetch_globus_job.rb
+++ b/app/jobs/fetch_globus_job.rb
@@ -11,6 +11,8 @@ class FetchGlobusJob < BaseDepositJob
     work_version.attached_files.destroy_all
 
     filepaths_for(work_version).each do |path|
+      next if ignore?(path)
+
       work_version.attached_files << new_attached_file(path, work_version)
     end
     work_version.upload_type = 'browser'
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

app/jobs/fetch_globus_job.rb
```diff
@@ -26,6 +28,10 @@ def globus_prefix(work_version)
     "#{Settings.globus.uploads_directory}#{work_version.globus_endpoint}/"
   end
 
+  def ignore?(path)
+    path.start_with?('__MACOSX') || path.end_with?('.DS_Store')
```
edsu, line 32: I was a bit confused why you wouldn't want to ignore` __MACOSX` anywhere in the path.

mjgiarlo: @edsu Good question! I'm not sure where to expect `__MACOSX` subdirectories, so I opted for parity w/ what the `UnzipJob` is currently doing: https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/unzip_job.rb#L33

PR #2945, opened 2023-01-09
Show file hierarchy on the work show page

### LINKED ISSUE #2943: User can view hierarchical file list in work show page
Currently the work show page shows a flat list of files, even if the filenames include paths (indicating there is a hierarchy).  

For example, it currently looks like this:

Screen Shot 2023-01-06 at 2 30 55 PM

We should instead show the files in a hierarchical view as shown in the mockup:

https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005875

It may be helpful to look at the similar work done in Argo: https://github.com/sul-dlss/argo/pull/3913

peetucket:

Fixes #2943 - show file hierarchy on the work show page, as mocked up here: https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005875

Borrowing from the Argo display: https://github.com/sul-dlss/argo/pull/3913/files

Questions:

- what does the empty vs filled in folder icon mean in the mockups? note: we will never have a folder with no files in it in H2 (and how to render an empty folder?)

- Localhost
- Existing tests
- Updated/added tests

### DIFF
--- a/app/components/works/show/attached_file_component.html.erb
+++ b/app/components/works/show/attached_file_component.html.erb
@@ -1,5 +1,5 @@
 <tr>
-    <td class="file-level" style="--index: <%=index+1%>rem"><li><%= basename %>
+    <td class="file-level" style="--depth: <%=depth+1%>rem"><li><%= basename %>
         <% if can_download? %>
             <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
                 <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download</span>
--- a/app/components/works/show/attached_file_component.html.erb
+++ b/app/components/works/show/attached_file_component.html.erb
@@ -1,10 +1,13 @@
 <tr>
-    <td class="file-level" style="--depth: <%=depth+1%>rem"><li><%= basename %>
-        <% if can_download? %>
-            <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
-                <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download</span>
+    <td class="file-level" style="--depth: <%= depth+1 %>rem">
+        <li>
+            <%= basename %>
+            <% if can_download? %>
+                <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
+                    <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download</span>
+                <% end %>
             <% end %>
-        <% end %></li>
+        </li>
     </td>
     <td><%= label %></td>
     <td><%= hide? ? 'Yes' : 'No' %></td>
--- a/app/components/works/show/attached_file_component.html.erb
+++ b/app/components/works/show/attached_file_component.html.erb
@@ -1,10 +1,10 @@
 <tr>
-    <td><%= basename %>
+    <td class="file-level" style="--index: <%=index+1%>rem"><li><%= basename %>
         <% if can_download? %>
             <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
                 <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download</span>
             <% end %>
-        <% end %>
+        <% end %></li>
     </td>
     <td><%= label %></td>
     <td><%= hide? ? 'Yes' : 'No' %></td>
--- a/app/components/works/show/attached_file_component.html.erb
+++ b/app/components/works/show/attached_file_component.html.erb
@@ -1,5 +1,5 @@
 <tr>
-    <td><%= filename %>
+    <td><%= basename %>
         <% if can_download? %>
             <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
                 <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download</span>
--- a/app/components/works/show/attached_file_component.html.erb
+++ b/app/components/works/show/attached_file_component.html.erb
@@ -1,13 +1,12 @@
 <tr>
     <td class="file-level" style="--depth: <%= depth+1 %>rem">
-        <li>
-            <%= basename %>
-            <% if can_download? %>
-                <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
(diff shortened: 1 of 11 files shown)


peetucket:
@astridu and @amyehodge ... this is an example of what the file hierarchy display looks like in the work show/detail page with this PR:

Screen Shot 2023-01-10 at 1 33 24 PM

(the repeated filenames are expected, I just made copies of the files and put into folders for testing)

I can deploy to stage or qa if you want to see it live (screenshots are from my laptop).

One note...the mockup (https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005875) shows some empty folders with a different folder icon (white instead of black).  We won't ever had blank folders in our data model, so the current view only ever shows black folder icons (representing folders that either have sub-folders and/or files in them).

spec/services/file_hierarchy_service_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+          FileHierarchyService::File.new(attached_file3),
+          FileHierarchyService::File.new(attached_file4)
+        ], 1)
+      ], [
+        FileHierarchyService::File.new(attached_file5)
+         ], 0)
+    )
+  end
+  # rubocop:enable Layout/FirstArrayElementIndentation
+end
```
peetucket, line 48: disabled this rubocop rule here as it wanted to indent this structure to a ridiculous degree, which it less readable in my view

peetucket:
Just deployed to QA.  Best way to see it in action (if you don't already have an object with hierarchy to view) is to create a ZIP file with nested hierarchy and upload to a new draft object.  You don't need to deposit.

amyehodge:
@peetucket yes, sorry. it's at https://sul-h2-qa.stanford.edu/works/351

amyehodge:
yep @peetucket looks right to me now as well

app/components/works/show/attached_file_component.html.erb
```diff
@@ -1,10 +1,10 @@
 <tr>
-    <td><%= filename %>
+    <td class="file-level" style="--depth: <%=depth+1%>rem"><li><%= basename %>
```
mjgiarlo, line 2: Can you say more about using an `<li>` outside the context of an ordered or unordered list? I know browsers don't choke on it but I'm not sure if it's valid HTML.

mjgiarlo, line 2: Can you add a line break between the `td` element and the `li` that follows?

app/components/works/show/attached_file_component.html.erb
```diff
@@ -1,10 +1,10 @@
...
         <% if can_download? %>
             <%= link_to path, class: 'ms-2', data: { turbo: false } do %>
                 <span class="fa-solid fa-download" aria-hidden="true"></span><span class="visually-hidden">Download</span>
             <% end %>
-        <% end %>
+        <% end %></li>
```
mjgiarlo, line 7: Can you move the `/li` element to the next line?

app/components/works/show/attached_file_component.rb
```diff
@@ -24,10 +24,6 @@ def path
       def can_download?
         !attached_file.in_globus?
```
mjgiarlo, line 25: It's not clear to me why we delegate some bits to `attached_file` but not others. Could remedy this by delegating `:file` and `:in_globus?` (for lines 21 and 25).

app/components/works/show/attached_file_component.html.erb
mjgiarlo, line 2: For consistency (until we merge the erb-lint PR), can you add spaces between `depth+1` and `<%%>` ?

app/components/works/show/directory_component.html.erb
```diff
@@ -0,0 +1,13 @@
...
+    <td class="file-level" style="--depth: <%=depth%>rem" colspan="3">
+       <i class="fa fa-folder" aria-hidden="true"></i> <strong><%= name %></strong>
+    </td>
+  </tr>
+<% end %>
+<% children_files.each do |file| %>
+   <%= render Works::Show::AttachedFileComponent.new(attached_file: file.attached_file, depth:) %>
+<% end %>
+<% children_directories.each do |directory| %>
+   <%= render Works::Show::DirectoryComponent.new(directory: directory) %>
```
mjgiarlo, line 12: It looks like line 9 is using the new keyword/hash shorthand syntax, `depth:)`, and this line is not. Consider `...(directory:)`?

app/components/works/show/attached_file_component.html.erb
peetucket: I was trying to match Astrid's design, which shows a bullet point in front of files.  Perhaps there is a better way to do this other than an `<li>` tag, or perhaps we just need to wrap the whole thing in a `<ul>` tag, though not sure if this is legal since the inner `<li>` elements will be in table rows

mjgiarlo: OK. I wonder if we could build on what we're doing for folders and find a FontAwesome icon that works for files? https://fontawesome.com/ Maybe @astridu and/or @amyehodge could suggest one?

app/components/works/show/attached_file_component.rb
peetucket: Likely oversight.  Delegated both of those to `attached_filed` and updated methods as needed.

app/components/works/show/attached_file_component.html.erb
peetucket: Closest free icon I found was a filled in circle, but at it's smallest, it still looks a bit big: 

Screen Shot 2023-01-12 at 10 41 30 AM

spec/services/file_hierarchy_service_spec.rb
mjgiarlo: WFM

app/components/works/show/attached_file_component.html.erb
mjgiarlo: @peetucket Give https://fontawesome.com/docs/web/style/size a shot.

mjgiarlo: If so, you could try inserting an HTML character so we get the literal bullet without the list semantics: `&bull;`

mjgiarlo: @peetucket OK, super. Thanks for rolling with the punches there.

mjgiarlo (approved):

PR #2947, opened 2023-01-11
Prevent deposit of Globus works without files

### LINKED ISSUE #2942: Prevent user from depositing without uploading any files to Globus
This applies to when the Globus option is checked. Currently the user can check the box to say they have uploaded files and then deposit, even if there aren't any files.

The inclusion of a file has always been a requirement for self-deposit, and file upload is currently required when uploading files directly to H2. 

The only exception to this is the current implementation of Globus in production where they check a box to say they are using Globus and then admins are alerted to that and follow up with them to accession the files manually.

mjgiarlo:
# Why was this change made? 🤔

Fixes #2942

* Shift knowledge of Globus endpoint template and full path to the `WorkVersion`
  * This commit concentrates knowledge about Globus endpoint construction and the full path in the `WorkVersion` model, a more logical home than, e.g., a one-off line in a job. This minor refactoring sets up the following commit to finish #2942.
* Invalidate the draft work form if a Globus upload includes no files
  * This commit uses the existing work form machinery to detect when a user is attempting to deposit a Globus-file-having work without having already uploaded files.
* Surface Globus errors to the user and Honeybadger
  * Without this commit, Globus connection errors would cause 500 errors and interrupt user deposits.
* Regenerate Rubocop TODOs

CI & will test in a deployed env with Amy


app/forms/draft_work_form.rb
```diff
@@ -32,6 +32,8 @@ class DraftWorkForm < Reform::Form
   end)
   property :embargo_date, embargo_date: true, on: :work_version
   property :assign_doi, on: :work, type: Dry::Types['params.nil'] | Dry::Types['params.bool']
+  property :upload_type, on: :work_version
+  property :fetch_globus_files, virtual: true, type: Dry::Types['params.nil'] | Dry::Types['params.bool']
```
mjgiarlo, line 36: Moved from below so that all form properties are co-located together up here.

app/jobs/fetch_globus_job.rb
```diff
@@ -20,12 +20,9 @@ def perform(work_version)
...
   def filepaths_for(work_version)
-    filepaths = GlobusClient.get_filenames(path: work_version.globus_endpoint, user_id: work_version.work.owner.email)
-    filepaths.map { |filepath| filepath.delete_prefix(globus_prefix(work_version)) }
-  end
-
-  def globus_prefix(work_version)
-    "#{Settings.globus.uploads_directory}#{work_version.globus_endpoint}/"
+    GlobusClient
+      .get_filenames(path: work_version.globus_endpoint, user_id: work_version.work.owner.email)
+      .map { |filepath| filepath.delete_prefix(work_version.globus_endpoint_fullpath) }
```
mjgiarlo, line 25: Two changes here:

1. Ditch the local variable since it's already clear that we're dealing with a list of file paths.
2. Replace the `#globus_prefix` method in the job with a call to `#globus_endpoint_fullpath` on the `WorkVersion` instance

app/jobs/globus_setup_job.rb
```diff
@@ -36,7 +36,10 @@ def globus_user_exists?(user_id)
   def create_globus_endpoint(work_version)
     user = work_version.work.owner
     # e.g. 'mjgiarlo/work1234/version1'
-    endpoint_path = "#{user.sunetid}/work#{work_version.work.id}/version#{work_version.version}"
+    endpoint_path = format(WorkVersion::GLOBUS_ENDPOINT_TEMPLATE,
+                           user_id: user.sunetid,
+                           work_id: work_version.work.id,
+                           work_version: work_version.version)
```
mjgiarlo, line 42: As above, moving work version-specific logic out of jobs and into the `WorkVersion` model.

app/forms/draft_work_form.rb
```diff
@@ -79,6 +82,16 @@ def unique_filenames
...
 
+  # Ensure there is more than zero files if fetching from Globus
+  def globus_files_provided
+    return if GlobusClient.get_filenames(path: work_version.globus_endpoint, user_id: work.owner.email).any?
+
+    errors.add(
+      :attached_files,
+      "must include at least one file uploaded to Globus at #{work_version.globus_endpoint_fullpath}"
+    )
+  end
```
mjgiarlo, line 93: Use existing form machinery to validate the new requirement and present errors to the user where they expect to see them.

spec/models/work_version_spec.rb
```diff
@@ -85,6 +85,25 @@
     end
   end
 
+  describe '#globus_endpoint_fullpath' do
+    subject(:work_version) { create(:work_version, work: create(:work)) }
+
+    before do
+      work_version.globus_endpoint = globus_endpoint
```
mjgiarlo, line 92: This can't be patched into line 89 or we get a stack too deep error, as the globus endpoint needs to know the work's ID and the work version's version but that's not known until *after* line 89 fires.

spec/requests/edit_work_spec.rb
```diff
@@ -147,27 +147,30 @@
...
-
-          it 'redirects to the work page and starts fetching globus' do
-            patch "/works/#{work.id}", params: { work: work_params, commit: 'Save as draft' }
-            expect(work.reload.head).to be_fetch_globus_version_draft
-            expect(response).to redirect_to(work)
+        context 'with a Globus upload type' do
+          before do
+            allow(GlobusClient).to receive(:get_filenames)
+              .and_return(["/uploads/#{user.sunetid}/work#{work.id}/version1/file1.txt"])
           end
```
mjgiarlo, line 154: Changes to this test were added when the Globus client operation was added to the draft work form, so that the tests could keep passing.

app/forms/draft_work_form.rb
```diff
@@ -79,6 +82,16 @@ def unique_filenames
     errors.add(:attached_files, 'must all have a unique filename.') unless filenames.size == filenames.uniq.size
   end
 
+  # Ensure there is more than zero files if fetching from Globus
+  def globus_files_provided
```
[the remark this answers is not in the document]
mjgiarlo: @justinlittman I was thinking this was the cleanest way to implement the feature, and if it turns out to be slow in testing or in prod, we can always change it then. A change I am considering making in the meantime which may help: add a new method to the globus client that can do a quick sniff if there any files, which should be faster than navigating a potentially highly recursive set of subdirectories and returning filenames. If you think that is worth doing, maybe we could get this moving and I can immediately follow up with this work?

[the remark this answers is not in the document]
justinlittman: My bigger concern is not performance, but stability. Unless I'm mistaken, if there is a problem with Globus, this code path (and hence H2) stops working.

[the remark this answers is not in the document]
mjgiarlo: @justinlittman If a user says "this Globus deposit is ready" and clicks "Deposit," and there's a problem with Globus, what do you think the behavior should be? cc: @amyehodge 

I guess my main question is whether we know enough to toss in code to handle this in advance or if it's better to wait until this case looks clearer? Like, I can add `rescue`s and Honeybadgers in here willy-nilly but is now the time for that, do you think?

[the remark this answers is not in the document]
mjgiarlo: This would lead to H2 works w/o files being accessioned if there are Globus problems. @amyehodge @andrewjbtw Is this acceptable?

amyehodge:
@mjgiarlo What exactly happens here? Does it look to the user like it's gone through okay -- except that if they check the purl there are no files? Are we alerted somehow if there is something we need to manually fix for the user? I don't want these slipping through the cracks and ending up with deposits without files that we don't know are there.

mjgiarlo:
@amyehodge 💬 
> @mjgiarlo What exactly happens here? Does it look to the user like it's gone through okay -- except that if they check the purl there are no files? 

Yep, that's exactly right.

> Are we alerted somehow if there is something we need to manually fix for the user? I don't want these slipping through the cracks and ending up with deposits without files that we don't know are there.

We'd get a Honeybadger alert. Do you think that's adequate or are those too easy to miss? (I find them pretty easy to miss, personally.) The alternative currently is that if Globus is down, the deposit would blow up on the user (a 500 error), but all their data would still be there and they'd still be able to get back to the work via their dashboard and try again. We can talk about other possible alternatives too, though right now what's missing for me is a sense of how often or how prolonged any Globus outages would be which for me would influence how urgently this change should be prioritized (but you're the P.O., not me, so am happy to defer to you here!).

amyehodge:
@mjgiarlo My sense is that Globus has not been down much. Or at least I have never encountered this problem or heard of anyone encountering this problem. It is perhaps possible to inquire with srcc-support regarding historic down times. I'm happy to do that if it would be useful. 

I dislike 500 errors generally for users. If I knew exactly what this HB error would look like I might be able to identify them easily, but not guaranteed. For example, would the header say [sul-dlss/globus] or something else? Some other notification that stands out more would be nicer. And it's super helpful when those alerts include workids and/or usernames.

andrewjbtw:
I'm not sure I follow the current state vs alternatives. Are the scenarios that if Globus is unavailable then:

1. The deposit would silently pass through with no files
2. The deposit would pass through with no files but HB will be alerted
3. The deposit will not complete, HB will be alerted, the deposit will spin in progress until there's an intervention
4. The deposit will not complete initially, HB will be alerted, the deposit will spin in progress, but if Globus comes back the sidekiq retry will succeed
5. The deposit will not complete, HB will be alerted, user will see some error, item will return to draft
6. Not only will the deposit not complete, H2 will also go down if Globus isn't available

If those are realistic options, then I would prefer 5, where the item goes back to draft and the user can try again. Or if sidekiq would succeed, then 4 could work as well.

As Amy says, the biggest worry is that someone deposits no data and no one notices until later. Depositors run the range of diligently checking until things are done to not checking for months or possibly ever. I'm also concerned that the case where a deposit has no files will result in a "success: your deposit is complete" email even though that's not success.

Given that we are not deleting data from Globus automatically, at least we should notice empty deposits if we always check the purls before cleaning up Globus, and then we would still be able to deposit the data.

mjgiarlo:
@amyehodge @andrewjbtw I tossed another small commit at the branch that implemented scenario 5 above. cc: @justinlittman 

Here's what the user will see when depositing a Globus work that runs into a Globus problem:

Screenshot from 2023-01-11 12-35-38

app/forms/draft_work_form.rb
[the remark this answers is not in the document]
mjgiarlo: @justinlittman I added the `rescue` and HB integration, but left these form submissions invalid.

justinlittman (approved):

justinlittman:
@mjgiarlo approved, but did not merge since marked [HOLD].

Issue #2914, opened 2022-12-14
Users can add files at any level of the hierarchy to an existing work when editing


peetucket:
Follow on from #2893
MUST build on work in PR #2949 (i.e. start there and make a new PR that builds on that one)

Add the ability to add files at any level of the hierarchy, and (possibly replace files).
#2893 and #2949 add the ability to show the hierarchy, along with change descriptions, delete files and show/hide them on the work edit page.

As speced, drop zone will need to appear in different parts of the page.  May need to investigate if possible.

Simplification from design: do not implement 'replace' icon functionality, as this can be accomplished by delete, followed by add at the same level.

See https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005613 for the UI

The icons shown are:
- ![Screen Shot 2022-12-07 at 1.04.29 PM.png](https://images.zenhubusercontent.com/59b9a35fb0222d5de477a1fb/4c4a056b-307f-47df-bdb2-aa05e774d667): add a file in that folder; opens the file upload panel

Ignore the replace icon in the UI - it is not needed.

PR #2949, opened 2023-01-12
Edit existing files in hierarchy in work edit page

### LINKED ISSUE #2893: Users can edit lists of files that have been uploaded
Regardless of version 1 or later, or how files were uploaded to an object (globus, browser, zip expansion), users should be able to edit the list of files in an object in their existing hierarchy.  The current H2 UI allows users to delete a file, add a description or hide a file, or add a new file.  This new interface would replace the existing interface and allow for the same behaviors, in addition to replacing an existing file and adding files within existing folders.

Simplification: use checkboxes for hide/show changes, as this mimics current functionality and is easier to implement.

See https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005613 for the UI

For this ticket, implement:
- display of hierarchy
- show/hide
- edit description
- delete

Follow on ticket to implement add files/replace

The icons to implement as shown are are:
- : deletes the file

Ignore the hide icon and the replace icon.  We will checkboxes to show/hide as in the current design.  See https://stanfordlib.slack.com/archives/C69UZCJ8M/p1673986693423669 for discussion.

For some background/history, see #2852 which was simplified down to this ticket.

peetucket:

Fixes #2893 - users can edit files in work edit page, showing hierarchy

Proposed design change uses check boxes for hiding/showing instead of icon and using file icon instead of bullet point (matches current design):

Screen Shot 2023-01-18 at 12 50 58 PM

Note: adding a new file to an object only works at the top level of the hierarchy and shows the preview *below* the table in this PR, both of which would need to be fixed in the follow on ticket #2914

Localhost, updated tests


app/assets/stylesheets/details.scss
```diff
@@ -12,3 +12,8 @@
 td.file-level {
   padding-left: var(--depth);
 }
+
+/* zebra-stripe files table on edit page */
+tr.attached-files:nth-child(even) {
+  background-color: #dadada;
```
mjgiarlo, line 18: [Watch the video](https://youtu.be/LnBSc28Obs4)

app/components/works/add_files_component.rb
```diff
@@ -9,6 +9,10 @@ def initialize(form:)
 
     attr_reader :form
 
+    def has_attached_files?
+      form.object.attached_files.size.positive?
```
mjgiarlo, line 13: Could also consider `form.object.attached_files.any?` here, I think?

app/assets/stylesheets/details.scss
peetucket: is this 😎 or WTF? or both

mjgiarlo: Oh, it's a link to the music video for the song, "Da Da Da"

app/components/works/add_files_component.html.erb
```diff
@@ -24,18 +27,18 @@
           </div>
           <span style="margin: .7rem">or</span> <button type="button" class="dz-clickable btn btn-outline-primary">Choose files</button>
           <div class="invalid-feedback" data-dropzone-target="feedback">You must attach a file<%= error_message %></div>
-          <div class="dropzone-previews dropzone-files-previews" data-dropzone-target="previewsContainer">
-            <%= form.fields_for :attached_files do |file_form| %>
-              <%= render Works::FileRowComponent.new(form: file_form) %>
-            <% end %>
-          </div>
         </fieldset>
```
peetucket, line 30: this was the old style of showing a flat list of files ... we will replace this with the new hierarchical components, added to the new `edit` component namespace

app/components/works/edit/attached_file_component.html.erb
```diff
@@ -0,0 +1,22 @@
+<tr class="attached-files <%= 'dz-complete' if uploaded? %>" data-dropzone-target="preview">
+    <%= form.fields_for :attached_files, attached_file do |file_form| %>
```
mjgiarlo, line 2: Can you remind me what you learned about where a `<form>` element can live in a table? I assume/believe you tried having the form outside the `<tr>` and that didn't work?

FWIW, I don't think it matters much as long 1. browsers render it fine, and 2. it doesn't confuse screen readers.

app/components/works/edit/attached_file_component.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    <td class="file-level" style="--depth: <%= depth+1 %>rem" data-dropzone-target="fileName">
+        <span class="fa-regular fa-file" aria-hidden="true"></span> <%= basename %>
```
mjgiarlo, line 4: The filename currently looks like it's not vertically centered (like the hide checkbox) unlike the other elements in the row. Worth tossing that CSS around here?

app/components/works/edit/attached_file_component.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    </td>
+    <td>
+        <%= file_form.text_field :label, class: 'form-control' %>
+    </td>
+    <td>
+        <%= file_form.check_box :hide, style: 'align: center' %>
```
mjgiarlo, line 10: The hide checkbox currently looks like it's not vertically centered (like the filename) unlike the other elements in the row. Worth tossing some CSS around here?

app/components/works/edit/attached_file_component.html.erb
```diff
@@ -0,0 +1,22 @@
...
+    </td>
+    <td>
+        <%= button_tag type: 'button',
+                    class: "dz-remove pull-right btn remove-file",
+                    "aria-label": "Remove file",
+                    data: { action: "click->dropzone#removeAssociation" } do %>
+        <span class="fa-regular fa-trash-alt"></span>
+        <% end %>
+        <%= file_form.hidden_field :_destroy %>
+    <% end %>
```
mjgiarlo, line 20: I expected this `end` statement to be just inside the `</tr>` rather than the `</td>`. Typo? Intentional?

app/components/works/edit/attached_file_component.html.erb
```diff
@@ -0,0 +1,22 @@
+<tr class="attached-files <%= 'dz-complete' if uploaded? %>" data-dropzone-target="preview">
+    <%= form.fields_for :attached_files, attached_file do |file_form| %>
+    <td class="file-level" style="--depth: <%= depth+1 %>rem" data-dropzone-target="fileName">
```
mjgiarlo, line 3: Not a blocking concern, but a curiosity: might we be able to do the `+1` elsewhere? (like in the Ruby side of this component?)

app/components/works/edit/file_row_component.rb
```diff
@@ -0,0 +1,35 @@
...
+        @form = form
+        @zip_template = zip_template
+      end
+
+      attr_reader :form
+
+      def filename
+        return unless uploaded?
+
+        form.object.model.path
```
mjgiarlo, line 17: Lines are uncovered. Consider tossing a component test at this or is it more trouble than it's worth?

app/components/works/edit/attached_file_component.html.erb
peetucket: Yup, will align vertically.

app/components/works/edit/attached_file_component.html.erb
peetucket: Oddly enough, this was the fix to get the form rendering correctly within the table, so it's intentional and required.  I will add a comment about this.

I found a hint in this stack overflow question: https://stackoverflow.com/questions/6602579/fields-for-in-table-produces-technically-incorrect-html

app/components/works/edit/attached_file_component.html.erb
peetucket: What I learned is that the location of the <%end%> tag determines where a hidden input tag is added by rails, and if it's added in the wrong spot, the browser gets confused.  The current code works and produces valid HTML even though the nesting of the <%end%> looks wrong.

app/components/works/edit/attached_file_component.html.erb
mjgiarlo: Huh!

app/components/works/edit/attached_file_component.html.erb
peetucket: happy to move that `+1` to the component initialization

app/components/works/edit/file_row_component.rb
peetucket: good find - this actually appears to be a dead code path -- removing from here and the view

app/components/works/add_files_component.html.erb
```diff
@@ -64,7 +68,7 @@
 
         <template data-dropzone-target='template'>
           <%= form.fields_for :attached_files, AttachedFile.new, child_index: 'TEMPLATE_RECORD' do |file_form| %>
-            <%= render Works::FileRowComponent.new(form: file_form, zip_template: true) %>
+            <%= render Works::Edit::FileRowComponent.new(form: file_form, zip_template: true) %>
           <% end %>
```
peetucket, line 72: this component just got moved to the `Edit` namespace (and was modified slightly as well)

app/components/works/edit/directory_component.html.erb
```diff
@@ -0,0 +1,13 @@
...
+       <i class="fa fa-folder" aria-hidden="true"></i> <strong><%= name %></strong>
+    </td>
+  </tr>
+<% end %>
+<% children_files.each do |file| %>
+   <%= render Works::Edit::AttachedFileComponent.new(attached_file: file.attached_file, depth:, form:) %>
+<% end %>
+<% children_directories.each do |directory| %>
+   <%= render Works::Edit::DirectoryComponent.new(directory:, form:) %>
+<% end %>
```
mjgiarlo, line 13: Maybe worth considering for follow-on work: rendering these components using ViewComponents collection rendering capability (vs. Ruby iteration): https://viewcomponent.org/guide/collections.html

I don't think this is blocking in the slightest.

app/components/works/edit/file_row_component.rb
```diff
@@ -0,0 +1,29 @@
...
+      attr_reader :form
+
+      # file has been uploaded but may not yet have been saved to the database model (and may not validate)
+      def uploaded?
+        !form.object.try(:model)&.blob.nil?
+      end
+
+      # file has been uploaded and saved to the model
+      def saved?
+        form.object.persisted?
```
mjgiarlo, line 21: This may be untested 🤷🏻‍♂️

mjgiarlo (approved):

PR #2950, opened 2023-01-12
Display Globus message in deposit form only if a Globus type upload

### LINKED ISSUE #2932: Text above Save/Deposit button on form regarding Globus should only display when Globus option is checked
I am seeing this text even though I have selected the zip file upload option. It should not be there unless the user has selected the Globus option.

mjgiarlo:

Fixes #2932

This commit restores code that I suspect was inadvertently removed.

CI + local browser

### DIFF
--- a/app/javascript/controllers/deposit_button_controller.js
+++ b/app/javascript/controllers/deposit_button_controller.js
@@ -1,13 +1,14 @@
 import { Controller } from "stimulus"
 
 export default class extends Controller {
-  static targets = ["depositButton", "globusRadioButton", "globusCheckbox", ]
+  static targets = ["depositButton", "globusRadioButton", "globusCheckbox", "globusMessage", ]
 
   connect() {
     this.updateDepositButtonStatus()
   }
 
   updateDepositButtonStatus(_event) {
+    this.globusMessageTarget.hidden = !this.globusRadioButtonTarget.checked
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/deposit_button_controller.js
```diff
@@ -1,13 +1,14 @@
...
 export default class extends Controller {
-  static targets = ["depositButton", "globusRadioButton", "globusCheckbox", ]
+  static targets = ["depositButton", "globusRadioButton", "globusCheckbox", "globusMessage", ]
 
   connect() {
     this.updateDepositButtonStatus()
   }
 
   updateDepositButtonStatus(_event) {
+    this.globusMessageTarget.hidden = !this.globusRadioButtonTarget.checked
```
mjgiarlo, line 11: Removed in https://github.com/sul-dlss/happy-heron/pull/2913/files#diff-33148bcaafd73019c39a8ffb129631d1f60488cbb143812810a9b9f3d21e90e3L12 

@justinlittman Can you confirm this was inadvertent?

aaron-collier (approved):

app/javascript/controllers/deposit_button_controller.js
mjgiarlo: I'll assume so for now. Holler if not!

Issue #2957, opened 2023-01-17
Update file upload options and Globus instructions on deposit form for 


amyehodge:
**1. Update file upload options**
Update text of all options under **Add your files** to provide more clarity to users. They should read as follows:

- **Upload one or more files (less than 10GB) to be displayed as a flat list of files.**
- **Upload a single zip file (less than 10GB) which will be unzipped and displayed in its original order, including any file hierarchy**
   Viewers will see the file hierarchy and be able to download individual files.
- **Upload one or more files (more than 10GB) which will be displayed as uploaded, including any file hierarchy.**
   Viewers will see the file hierarchy and be able to download individual files.

**2. Update Globus instructions**
In addition, when the third option for Globus is selected, this section should expand so that users see the instruction steps below. Other than the links in the text to parts of the instructions, this section has no interactivity.

![Screen Shot 2023-01-17 at 3.35.48 PM.png](https://images.zenhubusercontent.com/5e9e06e81c23039800e23efc/ade418fc-d29a-4b60-ae66-81be35355c17)

See Step 1 at https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471628807_Globus_Integ_Enh

In case you want to copy and paste, here's all the text for the above:
STEP 1
Scroll to the bottom of this page and click "Save as draft." 
You can finish filling out the form now or later.

STEP 2
Set up a Stanford Globus account, following these [instructions|https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit?pli=1#heading=h.ptdws4k1wnlz]. 
Have a Stanford Globus account? Skip to Step 3.

STEP 3
Complete  Globus setup (may require software installation) following these [instructions|https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit?pli=1#heading=h.3h6eldy9ndya].

STEP 4
Return to your item at sdr.stanford.edu and click the button that says "Globus setup complete."
Don't see this button? Proceed to Step 5.

STEP 5
Check your email from "Globus Notifications."
Click the link in the email and follow these [instructions|https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit?pli=1#heading=h.f2rvtrjiw01] to complete your upload.

STEP 6
[...]

PR #2960, opened 2023-01-18
Update user message for globus setup complete button.

### LINKED ISSUE #2953: Update text of banner on view page for new Globus users
Existing banner text should be replaced with the following:

**You have initiated a Globus-mediated deposit. Here are your next steps. If you need help, visit the [help page|https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit?pli=1#heading=h.s00zxoulxsai].**
1. Click this button when you have completed setting up your Globus account and installing any necessary software.
2. Check for an email from "Globus Notifications" with the link to the location for uploading your files.
3. Follow these [instructions|https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit?pli=1#heading=h.f2rvtrjiw01] to transfer your files.

Banner should be in pink box with button on the right side. I think this is the same as now but is hard for me to confirm.

Step 4 at https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471628807_Globus_Integ_Enh

lwrubel:

Resolves #2953 to add instructions to globus setup complete button message. 

Unit.

Here's what this looks like, @amyehodge. I can deploy to QA/stage if you like. (Sorry, this was an admin user so that box is in the images):

BEFORE
Screen Shot 2023-01-18 at 8 59 56 AM

AFTER
Screen Shot 2023-01-18 at 9 27 39 AM

### DIFF
--- a/app/components/works/globus_setup_component.html.erb
+++ b/app/components/works/globus_setup_component.html.erb
@@ -1,9 +1,16 @@
 <% if work_version.globus_setup_draft? %>
    <div class="row my-5 mx-2 alert alert-danger">
       <div class="col-9">
-         Click this button when you have completed setting up your Globus account and we will email you instructions on where to upload your files.
+         <p><strong>You have initiated a Globus-mediated deposit. Here are your next steps. 
+            If you need help, visit the <%= link_to 'help page', Settings.globus.help_doc_url, target: '_blank' %>.</strong>
+         </p>
+         <ul>
+           <li>Click this button when you have completed setting up your Globus account and installing any necessary software.</li>
+           <li>Check for an email from "Globus Notification" with the link to the location for uploading your files.</li>
+           <li>Follow these <%= link_to 'instructions', "#{Settings.globus.help_doc_url}/edit?pli=1#heading=h.f2rvtrjiw01", target: '_blank' %> to transfer your files.</li>        
+         </ul>
       </div>
-      <div class="col-3">
+      <div class="col-3 d-flex align-items-center">
          <%= render Works::GlobusSetupButtonComponent.new(work_version:) %>
       </div>
    </div>


app/components/works/globus_setup_component.html.erb
```diff
@@ -1,9 +1,16 @@
...
    <div class="row my-5 mx-2 alert alert-danger">
       <div class="col-9">
-         Click this button when you have completed setting up your Globus account and we will email you instructions on where to upload your files.
+         <p><strong>You have initiated a Globus-mediated deposit. Here are your next steps. 
+            If you need help, visit the <%= link_to 'help page', Settings.globus.help_doc_url, target: '_blank' %>.</strong>
+         </p>
+         <ul>
+           <li>Click this button when you have completed setting up your Globus account and installing any necessary software.</li>
+           <li>Check for an email from "Globus Notification" with the link to the location for uploading your files.</li>
+           <li>Follow these <%= link_to 'instructions', "#{Settings.globus.help_doc_url}/edit?pli=1#heading=h.f2rvtrjiw01", target: '_blank' %> to transfer your files.</li>
```
lwrubel, line 10: This link goes to a specific heading in the doc. If we were to change the `help_doc_url` and this specific heading link were no longer valid, then Google just takes you to the top of the doc. An alternative would be to add settings for each heading, but that seems like overkill.

lwrubel:
@amyehodge Regarding the headers comment, I was just pointing out to whoever does code review that inserting the links to Google Doc headers is a point that could break if anything changes with the headers. But I know you're aware of that (saw your comment on #2957). I was wondering if we would want to add more easily-updated settings for each header link, but that has tradeoffs too. Anyway, if for some reason, the header does change/is deleted, the user still gets to the instructions, at least. 

I'm assuming the idea is that these instructions are going to change frequently, so having them on a library website page or in H2 itself makes that problematic to maintain.

mjgiarlo (approved):

PR #2961, opened 2023-01-18
Separate detail message for globus wait states

### LINKED ISSUE #2959: Update message when fetching files from Globus
This could take some time, and users weren't clear what they could do here in the meantime. They also had a hard time seeing this, so we'd like to make it larger and a different color.

When fetching files from Globus, message on item view page should read:

Transferring your files from Globus. This could take some time depending on file size. You may leave this page or close this window and return later.

lwrubel:

Resolves #2959 to show a different message, without a spinner, on the detail page when waiting for Globus files to transfer. 

Screen Shot 2023-01-18 at 3 33 58 PM

This will also wrap, @amyehodge was OK with that (see conversation in #2959). 

Unit, added spec. 

### DIFF
--- a/app/components/works/detail_component.html.erb
+++ b/app/components/works/detail_component.html.erb
@@ -14,10 +14,10 @@
     </div>
   <% end %>
 
-  <header class="title clearfix"><%# clearfix here prevents the EditButtonCompnent from spilling outside the container if the title is long %>
+  <header class="title clearfix"><%# clearfix here prevents the EditButtonComponent from spilling outside the container if the title is long %>
     <span class="header-text"><%= title %></span>
     <%= helpers.turbo_frame_tag dom_id(work, :edit), src: edit_button_work_path(work), target: '_top' %>
-    <span class="state"><%= render Works::StateDisplayComponent.new(work_version: work_version) %></span>
+    <span class="state"><%= render Works::DetailStateDisplayComponent.new(work_version: work_version) %></span>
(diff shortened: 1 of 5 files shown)


app/components/works/detail_component.html.erb
```diff
@@ -14,10 +14,10 @@
     </div>
   <% end %>
 
-  <header class="title clearfix"><%# clearfix here prevents the EditButtonCompnent from spilling outside the container if the title is long %>
+  <header class="title clearfix"><%# clearfix here prevents the EditButtonComponent from spilling outside the container if the title is long %>
     <span class="header-text"><%= title %></span>
     <%= helpers.turbo_frame_tag dom_id(work, :edit), src: edit_button_work_path(work), target: '_top' %>
-    <span class="state"><%= render Works::StateDisplayComponent.new(work_version: work_version) %></span>
+    <span class="state"><%= render Works::DetailStateDisplayComponent.new(work_version: work_version) %></span>
```
lwrubel, line 20: `Works::StateDisplayComponent` is also used on the dashboard and elsewhere, so it can't simply be changed to show this different message for globus wait states.

amyehodge:
@astridu Was your intent that the spinner be taken away?

astridu:
@amyehodge Yes, remove the spinner. The message is sufficient.

lwrubel:
@amyehodge Does this look good to go? I'll merge with your OK.

amyehodge:
@lwrubel Yep. Looks good!

PR #2977, opened 2023-01-23
Support adding files to hierarchy

justinlittman:
Allow users to work with items that have file hierarchies.

Unit

### DIFF
--- a/app/components/works/edit/attached_file_component.rb
+++ b/app/components/works/edit/attached_file_component.rb
@@ -18,6 +18,11 @@ module Works
       def uploaded?
         !attached_file.blob.nil?
       end
+
+      # Directory path for the file (excluding the file name)
+      def path
+        attached_file.paths.join('/')
+      end
     end
   end
 end
--- a/app/components/works/edit/directory_component.rb
+++ b/app/components/works/edit/directory_component.rb
@@ -9,7 +9,11 @@ module Works
         @form = form
       end
 
-      delegate :name, :children_directories, :children_files, :depth, to: :directory
+      delegate :name, :children_directories, :children_files, :depth, :full_path, to: :directory
+
+      def dir_id
+        "dir-#{directory.index}"
+      end
 
       attr_reader :directory, :form
     end
--- a/app/components/works/edit/files_component.html.erb
+++ b/app/components/works/edit/files_component.html.erb
@@ -2,7 +2,7 @@
     <thead class="table-light">
       <tr>
         <th width="50%">Filename</th>
-        <th>Label</th>
+        <th>Description</th>
         <th width="10%">Hide <%= render PopoverComponent.new key: 'work.hide_file' %></th>
         <th>Actions</th>
       </tr>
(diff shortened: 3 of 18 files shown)


peetucket:
Some early thoughts:

1. Looks like we need a rebase now (my PR was merged into main).
2. I wonder if we should remove the labels for the description and the hide checkbox since when it is rendered in the new list view, it is inline with the existing column that already says "hide"? See below:
Screen Shot 2023-01-23 at 4 10 23 PM
3. If we do 2, we may need to split the component since the ZIP dropzone would still need those labels.

app/components/works/edit/files_component.html.erb
```diff
@@ -0,0 +1,13 @@
+<table class="table table-sm" id="filesTable">
+    <thead class="table-light">
+      <tr>
+        <th width="50%">Filename</th>
+        <th>Label</th>
+        <th width="10%">Hide <%= render PopoverComponent.new key: 'work.hide_file' %></th>
```
peetucket, line 6: I think this may need to be "Description" to match what is in the item show page (and the designs for that show page): https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005875

justinlittman:
Uploading Screen Recording 2023-01-24 at 4.02.46 PM.mov…

justinlittman:
https://user-images.githubusercontent.com/588335/214419391-77ed6997-088e-445d-a803-9ffdf2f81ada.mov

app/components/works/edit/directory_component.rb
```diff
@@ -9,7 +9,11 @@ def initialize(directory:, form:)
         @form = form
       end
 
-      delegate :name, :children_directories, :children_files, :depth, to: :directory
+      delegate :name, :children_directories, :children_files, :depth, :full_path, to: :directory
+
+      def dir_id
+        "dir-#{directory.index}"
+      end
```
peetucket, line 17: any ideas why it claims that line is untested?  it seems like any render of the work edit page (like the new cypress test) should it?

app/components/works/edit/attached_file_component.rb
```diff
@@ -18,6 +18,10 @@ def initialize(attached_file:, depth:, form:)
       def uploaded?
         !attached_file.blob.nil?
       end
+
+      def path
+        attached_file.paths.join('/')
+      end
     end
```
peetucket, line 25: this name make sense here (given it is joining the paths), though we may want to consider dropping a comment indicating this the path to the file (excluding the filename)....confusion may arise since unfortunately there is a `path` method on `attached_file` which returns a string with the full path including filename

the response could be to change the somewhat poorly named `path` method in `attached_file` too, but that will touch a bunch of other code, so may be besst done in a future refactor

peetucket (approved):
Approved, with a few comments for consideration (assuming we also have thumbs up from Amy in testing).

app/components/works/edit/attached_file_component.rb
justinlittman: Totally agree with adding a comment. `attached file` `path` is actually a db field name so rename is non-trivial.

PR #2984, opened 2023-01-27
Adds globus integration test support.

justinlittman:
refs https://github.com/sul-dlss/infrastructure-integration-test/issues/534

To allow an integration test to bypass a user adding files to globus.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -15,6 +15,8 @@ globus:
   help_doc_url: https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk
   test_mode: false # for testing purposes in non-production only, simulates globus API calls
   test_user_exists: true # if test_mode=true, simulates if the globus user exists
+  integration_mode: false # set to true for integration test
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -15,6 +15,8 @@ globus:
   help_doc_url: https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk
   test_mode: false # for testing purposes in non-production only, simulates globus API calls
   test_user_exists: true # if test_mode=true, simulates if the globus user exists
+  integration_mode: false # set to true for integration test
```
lwrubel, line 18: How does this work--would we have H2 QA and stage always have `integration_mode` set to true? Seems like then `#make_dir` would never get to call `GlobusClient.mkdir()` so that would not work for using globus in QA/stage. So I think I'm not understanding something.

justinlittman: I suspect what you're missing is this:
```
if work_version.title.ends_with?('Integration Test') && Settings.globus.integration_mode
```

lwrubel (approved):

PR #3001, opened 2023-02-03
Test the recent double-depositing hotfix

### LINKED ISSUE #3000: Refactor work version state machine to be more clear on deposit transitions
This bug #2459 revealed that a DepositJob is incorrectly run twice due to how the transitions were setup.  This bug fix PR fixes the bug: https://github.com/sul-dlss/happy-heron/pull/2999/files but leaves the code in a somewhat suboptimal and potentially confusing state for a future engineer.

Consider:
- improving/adding states as needed to make it clear what is happening
- add a test for the situation

mjgiarlo:

Fixes #3000

CI

### DIFF
--- a/app/models/concerns/work_version_state_machine.rb
+++ b/app/models/concerns/work_version_state_machine.rb
@@ -13,10 +13,12 @@ module WorkVersionStateMachine
       before_transition WorkObserver.method(:before_transition)
 
       after_transition WorkObserver.method(:after_transition)
-      # we do not want to fire this WorkObserver event twice, which will occur after when a new object is
-      # registered, first on object registration, and second after the PID is assigned... so only run
-      # this method the first time we enter the depositing state
-      after_transition except_from: :depositing, to: :depositing, do: WorkObserver.method(:after_begin_deposit)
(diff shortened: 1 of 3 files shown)


app/models/concerns/work_version_state_machine.rb
```diff
@@ -13,10 +13,12 @@ module WorkVersionStateMachine
...
-      # we do not want to fire this WorkObserver event twice, which will occur after when a new object is
-      # registered, first on object registration, and second after the PID is assigned... so only run
-      # this method the first time we enter the depositing state
-      after_transition except_from: :depositing, to: :depositing, do: WorkObserver.method(:after_begin_deposit)
+      # NOTE: we do not want to fire this WorkObserver event twice, which will
+      #       occur after when a new object is registered, first on object
+      #       registration, and second after the PID is assigned... so only run
+      #       this method the first time we enter the depositing state for a
+      #       given version
+      after_transition except_from: :depositing, to: :depositing, do: WorkObserver.method(:after_depositing)
```
mjgiarlo, line 21: Renamed the observer method to be more accurate.

peetucket (approved):

PR #3015, opened 2023-02-22
add redirect for old contact form to new help form

### LINKED ISSUE #3014: Redirect legacy /contact links in PURL to new contact us form
Add redirect on to go from an old link to the contact form http://sdr.stanford.edu/contact to the new contact form? The new form is in a modal...and this URL brings it up: https://sdr.stanford.edu/#help

The issue is that we have thousands of druids with the old link in their Use and Reproduction statements. We’ll eventually have to remediate them, but a redirect would at least address the broken links for now.

peetucket:

Fixes #3014 - add a redirect for old contact form URL to new one

Localhost

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -84,6 +84,8 @@ Rails.application.routes.draw do
     { controller: 'welcome', action: 'show', anchor: 'help' }
   end
   # legacy /contact links in PURLs should redirect to new contact form URL
+  # see https://jirasul.stanford.edu/jira/browse/PURL-1206
+  # e.g. https://purl.stanford.edu/gp739cs2671 which has the old URL encoded in the rightsMetadata XML
   get 'contact', to: redirect('/#help')
 
   resource :help, only: %i[new create]
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -83,6 +83,9 @@ Rails.application.routes.draw do
   direct :contact_form do
     { controller: 'welcome', action: 'show', anchor: 'help' }
   end
+  # legacy /contact links in PURLs should redirect to new contact form URL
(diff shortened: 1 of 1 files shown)


config/routes.rb
```diff
@@ -83,6 +83,9 @@
   direct :contact_form do
     { controller: 'welcome', action: 'show', anchor: 'help' }
   end
+  # legacy /contact links in PURLs should redirect to new contact form URL
```
jcoyne, line 86: I think it would be helpful to provide an example of where this is "in PURLs".  Then we can check to see if this is still needed at some point in the future.

peetucket: @andrewjbtw do you have a couple examples we can drop in a comment here?  OR could link to the code ticket and/or to a remediation ticket for reference

andrewjbtw: There is a purl-feedback ticket: https://jirasul.stanford.edu/jira/browse/PURL-1206

The most recent conversation started with this purl: https://purl.stanford.edu/gp739cs2671 But it's come up a few times recently, leading me to think about redirects.

jcoyne: Peter so if you can say "encoded in the rightsMetadata XML", that would be helpful.

Excerpt from gp739cs2671
```xml 
<rightsMetadata>
...
<use>
  <human type="useAndReproduction">
Property rights reside with the repository. Literary rights reside with the creators of the documents or their heirs. To obtain permission to access, publish or reproduce this content, please contact the Stanford Digital Repository (http://sdr.stanford.edu/contact).
  </human>
  <license/>
</use>
</rightsMetadata>
```

peetucket: added code comments

ndushay (approved):

PR #3033, opened 2023-03-15
Allow re-upload of deleted files.

### LINKED ISSUE #3004: User should be able to delete a zip file and then reupload the same file
For the zip file upload option that expands the hierarchy, I am unable to upload a zip file, delete that file, and then upload the same file again. It tells me it's a duplicate file. The user should be allowed to do this.

The video clip below shows me initially 
- dragging the file in
- deleting it
- dragging it in again and getting the error and deleting it
- and lastly adding it with the chooser and getting the error and deleting it 

I get the error message on the second and third attempts no matter whether I start by dragging the file in or using the chooser and whether the second attempt is via the same method or the other method.

Note that the user CAN still upload a different .zip file. But once a .zip file has been deleted, you can't upload it again until the form is saved as a draft and reloaded.

https://user-images.githubusercontent.com/5558511/217279708-6a5c2eff-4bf7-42c6-ad39-18988801d48e.mov

justinlittman:
closes #3004

Bug

Amy

### DIFF
--- a/app/javascript/controllers/dropzone_controller.js
+++ b/app/javascript/controllers/dropzone_controller.js
@@ -41,6 +41,10 @@ export default class extends Controller {
     const filepaths = fileNameNodes.map(fileNameNode => {      
         const path = fileNameNode.getAttribute('data-dropzone-path')
         const filename = fileNameNode.innerText.trim()
+
+        const item = fileNameNode.closest('.dz-complete')
+        if(item && item.querySelector("input[name*='_destroy']").value == 1) return null
+    
         return path ? `${path}/${filename}` : filename
     })
 


app/javascript/controllers/dropzone_controller.js
```diff
@@ -41,6 +41,10 @@ export default class extends Controller {
     const filepaths = fileNameNodes.map(fileNameNode => {      
         const path = fileNameNode.getAttribute('data-dropzone-path')
         const filename = fileNameNode.innerText.trim()
+
+        const item = fileNameNode.closest('.dz-complete')
+        if(item && item.querySelector("input[name*='_destroy']").value == 1) return null
```
ndushay, line 46: don't we usually see these guard clauses as 

```
return null if condition
``` 

?

ndushay (approved):
approving;  comment for your consideration

app/javascript/controllers/dropzone_controller.js
justinlittman: In Ruby, but not in JS.

ndushay: oh duh.

justinlittman: To be honest, I initially coded it as `return if condition`.

PR #3036, opened 2023-03-21
do not alert HB for a purge

peetucket:

When purging objects that have images, we do not want ActiveStorage to do anything.  Currently we get an HB alert from a raise in the ActiveStorage stack, this just makes it a no-op, which we already have for the `#delete` method.  This `#delete_prefixed` method is also called for images by ActiveStorage and also needs a noop. It is the same change as was done for the globus service a couple months ago: https://github.com/sul-dlss/happy-heron/pull/2969/files

The HB alert: https://app.honeybadger.io/projects/77112/faults/88770594

Existing CI

### DIFF
--- a/app/services/active_storage/service/sdr_service.rb
+++ b/app/services/active_storage/service/sdr_service.rb
@@ -23,6 +23,12 @@ module ActiveStorage
         # to be raised to HB either: https://app.honeybadger.io/projects/77112/faults/88770594
       end
 
+      def delete_prefixed(prefix)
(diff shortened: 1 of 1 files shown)


ndushay (approved):
Approving;  just wondering if there's any need to check with Andrew or Amy (or if you already did)

peetucket:
> Approving; just wondering if there's any need to check with Andrew or Amy (or if you already did)

I have not checked.  Thought process was that it is already skipped in the other cases, so just catches an edge case (an object being purged that has files that also have images).

peetucket:
But I will check just in case.

PR #3047, opened 2023-04-11
T3043 email update

### LINKED ISSUE #3043: Update three email notifications to include link to item view page
This will make it easier for collection managers to get to the new/deposited/modified item. 

Please update these emails to add a link to the view page for the item as described below. I think I got the format right, but please check me!

1.  app/views/collections_mailer/first_draft_created.html.erb

Line 3 should read:
<p>The Depositor <%= @owner.name %> has created a <%=link_to 'draft', work_url(@work)%> in the <%= @collection_version.name %> collection.</p>

2. app/views/collections_mailer/item_deposited.html.erb

Line 3 should read:
<p>The Depositor <%= @owner.name %> has submitted a <%=link_to 'deposit', work_url(@work)%> in the <%= @collection_version.name %> collection.</p>

3.  app/views/collections_mailer/version_draft_created.html.erb

Line 3 should read:
<p>The Depositor <%= @owner.name %> has started a <%=link_to 'new version', work_url(@work)%> in the <%= @collection_version.name %> collection.</p>

amyehodge:

fixes #3043 - add links to the item that triggered the email notification to three existing email notifications

~~No testing~~
- Updated the specs to look for the new URLs
- tested on stage

### DIFF
--- a/app/mailers/collections_mailer.rb
+++ b/app/mailers/collections_mailer.rb
@@ -51,6 +51,7 @@ class CollectionsMailer < ApplicationMailer
     @user = UserPresenter.new(user: params[:user])
     @collection_version = params[:collection_version]
     @owner = params[:owner]
+    @work = params[:work_version].work
 
     mail(to: @user.email, subject: "Draft item created in the #{@collection_version.name} collection")
   end
@@ -77,6 +78,7 @@ class CollectionsMailer < ApplicationMailer
     @user = UserPresenter.new(user: params[:user])
     @collection_version = params[:collection_version]
     @owner = params[:owner]
+    @work = params[:work_version].work
 
     mail(to: @user.email, subject: "Item deposit completed in the #{@collection_version.name} collection")
   end
@@ -85,6 +87,7 @@ class CollectionsMailer < ApplicationMailer
     @user = UserPresenter.new(user: params[:user])
     @collection_version = params[:collection_version]
     @owner = params[:owner]
+    @work = params[:work_version].work
 
     mail(to: @user.email, subject: "New version created in the #{@collection_version.name} collection")
   end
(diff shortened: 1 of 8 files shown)


app/mailers/collections_mailer.rb
```diff
@@ -85,6 +87,7 @@ def version_draft_created
     @user = UserPresenter.new(user: params[:user])
     @collection_version = params[:collection_version]
     @owner = params[:owner]
+    @work = params[:work_version].work
```
peetucket, line 91: we need the work available in the views, the work is available via the work_version from the observer, see https://github.com/sul-dlss/happy-heron/blob/main/app/services/collection_observer.rb#L5-L27

peetucket:
@amyehodge this turned out to be a trickier change since the work object itself was not available in the views by default.  I had to make some code changes to make it available, and then I added the <p> tags back in that got removed (unless that was an intentional change, in which case we can re-remove).

Perhaps best to push this to stage if you want to be sure the emails work as expected, since the specs mock a bunch of stuff.  Might be nice to see them all working for real?

app/services/collection_observer.rb
```diff
@@ -66,9 +66,10 @@ def self.collection_managers_excluding_owner(work_version)
...
 
   def self.mailer_with_owner(user:, work_version:)
-    owner = work_version.work.owner
-    collection = work_version.work.collection
-    CollectionsMailer.with(user:, collection_version: collection.head, owner:)
+    work = work_version.work
+    owner = work.owner
+    collection_version = work.collection.head
+    CollectionsMailer.with(user:, collection_version:, owner:, work:)
   end
```
peetucket, line 73: some small refactoring here, and make sure to pass the `work` along to the mailer, which now needs it

ndushay (approved):
re-approved.  And go for it on the deploys, too.

PR #3053, opened 2023-04-21
allow admin user to edit message shown on home page/dashboard

### LINKED ISSUE #3052: Admin option to add user message to home page/dashboard
There are periodic updates that occur to H2 that we would like to highlight to users when they come to the site.  SDR managers (e.g. Hannah or Amy) would like to be able to easily update/change/hide/show a message to users on both the home page and the dashboard page.  The message would be plain text only (no HTML allowed), would be editable in the admin interface only, with the option to either hide/show it.  The styling would be fixed when shown, with a mockup to be provided by Astrid.  A fixed link out to a different (non-H2) webpage for more information would be added to the user entered message each time (and would not be editable in the admin interface, but would instead be hardcoded and only modified by developers if needed).

Add feature to existing admin dashboard here, called "Edit Recent News Message".   When admin feature selected from drop-down menu, this would present a text box and a hide/show checkbox and a "Save" button.  Presumably this would be stored in the database somewhere.  

Note - the text entered should be brief, on the order of a few sentences.  Suggest enforce a 1000 character limit to avoid the admin user inadvertently entering too much text and causing formatting issues on the page.

When the "show" checkbox is selected, the user entered message will be displayed on both the home page and the top of the dashboard page along with the link/button going to the other webpage for more information.  Location/styling of the message on both pages, and the styling/location of webpage to be provided by Amy, Hannah and Astrid, but work could possibly begin using placeholders.

Screen Shot 2023-04-19 at 11 23 49 AM

Screen Shot 2023-04-19 at 11 26 53 AM

Screen Shot 2023-04-19 at 11 27 11 AM

peetucket:

Fixes #3052 - new admin option to admins can easily edit text and link shown on home page and dashboard page

Waiting on:
- [x] design for how to show on home page
- [x] design for how to show on dashboard page
- [x] ~~link for more info to update in settings~~ made this editable so we don't need to wait for it

Note: the database design will allow for this same basic principle to be used on other pages as well, but for now it's all just hard-coded to use a single page ('home') which is shown on both the dashboard and the home page.  We still have two separate components for display, however, since the design will be different on the home page and the dashboard page.

Design provided by Astrid: https://projects.invisionapp.com/share/ER1387CAQNS5#/screens/473002124?browse

**As implemented in this PR:**

Screen Shot 2023-04-24 at 10 12 18 AM

Screen Shot 2023-04-24 at 10 14 11 AM

 

**Admin UI:**

Screen Shot 2023-04-25 at 10 08 13 AM

Localhost and new spec

### DIFF
--- a/app/components/admin/navigation_component.rb
+++ b/app/components/admin/navigation_component.rb
@@ -9,7 +9,7 @@ module Admin
                                     ['Search for user', admin_users_path],
                                     ['Generate collection report', new_admin_collection_report_path],
                                     ['Generate item report', new_admin_work_report_path],
-                                    ['Edit home page message', admin_page_content_index_path]],
+                                    ['Edit page content', admin_page_content_index_path]],
                                    request.env['PATH_INFO'])
       select_tag 'path', options, class: 'form-select', onchange: 'window.location.href = this.value'
     end
--- a/app/components/admin/navigation_component.rb
+++ b/app/components/admin/navigation_component.rb
@@ -3,6 +3,12 @@
 module Admin
   # Displays the drop down navigation for the admin pages
   class NavigationComponent < ApplicationComponent
+    def initialize(selected: nil)
+      @selected = selected
+    end
+
+    attr_reader :selected
+
     def dropdown
       options = options_for_select([['Admin page', admin_path],
                                     ['Search for DRUID', admin_druid_searches_path],
@@ -10,7 +16,7 @@ module Admin
                                     ['Generate collection report', new_admin_collection_report_path],
                                     ['Generate item report', new_admin_work_report_path],
                                     ['Edit page content', admin_page_content_index_path]],
-                                   request.env['PATH_INFO'])
+                                   selected || request.env['PATH_INFO'])
       select_tag 'path', options, class: 'form-select', onchange: 'window.location.href = this.value'
     end
   end
--- a/app/components/admin/navigation_component.rb
+++ b/app/components/admin/navigation_component.rb
@@ -8,7 +8,8 @@ module Admin
                                     ['Search for DRUID', admin_druid_searches_path],
                                     ['Search for user', admin_users_path],
                                     ['Generate collection report', new_admin_collection_report_path],
-                                    ['Generate item report', new_admin_work_report_path]],
+                                    ['Generate item report', new_admin_work_report_path],
+                                    ['Edit home page message', admin_page_content_index_path]],
                                    request.env['PATH_INFO'])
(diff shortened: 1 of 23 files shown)


app/controllers/dashboards_controller.rb
```diff
@@ -10,6 +10,7 @@ def show
 
     authorize! :dashboard
     @presenter = build_presenter
+    @page_content = PageContent.find_by(page: 'home')
   end
```
peetucket, line 14: The dashboard page will display the same text as the home page (but could easily be different in the future if we wanted with this design).

app/components/admin/navigation_component.rb
```diff
@@ -3,13 +3,20 @@
 module Admin
   # Displays the drop down navigation for the admin pages
   class NavigationComponent < ApplicationComponent
+    def initialize(selected: nil)
+      @selected = selected
+    end
+
+    attr_reader :selected
+
     def dropdown
```
peetucket, line 12: this `selected` value allows us to choose a different item in the admin selection drop-down menu to be selected from the page currently being rendered (which is useful in this PR)

app/components/page_content_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="page-content-message">
+    <p><i class="fa-solid fa-asterisk" style="color: #FEC51D;"></i> <%= page_content.value %> <%= link_to page_content.link_text, page_content.link_url if page_content.link_visible %></p>
```
justinlittman, line 2: Can we not put the `style` in here?

app/components/dashboard/page_content_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Dashboard
+  # Renders an optional user message on the dashboard page
+  class PageContentComponent < ApplicationComponent
```
justinlittman, line 5: Can you explain why there are multiple variations of the PageContentComponent?

peetucket: One is for the home page, the other is for the dashboard.  The styling is different between them.

We could use a single component, but there would be some additional logic required, so tradeoff between simpler component with some duplication (current), vs less duplication but more complicated (potential change).

app/components/page_content_component.html.erb
peetucket: like make the color a css class and then add it here instead of inlining via style tag?

justinlittman: Yes, use a CSS class rather than an inline style.

peetucket: ok, changed to css style

PR #3064, opened 2023-05-08
Fix ruby version on readme

jcoyne:
See https://github.com/sul-dlss/puppet/blob/production/hieradata/node/sul-h2-prod.stanford.edu.eyaml#LL11C20-L11C26

Fixes incorrect data. Adds herons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Self-Deposit for the Stanford Digital Repository (SDR)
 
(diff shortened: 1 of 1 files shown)


mjgiarlo (approved):

ndushay (approved):

peetucket:
Looks like we updated to 3.2.2 and the readme got updated since then, so except for our nifty little extra icons, this is not needed (or we need a rebase).

PR #3109, opened 2023-06-15
[WIP] fix year issue in chrome

### LINKED ISSUE #2432: Year field fix from #2357 should also work in Chrome

peetucket:

Fixes #2432 (code stolen from #2442), then making modifications to the tests to get around the problem with JS events not working as expected in our CI.

Everything works as expected in browser for Firefox and Chrome.

Localhost Chrome and Firefox

### DIFF
--- a/app/javascript/controllers/year_field_controller.js
+++ b/app/javascript/controllers/year_field_controller.js
@@ -1,7 +1,7 @@
 import { Controller } from "stimulus"
 
 export default class extends Controller {
-  
+
   connect() {
     this.hasChanged = false
   }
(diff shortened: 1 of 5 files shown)


peetucket:
Doh.  Tests fail here just like in #2442 👎

app/javascript/controllers/year_field_controller.js
```diff
@@ -12,8 +12,11 @@ export default class extends Controller {
     if (this.hasChanged)
         return // nothing to do
 
-    if (evt.inputType === "insertReplacementText") // So that typing in digits doesn't do this, only arrow keys or the arrow buttons
-    evt.target.value = evt.target.max
+    // Be careful refactoring this because Firefox and Chrome send different event types when doing arrow buttons
```
peetucket, line 15: basically, Firefox sends "insertReplacementText" for arrow keys, and Chrome sends nothing...so instead of checking for an event to trigger the default year, we instead check for typing events to ignore (which can be several different ones)

justinlittman:
@amyehodge Rather than add more sketchy javascript, can we just get rid of stepper arrows? For years, are they really helpful compared to just entering a 4 digit year?

amyehodge:
@justinlittman @peetucket As I mentioned over in the ticket, I prefer getting rid of the arrows to leaving it in its current state. Given the amount of brain power already devoted to this, I'm thinking we won't come up with anything easy to implement at this point, so removing the arrows and then adding whatever validation is needed to make sure this is a year is fine with me.

peetucket:
We just removed the arrows from this field -- this is no longer needed.

Issue #3119, opened 2023-06-22
Provide a valid label for form fields on deposit form


amyehodge:
This was identified as a critical issue by SODA during their recently completed accessibility review of H2. The following information was provided in that report.

[Issue]
There are form fields that do not provide a programmatically associated label that describes the purpose of each field.  The "Add your files*" radio buttons do not have valid labels because the ids are not unique.  In the "Enter dates related to your deposit" section, the "Approximate (can't be selected if 'day' is entered)." checkbox has a <label> with a for value of "work_created(approx0)" but the input id is "work_created_approx0_" for the field it should label.

[User Impact]
Users may not be able to fill out form fields correctly, if there is no programmatically associated and visible label for each field.

[Recommendation]
Without unique id values, or <label> for attribute values that exactly match the form control id, a valid label cannot be determined.  Ensure id values are unique.  Ensure that the <label> for value matches the input id value .

Input form fields should provide visible, programmatically associated labels that can be announced properly to screen reader users. Typically labels and form fields need to be explicitly programmatically associated which is commonly done by providing a "for" attribute on the <label> element with its value set to the input's id value or by providing another mechanism that specifies an accessible name.

Astrid will validate if fixed.

justinlittman:
@amyehodge  Are the fields given in the description the exhaustive list of that need to be remediated?

amyehodge:
@justinlittman I would not assume that what is included above is an exhaustive list. I'm basing this on the following from the report from SODA:

"Please note that only a limited number of pages were tested and, for each page, a limited number of each violation were noted. Developers should take the information provided in this report and apply it across all pages of the website or application."

justinlittman:
@astridu How do we test this?

astridu:
I'll ask Kevin for a specific example, but I'm guessing this applies to all radio buttons where the label does not describe the radio button. So the screen reader would just read 'work upload type' for all 3 radio buttons:
Current: <label class="form-check-label fw-semibold" for="work_upload_type">Upload one or more files (less than 10GB) to be displayed as a flat list of files.</label>
Solution: <label class="form-check-label fw-semibold" for="Files less than 10GB displayed as flat file list">Upload one or more files (less than 10GB) to be displayed as a flat list of files.</label>

astridu:
This is the response from Kevin. Looks like we can either label each of them distinctly or put the text of the radio button in the label:
[...]

PR #3122, opened 2023-06-23
Label accessibility on deposit form.

justinlittman:
refs #3119

Accessibility

Local

### DIFF
--- a/app/components/works/dates_component.html.erb
+++ b/app/components/works/dates_component.html.erb
@@ -72,7 +72,7 @@
                          date_clear_target: "approximate",
                          action: "date-approximate#change"},
                   class: "form-check-input" %>
-            <%= form.label "created(approx0)", "Approximate (can't be selected if 'day' is entered)" %>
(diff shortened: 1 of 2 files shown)


app/components/works/dates_component.html.erb
```diff
@@ -72,7 +72,7 @@
                          date_clear_target: "approximate",
                          action: "date-approximate#change"},
                   class: "form-check-input" %>
-            <%= form.label "created(approx0)", "Approximate (can't be selected if 'day' is entered)" %>
+            <%= form.label "created_approx0_", "Approximate (can't be selected if 'day' is entered)" %>
```
ndushay, line 75: are these underscore changes needed in these form labels?

ndushay (approved):
approving with a small question

app/components/works/dates_component.html.erb
justinlittman: Yes, it what makes them match. (Don't ask me to explain why ...)

PR #3128, opened 2023-06-27
Adds admin function for work types.

### LINKED ISSUE #3092: Admins should be able to edit a work type after an object is created
The work type is stored in the `WorkVersion` model and is set when a user first creates a work in H2 via a modal.  After that point, it is not editable by anyone.

As an admin, I want to be able to change the work type for an item so that it meets best practice guidelines, displays in SW in a way that is appropriate for the kind of work that it is, and has a work type that is consistent with other similar works.

- The desired functionality is to add a new admin function that does not open the full work edit form, but showing the full work form is acceptable. 
- Add this feature as one of the list items in the Admin functions on each item view page (see below).
 
 

- If the item is in a disallowed editing state, the option should be greyed out/unavailable.
- If the item is in the deposited state, the admin can select this option. They will then see the work type modal with the current work type selected so that they can make a new selection. 
- The "what's changed" field should automatically be set with "work type changed".
- Once the work type and subtypes on a deposited work are modified, the item will need to be re-accessioned, which means it likely needs to be versioned (in H2 and thus in SDR as well).   This will ensure the PURL is republished and SW record is updated.

Disallowed editing states: Reserving PURL, PURL reserved, Initial draft, Pending approval, Returned, Deposit in Progress, New version in draft, Decommissioned

See https://docs.google.com/document/d/1AeWIp1NulYb3AwXB2ZxwgbFxD33ZAOrqseq7Ixa7_yc/edit and https://projects.invisionapp.com/share/NP138M0QX7T3#/screens/473244398_admin

justinlittman:
closes #3092

* Allow admins to change work types without using Argo.
* Update stimulus.

Amy


app/components/collections/show/details_header_component.html.erb
```diff
@@ -4,7 +4,7 @@
   <% end %>
   <div class="row">
     <div class="col"> </div>
-    <div class="col-6" data-controller="work-type purl-reservation">
+    <div class="col-6" data-controller="work-type work-type-modal purl-reservation">
```
justinlittman, line 7: The work type functionality was split into 3 different controllers for base work type modal functionality, select work type when creating work functionality, and change work type admin functionality.

This required adding in additional controllers in various places for this change.

app/components/works/work_type_modal_component.html.erb
```diff
@@ -14,39 +14,45 @@
       </div>
 
       <div class="modal-body">
-        <form method="get" action="/collections/1/work/new" class="mb-4" data-work-type-target="form">
+        <form method="<%= form_method %>" action="/collections/1/work/new" class="mb-4" data-work-type-modal-target="form" data-work-type-target="form" data-work-type-edit-target="form" <% unless turbo? %> data-turbo=false<% end %>>
```
justinlittman, line 17: The work type form had to become way more complicated to handle this use case (accounting for using PUT and working around turbo).

app/javascript/controllers/admin_date_range_controller.js
```diff
@@ -1,4 +1,4 @@
-import { Controller } from "stimulus"
+import { Controller } from "@hotwired/stimulus"
```
justinlittman, line 1: To be able to use outlets, stimulus had to be updated. In the update, the package name for stimulus changed.

package.json
```diff
@@ -16,8 +16,8 @@
     "lodash.debounce": "^4.0.8",
     "sass": "^1.43.2",
     "simple-datatables": "^3.0.0",
-    "stimulus": "2.0.0",
-    "stimulus-autocomplete": "^2.0.1-phylor-6095f2a9"
+    "@hotwired/stimulus": "^3.2.1",
+    "stimulus-autocomplete": "^3.1.0"
```
justinlittman, line 20: Also updated stimulus-autocomplete to work with newer stimulus.

aaron-collier (approved):

PR #3130, opened 2023-06-27
Update PR template to reflect new accessibility requirement

mjgiarlo:
# Why was this change made? 🤔

Discussed during dev planning earlier today. We are taking a "first, do no harm" approach to accessibility going forward so we stop introducing new a11y problems.

N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,16 @@
-## Why was this change made? 🤔
+# Why was this change made? 🤔
 
 
 
-## How was this change tested? 🤨
+# How was this change tested? 🤨
 
 ⚡ ⚠ If this change involves consuming from or writing to another service (or shared file system), ***run [integration test create_object_h2_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test manually in [stage|qa] environment, in addition to specs. ⚡
+
+
+
+# Does your change introduce accessibility violations? 🩺
+
+⚡ ⚠ Please ensure this change does not introduce accessibility violations (at the WCAG A or AA conformance levels); if it does, include a rationale. See the [Infrastructure accessibility guide](https://github.com/sul-dlss/DeveloperPlaybook/blob/main/best-practices/infra-accessibility.md) for more detail. ⚡
+
 
 


.github/pull_request_template.md
```diff
@@ -1,9 +1,10 @@
-## Why was this change made? 🤔
```
mjgiarlo, line 1: Get out of the practice of choosing a heading level based on how it looks vs. its semantics (an accessibility concern itself).

.github/pull_request_template.md
```diff
@@ -1,9 +1,10 @@
...
+# Why was this change made? 🤔
 
 
 
-## How was this change tested? 🤨
+# How was this change tested? 🤨
 
 ⚡ ⚠ If this change involves consuming from or writing to another service (or shared file system), ***run [integration test create_object_h2_spec.rb](https://github.com/sul-dlss/infrastructure-integration-test)*** and/or test manually in [stage|qa] environment, in addition to specs. ⚡
 
+# Does your change introduce accessibility violations (at the WCAG A or AA conformance levels)? 🩺
```
ndushay, line 9: consider leaving space to answer this and the previous question.

ndushay: consider:

```
# If your change affects user facing pages, did you check before and after with the siteimprove Accessibility Checker?
```

ndushay (commented):
a couple comments for consideration.

.github/pull_request_template.md
mjgiarlo: @ndushay I added space, re-phrased, and added a link to the (in-progress) Infra a11y guide. Have another look?

aaron-collier (approved):

.github/pull_request_template.md
jmartin-sul: i really wish there was a  way to downsize the headings across the board in a doc/comment/etc if desired, so that we could have less giant headings that still made sense semantically.  but alas, i couldn't find anything in the github markdown docs indicating that's possible.  but agree that this change is the right thing to do.

jmartin-sul (approved):

.github/pull_request_template.md
mjgiarlo: Something something user-specific stylesheets, but I have no idea what that means in practice.

.github/pull_request_template.md
ndushay: great - thanks Mike!

PR #3131, opened 2023-06-27
fix accessibility for help form by adding namespace

### LINKED ISSUE #3118: Provide a valid label for form fields on welcome screen
This was identified as a critical issue by SODA during their recently completed accessibility review of H2. The following information was provided in that report.

[Issue]
There are form fields that do not provide a programmatically associated label that describes the purpose of each field.  Each of the form fields is duplicated in the Help modal dialog and the relationships between the for and id values are negated because they are duplicated.

[User Impact]
Users may not be able to fill out form fields correctly, if there is no programmatically associated and visible label for each field.

[Recommendation]
Input form fields should provide visible, programmatically associated labels that can be announced properly to screen reader users. Typically labels and form fields need to be explicitly programmatically associated which is commonly done by providing a "for" attribute on the <label> element with its value set to the input's id value or by providing another mechanism that specifies an accessible name.  However, each id must be unique.

Astrid will validate if fixed.

ndushay:

Fixes #3118.

The first time user page uses the same form as the help link in the header (compare the background form with the help modal):

image

Adding a namespace to the [ActiveRecord::FormHelper](https://api.rubyonrails.org/v7.0.5.1/classes/ActionView/Helpers/FormHelper.html#method-i-form_with) call fixes the problem.

## After Change

image

## Before Change

image

### DIFF
--- a/app/views/helps/new.html.erb
+++ b/app/views/helps/new.html.erb
@@ -1,5 +1,5 @@
 <turbo-frame id="<%= params.fetch(:id, "help") %>">
-  <%= form_with(url: help_path(id: params[:id]), data: {controller: "help"}) do |form| %>
+  <%= form_with(url: help_path(id: params[:id]), data: {controller: "help"}, namespace: params[:id]) do |form| %>
     <strong><sup>*</sup>Required fields</strong>
     <div class="container-fluid">
       <div class="row">


justinlittman (approved):
Love it!

Issue #3060, opened 2023-04-26
GlobusSetupJobs continues to retry after the work version has been deleted


amyehodge:
The GlobusSetupJob fails and is left retrying when its associated draft/WorkVersion is discarded. This can occur when a user saves a new item draft or a new version of an existing item with the Globus option selected and then later discards the draft.

HB catches this situation (see https://app.honeybadger.io/projects/77112/faults/95498535), but we may want to consider a way to catch and deal with this in an automated way if it becomes a nuisance.

Users are not impacted by this, nor are the H2 admins -- outside of seeing this error in HB. So, it's a question for the developers whether this reaches nuisance level or not and whether it is worth figuring out the best way to handle WorkVersions not being found by the job or what would be involved in doing this.

PR #3133, opened 2023-06-27
Discard GlobusSetupJob for deserialization error.

justinlittman:
closes ~~#3050~~ #3060

Avoid retrying jobs that will never succeed.


ndushay:
Your issue link is wrong -- hard to review without context.  I found the correct issue of #3060 and changed it

ndushay (approved):
Seems like a good approach to me;  would like someone else to approve as well.

aaron-collier (approved):

PR #3136, opened 2023-06-27
Adds gtag.

### LINKED ISSUE #3089: Add Google Analytics to H2
1. Need to figure out what is the current and correct way of setting up a new analytics project.  Track down who is responsible for managing any shared DLSS accounts.
2. Create a new project for H2 in Google Analytics.
3. Figure out what is the current way of adding code to the project, i.e. is the analytics key ok to put in Github, does it need to in shared configs, etc?
4. Add the relevant code/config to the H2 project.
5. Verify analytics are being collected.

May require discussion with Ops about limits of properties we can connect to google analytics.

This google doc describes the requirements of what needs to be tracked: https://docs.google.com/document/d/1-f7NumL-VkSlo2ZvYmRoHIV34pnJWl-Zequ0s3TuaT4/

Check with Marlo on Access on next steps/logistics.

See https://developers.google.com/analytics/devguides/collection/ga4/tag-options

justinlittman:
closes #3089

Analytics!

QA

image

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,6 +1,18 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <% if Settings.google_analytics %>
+      <!-- Google tag (gtag.js) -->
+      <script async src="https://www.googletagmanager.com/gtag/js?id=G-L3QLH9N3QH"></script>
+      <script>
+        window.dataLayer = window.dataLayer || [];
+        function gtag(){dataLayer.push(arguments);}
+        gtag('js', new Date());
+
+        gtag('config', 'G-L3QLH9N3QH');
+      </script>
+    <% end %>
+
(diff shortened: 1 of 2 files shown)


justinlittman:
Please also review https://github.com/sul-dlss/shared_configs/pull/2013

app/views/layouts/application.html.erb
```diff
@@ -1,6 +1,18 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <% if Settings.google_analytics %>
+      <!-- Google tag (gtag.js) -->
+      <script async src="https://www.googletagmanager.com/gtag/js?id=G-L3QLH9N3QH"></script>
```
aaron-collier, line 6: Is the ID safe to hardcode? Or will we only deploy this with `google_analytics: true` in one environment or something?

aaron-collier: It looks like we only turn this on in prod, so this is fine for now. If we change that, then a follow up PR can address :+1:

aaron-collier (approved):

PR #3141, opened 2023-06-27
Remove bullet on files list

### LINKED ISSUE #3025: Remove bullets from list of files on view page.
These bullets are really cumbersome for people using screen readers, because it reads "bullet" every time. This is confusing and sounds like there are a bunch of files named "bullet."

aaron-collier:
# Why was this change made? 🤔

Closes #3025 

Before:
Screenshot 2023-06-27 at 1 31 29 PM

After:
Screenshot 2023-06-27 at 1 31 43 PM

### DIFF
--- a/app/components/works/show/attached_file_component.html.erb
+++ b/app/components/works/show/attached_file_component.html.erb
@@ -1,6 +1,5 @@
 <tr>
     <td class="file-level" style="--depth: <%= depth + 1 %>rem">
-        &bull;
         <%= basename %>
         <% if can_download? %>
             <%= link_to path, class: "ms-2", data: {turbo: false} do %>


mjgiarlo (approved):

mjgiarlo (changes requested):
I approved, but now I'm wondering if we need a different fix. These bullets were only recently added to support hierarchical files:

https://github.com/sul-dlss/happy-heron/pull/2945/files

What will a folder upload look like without these bullets? cc: @peetucket @astridu @amyehodge These came from the following mock-up? https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/471005875 Do we need a new one?

peetucket:
Yes, thanks Mike.  Can we do a test of this PR with hierarchical content to see if the indentation is kept?  My concern is it may not be. You can do this by zipping up a local folder that has some nesting and files at each nested level, uploading the zip and let it get expanded.  

I presume if the indenting for hierarchy is kept, and the bullets are gone, that would probably be fine.  If the indenting is gone, we'd need to come up with another solution.

aaron-collier:
Nested display:

Screenshot 2023-06-27 at 3 20 33 PM

jmartin-sul (approved):

mjgiarlo (approved):

Issue #3114, opened 2023-06-21
Investigate limit of what can be deposited in H2


peetucket:
Note: test in stage, which is where disk limits were increased to accommodate testing

We need to understand based on limits provided to us by PO, and for this we need to have the requested maximum number of files, maximum single file size, and maximum total file size that should be possible to accession so that we can test if this is possible in QA or stage.  If it is, we could presume it would also be possible in production (which has more resources).

We can then determine if we can accession without problems (e.g. timeouts, pages not rendering, stalls in workflows, filling up stacks, etc.):

1. The maximum number of files in an object required
2. The maximum single file size in an object required
3. The maximum combined size of all files in an object required

We can then take further steps:

1. Investigate how to increase the limits if not sufficient.
2. Add validation/messages to the user if the limits are sufficient, but we need to prevent users from exceeding them

From Amy June 27:

Requested limits (specifically to H2 and assuming usage of globus):
- 4 TB total object size
- 25k total number of files
- 4 TB total single file limit

We are not sure if we currently enforce any limits in the UI (i.e. browser uploads), other than ZIP upload being a single file. See #3183

Question:
- can we reject objects that exceed these limits if deposited into H2?  YES - early in process is better

amyehodge:
@peetucket I saw this and have reached out to Andrew and Hannah to discuss. I understand that problem with trying to test to make sure "anything" will work, so I'll work to get you some concrete numbers as soon as possible.

peetucket:
Thanks!  I was wondering too if we could try and find the current limits, but the challenge of this approach is that the limits are likely to be different in qa/stage vs prod due to differences in the resources/disk space/compute power in each environment.  

For example, if we spent time finding the limits of QA, it's likely prod will be higher (but unclear as to how much higher).  And relatedly that if we determine that a reasonable limit is possible in QA, its very likely this limit is possible in prod.

If we really wanted to test the current limits, we'd have to do in production to be accurate, and this may not be desirable (it.e. breaking prod on purpose) it we can avoid it.

amyehodge:
@peetucket Agreed that testing limits in prod that could interrupt depositing for everyone does not seem optimal. And your other logic makes sense as well.

Though, conversely, if we could find the limits on qa/stage, could we reasonably surmise that the limits in prod are at least that much? That might be helpful if it's maybe half the limit of prod, but if it's 1/10th, that seems less useful.

peetucket:
[...]

PR #3157, opened 2023-06-28
Adds script for generating test data.

justinlittman:
refs #3114

# Why was this change made? 🤔

### DIFF
--- a/bin/test-files
+++ b/bin/test-files
@@ -1,0 +1,73 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+require 'optparse'
+require 'faker'
+require 'fileutils'
+require 'securerandom'
+require 'tty-progressbar'
+require 'pathname'
+
+options = { segments: 0, files: 1, zip: false }
+parser = OptionParser.new do |option_parser|
+  option_parser.banner = 'Usage: bin/test-files [options]'
+  option_parser.on('-fFILES', '--files FILES', Integer, "Number of files. Default is #{options[:file]}.")
+  option_parser.on('-sSEGMENTS', '--segments SEGMENTS', Integer, "Number of directory path segments. Default is #{options[:segments]}.")
+  option_parser.on('-dDIR', '--dir DIR', String, "Directory for the files. If not will be provided, a directory name will be generated.")
+  option_parser.on('-z', '--zip', "Zip the files.")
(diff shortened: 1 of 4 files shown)


bin/test-files
```diff
@@ -0,0 +1,72 @@
...
+base_path = options[:dir] || "test-data/test-data-f#{options[:files]}-s#{options[:segments]}"
+FileUtils.rm_rf(base_path) if Dir.exist?(base_path)
+
+progress_bar = tty_progress_bar(options[:files])
+progress_bar.start
+
+
+for i in 1..options[:files]
+  path = generate_path(base_path: base_path, segments: options[:segments])
+  filename = Faker::File.file_name(dir: '')
```
aaron-collier, line 66: @justinlittman there is a small limitation here where the name pool seems to be limited to 5621 items, so this is the maximum number of files that will be generated (even through the script doesn't stop, runs all the way through without error, files are just overwritten): 

```
Generating [▣▣▣▣▣▣] (100% (25000/25000), rate: 1397.23/s, mean rate: 1250.61/s, 19s total, ETA: 12:44:18)
ls -l test-data/test-data-f25000-s0 | wc -l
    5621
```

Updating this call to `Faker::File.unique.file_name(dir: '')`  results in an error:
```
Generating [▣▣▣⬚⬚⬚⬚⬚⬚⬚⬚] (23% (5824/25000), rate:  6.16/s, mean rate: 589.82/s,  9s total, ETA: 12:47:11)/Users/amcollie/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/faker-3.2.0/lib/helpers/unique_generator.rb:22:in `method_missing': Retry limit exceeded for file_name (Faker::UniqueGenerator::RetryLimitExceeded)
	from bin/test-files:66:in `block in <main>'
	from bin/test-files:64:in `each'
	from bin/test-files:64:in `<main>'
```
Note: this is in the README: https://github.com/faker-ruby/faker#ensuring-unique-values


I don't necessarily think this should hold up this PR, but should be called out in a comment or something. I'll poke at the library a bit more to see if I can offer a suggested approach.

aaron-collier: This may not be the best approach, but for me locally, I updated line 66 here to be:

```
filename = "/tf_#{i}_#{File.basename(Faker::File.file_name(dir: ''))}"
```

aaron-collier (approved):

Issue #3116, opened 2023-06-21
Provide link to Globus folder we create to user on view page


peetucket:
Provide the Globus link to the user on the view page for the item instead of having to rely on the Globus email. This totally eliminates the requirement for the user to check their email during this workflow. The link to the instructions shown here should go to https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit (top of the page and not a specific header).

![Screenshot 2023-04-26 at 10 33 51 AM](https://user-images.githubusercontent.com/5558511/234657037-985ea3bf-6f6c-41cc-90b5-ec72b88c2373.png)

See also #3021, #3022 and #3049

Full design: See designs at https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/472346509_Globus_Auto_Account

Information from the folks at Globus on how to create these links:
 
- File Manager application links with pre-selected endpoints
— Query parameters supported: destination_id, destination_path, origin_id and origin_path
-- Example https://app.globus.org/file-manager?destination_id=ddb59aef-6d04-11e5-ba46-22000b92c6ec&destination_path=%2F~%2F&origin_id=2fde89c0-6fb4-11eb-8c47-0eb1aa8d4337

Note that we currently have Globus send the link to the shared folder, I think because we didn't know how to figure out what the direct link would be, but they say this is knowable:

origin_id is the endpoint id of source, origin_path is the source path. Those must be available from when you set permission for the user on a specific folder. You can use that to construct the URL.

PR #3158, opened 2023-06-28
[HOLD] Globus workflow allows all valid users to proceed

### LINKED ISSUE #3021: Create user folder for Globus workflow on initial save, whether user has logged into Globus before or not
For those who have not logged into a Stanford Globus account before, making this change will eliminate the need for that user to save the item, log into Globus, then return to H2 and notify us that they have completed that step. So, the user journey for this feature would be the same for all users.

We recently learned from the people at Globus (U. Chicago) that it is possible via the API to create the shared folder and later have the correct permissions assigned automatically once the user has logged into Globus. These are the details they provided to us on how to do this:

— Permissions can be set for a user who has not logged into Globus before
— The API call to get the identity UUID, Auth get_identities call is used. The provision=True option ensures that identity is created, as unclaimed. Globus SDK documentation: https://globus-sdk-python.readthedocs.io/en/stable/services/auth.html#globus_sdk.AuthClient.get_identities 

See also #3021, #3022 and #3049.
See designs at https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/472346509_Globus_Auto_Account.

lwrubel:
# Why was this change made? 🤔
HOLD since it needs to be deployed to qa/stage for @amyehodge's review in addition to code review.  

Resolves #3021 and #3116.

This removes the following states:
* globus_account_setup
* globus_setup_pending
* globus_setup_first_draft
* globus_setup_version_draft

If user has a valid status in Globus, on saving as draft they will see instructions and link to the endpoint:

Screenshot 2023-06-28 at 5 24 37 PM

If user has an invalid status, they will see an error. This can be re-worded. 
Screenshot 2023-06-28 at 5 24 06 PM

Testing in QA and integration tests. Unit tests updated/added. 

No


Gemfile
```diff
@@ -63,7 +63,8 @@ gem "druid-tools"
 gem "dry-types"
 gem "edtf"
 gem "faraday", "~> 2.0"
-gem "globus_client", "~> 0.8"
+# gem "globus_client", "~> 0.8"
+gem "globus_client", git: "https://github.com/sul-dlss/globus_client", branch: "t3021-add-folder-access"
```
lwrubel, line 67: This will need to be updated before merging, once the globus_client release is cut.

lwrubel: Updated now.

justinlittman (approved):
All those deletions must have been very satisfying ...

lwrubel:
It really was nice to pull out all those steps.

PR #3162, opened 2023-06-29
Ror autosuggest

justinlittman:
# Why was this change made? 🤔
Prep work for supporting affiliation

Also, renames existing FAST autosuggest to avoid confusion.

Unit

image


app/components/works/contributor_row_component.html.erb
```diff
@@ -63,6 +63,10 @@
             </div>
           </div>
         </fieldset>
+        <%# This can be removed after affiliation implementation. %>
+        <% if Settings.ror_demo %>
+          <%= render("works/organization_input") %>
+        <% end %>
```
mjgiarlo, line 69: If we didn't want this logic living in a view, we could make the `works/organization_input` view a new view component, and the Ruby side of the component could have:

```ruby
def render?
  Settings.ror_demo
end
```

But since this code looks like it's going to be remove before too long, it may not be worth the effort?

app/controllers/ror_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# A proxy to ROR for typeahead
+class RorController < ApplicationController
+  include Dry::Monads[:result]
+
+  def show
+    result = lookup(params.require(:q))
+    @suggestions = result.value_or([])
+    if result.success?
+      render status: :no_content, html: "" and return if @suggestions.empty?
```
mjgiarlo, line 11:
```suggestion
      return render status: :no_content, html: "" if @suggestions.empty?
```

app/controllers/ror_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+      render status: :ok, layout: false
+    else
+      logger.warn(result.failure)
```
mjgiarlo, line 15: Probably not honeybadger-worthy, right?

app/views/ror/show.html.erb
```diff
@@ -0,0 +1,3 @@
+<% @suggestions.each do |value, label| %>
+  <li class="list-group-item" role="option" data-autocomplete-value="<%= value %>" data-autocomplete-label="<%= label %>"><%= html_escape(label) %></li>
```
mjgiarlo, line 2: Is Siteimprove happy with this `ul`/`li` arrangement?

app/views/works/_organization_input.html.erb
```diff
@@ -0,0 +1,7 @@
+<%# This is only for demo purposes before affiliation implementation %>
+<div data-controller="autocomplete" data-autocomplete-url-value="/ror" data-autocomplete-min-length-value="3" class="dropdown" role="combobox" aria-expanded="false" data-autocomplete-ready-value="true">
+  <input class="form-control" data-autocomplete-target="input" type="text" name="DEMO-ORGANIZATION" id="DEMO-ORGANIZATION" autocomplete="off" spellcheck="false">
+  <!-- This input is populated by autocomplete controller.-->
+  <input type="hidden" data-autocomplete-target="hidden" data-autocomplete-edit-target="value">
+  <ul class="list-group" data-autocomplete-target="results" hidden=""></ul>
```
mjgiarlo, line 6: Is Siteimprove happy with these changes or do we need some aria labels?

app/components/works/contributor_row_component.html.erb
justinlittman: Yeah, the input is throw away code.

app/controllers/ror_controller.rb
justinlittman: Probably not. (This was copied from FAST autosuggest, where we don't HB either.)

app/views/ror/show.html.erb
justinlittman: It is what is currently used for FAST autosuggest and I don't think it was flagged.

app/views/works/_organization_input.html.erb
justinlittman: Since this is going to be thrown away, I didn't SiteImprove it. All this does is put a plain-old input on the page without even a label.

mjgiarlo (approved):

PR #3167, opened 2023-06-29
Allows user to select globus origin.

### LINKED ISSUE #3022: Allow user to specify origin location on Globus
1. Allow the user to tell us if their content is on Google Drive, Oak, or Sherlock so that we can send a more precise link that will open the Globus file manager with the origin open on one side and the destination open on the other side. This will make getting to the right place in Globus much easier for users, because it's very challenging to navigate around the Globus interface.
2. Allow the user to tell us if their content is on a different endpoint (by not selecting any of the fixed options) ~~and provide us with the info we need to generate the more precise link to that endpoint. What we need from them may be the UUID for the endpoint, but we need to look into that.~~ 
3. The link in box/step 3 should go to https://docs.google.com/document/d/10b7y3yZCOfyVJ_uP4l7QHbkILQKnfdJi9hj-MpmmdIk/edit (top of the page) and not to a specific header (I'm not sure where that points now). Also, we are not including the "Print instructions" link show in this image.
Updated design: image

4. Make additional changes to the "steps" on top of deposit form as shown here:
https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/472346509_Globus_Auto_Account

See also #3116, #3021 and #3049 
See designs at https://projects.invisionapp.com/share/YQ132HXFX2GR#/screens/472346509_Globus_Auto_Account

Information from the folks at Globus on how to create these links:
 
- File Manager application links with pre-selected endpoints
— Query parameters supported: destination_id, destination_path, origin_id and origin_path
-- Example https://app.globus.org/file-manager?destination_id=ddb59aef-6d04-11e5-ba46-22000b92c6ec&destination_path=%2F~%2F&origin_id=2fde89c0-6fb4-11eb-8c47-0eb1aa8d4337

Note that we currently have Globus send the link to the shared folder, I think because we didn't know how to figure out what the direct link would be, but they say this is knowable:

origin_id is the endpoint id of source, origin_path is the source path. Those must be available from when you set permission for the user on a specific folder. You can use that to construct the URL.

justinlittman:
closes #3022

# Why was this change made? 🤔

Unit

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -79,42 +79,29 @@
           <div class="col">
             <div class="border border-dark h-100 p-2 bg-light">
               <p>STEP 1</p>
-              <p>Scroll to the bottom of this page and click "Save as draft."</p>
-              <p><em>You can finish filling out the form now or later.</em></p>
+              <p>If your files are located on Stanford Google Drive, Oak, or Sherlock, select where they are located below.</p>
+              <%= form.select :globus_origin, origin_options, {prompt: "Select..."}, {class: "form-select", "aria-label": "Origin for Globus deposit"} %>
             </div>
           </div>
           <div class="col">
             <div class="border border-dark h-100 p-2 bg-light">
               <p>STEP 2</p>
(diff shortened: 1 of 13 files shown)


justinlittman:
image

ndushay (approved):
approving, but does this need PO approval before merge?

Issue #3171, opened 2023-06-30
Welcome page: "major" accessibility issues that may not need design


ndushay:
For the Welcome form (which can be the page for a first time user OR what we all see as the welcome page ... see PR #3131)

We addressed accessibility level A and AA problems noted by the SiteImprove, WAVE and Axe browser plugins in PRs #3169 and #3131.

When I reviewed https://docs.google.com/spreadsheets/d/1Fs5sY1zgpovWVdHMqpJWRu2Zvyr53jw7 afterwards, there appear to be some "major" accessibility issues that we can fix without design changes.

These can be separate PRs, and/or split into separate tickets.

- [ ] The programmatic (DOM) order does not match the visual reading order.  When one of the modal dialogs is open, the content behind the dialog is still conveyed.
- [ ] Although a skip link is present, the page does not contain a valid target for the skip link.  ```<a href="#content">Skip to main content</a>```
- [ ] The error messages appear briefly, but they do not remain visible. (filling out forms)
- [ ] When "Request access to another collection" is selected and the tooltips are visible, the controls that open the tooltips are not associated with the tooltip content. (Form field errors are not programmatically associated with the field to which they apply, and are not indicated programmatically.  The error messages are not programmatically associated with the form controls to which they apply.)
- [ ] There are custom controls that do not provide textual name (link or button label, form label text), role (i.e. identify as button, link, text field, etc.), and/or state information (i.e. "expanded", "collapsed", selected, current).
- [ ] There are lists that are only styled to look like a list but are not indicated as such programmatically.  Within the Terms of Deposit dialog, the list of links should be structured as a list.
- [ ] There are link elements that do not have meaningful text.  The image link with the logo for Stanford Libraries has an aria-label in addition to an image with alternative text and both instances have a value of "Stanford Digital Repository" without including the visible Libraries text.
- [ ] There are common input fields that lack autocomplete attributes to communicate their purpose programmatically: the "what is your name" and "what is your email address" fields in the main screen and in the Help dialog.
- [ ] There is no visual indication of keyboard focus when using TAB or SHIFT+TAB to navigate through some controls of the website, such as the Cancel control.

PR #3179, opened 2023-06-30
Some accessibility fixes for welcome/home page

peetucket:
# Why was this change made? 🤔

Addresses some of the accessibility issues identified in #3171 

Localhost and CI

I hope it reduces them!

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -97,7 +97,6 @@
         <%= render "shared/terms_of_deposit" %>
         <%= render "shared/help_modal" %>
         <%= yield %>
-
       </div>
     </div>
 
--- a/app/assets/stylesheets/frontpage.scss
+++ b/app/assets/stylesheets/frontpage.scss
@@ -107,6 +107,10 @@
   color: #FEC51D;
 }
 
+.btn-default {
+   --bs-btn-hover-border-color: var(--bs-link-hover-color);
+}
+
 .page-content-message {
   font-size: 1.1rem;
   font-style: italic;
--- a/app/components/help_collection_component.html.erb
+++ b/app/components/help_collection_component.html.erb
@@ -1,6 +1,6 @@
 
 <div class="form-check">
   <%= form.check_box(:collection, {multiple: true, class: "form-check-input"}, label) %>
-  <%= form.label :collection, label, value: label, class: "form-check-label" %>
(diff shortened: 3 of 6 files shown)


mjgiarlo (changes requested):
Should we sit on this until we can convert `I hope it reduces them!` to `We ran a set of tools to verify these issues, and now they're gone, and no new ones are introduced` ❓

ndushay (commented):
if you split this PR into the two commits, I'll merge a PR with the first commit (with one tweak I noted in a comment).

app/views/layouts/application.html.erb
```diff
@@ -96,6 +96,7 @@
         <%= render "shared/flashes" %>
         <%= render "shared/terms_of_deposit" %>
         <%= render "shared/help_modal" %>
+        <a name="content"></a>
```
mjgiarlo, line 99: @peetucket Can you say more about why we need this anchor? I thought the skip-to-content links were working fine whenever the DOM contained `<main id="content">`. Is that not the case? Or do we have pages that lack this in the DOM?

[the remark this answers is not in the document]
peetucket: both work fine, but sure (I've used empty names in the past like this and they work just fine as a target for an anchor link)

peetucket: It's not clear that we do, but it was listed as a ding (that the "skip to" link doesn't lead to a specific location, so I added it as a precaution in the layout in case there is a page somewhere that doesn't have a `<main id="content">`

i don't think it will cause any harm if duplicated

mjgiarlo: Yeah, I guess this is why I was so concerned about having tools for auditing this kind of stuff, so we don't add cruft where not needed.

mjgiarlo: > It's not clear that we do, but it was listed as a ding (that the "skip to" link doesn't lead to a specific location, so I added it as a precaution in the layout in case there is a page somewhere that doesn't have a `<main id="content">`
> 
> i don't think it will cause any harm if duplicated

FWIW, in my earlier a11y PRs in H2, I found one `<main>` element lacking the `id="content"` and addressed it. So this may already be addressed without needing the anchor.

peetucket:
1. split PR into two, second is here: #3186
2. removed `<a name="content">` tag based on Mike's comments

ndushay (approved):
approving this.  Agree that it would be good to have a tool to show us this fixed something and didn't break anything else.

mjgiarlo (approved):

Issue #3099, opened 2023-06-14
Map affiliation and multipart organizational contributor fields to Cocina and DataCite


peetucket:
Design: https://projects.invisionapp.com/share/NP138M0QX7T3#/screens/473166568
Requirements: https://docs.google.com/presentation/d/1FI9yn7vUSeX7Yd5IemrLxnYH2__JlTxNwgP8guATb18 

Background:
- we want to record contributor affiliations along with names (and ORCIDs)
- recording affiliations, especially with organization identifiers (RoRs) will enrich metadata and discoverability

We need a metadata mapping from the fields shown on the UI (i.e. affiliation string and possible RoR entry) to how it should appear and where in the cocina data model (for multiple cases, i.e. string only, RoR present with string, etc.).  We can then decide how to store locally in the database until it needs to be mapped to cocina.

See #3156 for the work that requires this

justinlittman:
What does "Save schema, identifierID, schema URI, identifier name, etc." mean? What is the exact data model for an affiliation? How does that data relate to what is available on the form? How does it get mapped to Cocina?

edsu:
I don't know if this helps but here is what contributors look like currently (via [cocina-shape](https://sul-dlss-labs.github.io/cocina-shape/)):

![Screenshot 2023-06-28 at 5 11 20 PM](https://github.com/sul-dlss/happy-heron/assets/33829/46c127b0-554a-43a6-a4b3-772ea39047d5)

arcadiafalcone:
Mappings:
- [ ] One affiliation with part 1/ROR
- [ ] One affiliation with part 1/ROR and part 2
- [ ] Multiple affiliations for single contributor
- [ ] Organizational author/contributor with ROR
- [ ] DataCite

If this info needs to go to ORCID I'll look at that on the ORCID ticket.

arcadiafalcone:
Person contributor with one affiliation, part 1 only, with ROR
```
{
    "contributor": [
        {
            "name": [
                {
                    "value": "Smith, Jane"
                }      
            ],
            "type": "person",
            "status": "primary",
            "identifier": [
                {
                    "uri": "https://orcid.org/0000-0000-0000-0000",
                    "type": "ORCID",
                    "source": {
                        "code": "orcid"
                    }
                }
            ],
            "note": [
                {
                    "type": "affiliation",
                    "value": "Stanford University",
                    "identifier": [
                        {
                            "uri": "https://ror.org/00f54p054",
[...]

PR #3189, opened 2023-07-05
Adds active record model for affiliations.

justinlittman:
refs #3099

# Why was this change made? 🤔
Support adding affiliations to UI and mapping to Cocina.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -10,13 +10,6 @@ SET client_min_messages = warning;
 SET row_security = off;
 
 --
--- Name: public; Type: SCHEMA; Schema: -; Owner: -
---
-
--- *not* creating schema, since initdb creates it
-
-
---
 -- Name: work_access; Type: TYPE; Schema: public; Owner: -
 --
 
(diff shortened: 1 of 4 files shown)


db/structure.sql
```diff
@@ -261,13 +288,13 @@ CREATE TABLE public.collections (
     access character varying,
     required_license character varying,
     default_license character varying,
-    email_when_participants_changed boolean DEFAULT true NOT NULL,
```
ndushay, line 264: are there settings to be changed so there is less structure.sql churn?  I think this got added by you on an earlier PR, but I'm not certain.

ndushay (approved):
So the idea is to have a basic table and the model will get additions as needed?

justinlittman:
Thanks for the review @ndushay. I have no idea what causes the structure.sql churn. If it persists, we can investigate further. The idea is to start with a basic table that only models what is necessary for the UI and the mapping to Cocina. This is consistent with how we have modeled other entities like keywords.

Thus, for the mapping to Cocina, I think you can safely assume that any URI will be a ROR URI.

PR #3199, opened 2023-07-07
Limit maximum files to 250

### LINKED ISSUE #3183: Verify number of files that can be uploaded via the browser
It's possible there is a limit of 1000 files that can be uploaded to H2 via browser uploads:

1. Verify if there is a total number of files limit that can be uploaded.
2. Verify if we enforce a total single file size limit of 10 GB (i.e. will it block you if you try and do this)
3. If there is a limit to total number of files, see if it actually works.
3. If there is not a limit, try to limit to 1000 files (check with Amy first).
4. Add instruction text to inform user of limit.

Current limit in instructions to user is 10 GB total for browser uploads, so the 1000 file limit should total to no more than 10 GB.

edsu:
# Why was this change made? 🤔

Limit the maximum number of files that can be uploaded individually to 250.
Doing any more causes DropZone to behave erratically in Firefox and Chrome.

It does not appear to be feasible to change the error message that is
displayed when a file exceeds 10GB to use GB instead the less common MiB
units. But a message is displayed so that's good.

Fixes #3183

By testing in my development environment with 250 small files, multiple 5GB files, and a 10GB file.

No


app/javascript/controllers/dropzone_controller.js
```diff
@@ -305,7 +305,7 @@ function createDropZone(controller, template) {
   return new Dropzone(controller.element, {
     url: controller.url,
     headers: controller.headers,
-    maxFiles: 1000,
+    maxFiles: controller.maxFiles,
```
edsu, line 308: The `maxFiles` is now configurable through data properties.

PR #3211, opened 2023-07-10
Update dependencies

sul-devops-team:


ndushay (approved):
I'm okay with this way or another way.  If changing the CI task still runs the other rubocops, it seems like a better way to go.  Or maybe using standard should go back on the dev planning agenda.

jmartin-sul:
> I'm okay with this way or another way. If changing the CI task still runs the other rubocops, it seems like a better way to go. Or maybe using standard should go back on the dev planning agenda.

fwiw, my intent is to circle back on this once this week's dep updates are shipped.  didn't seem like we should hold that deployment up over getting a new version of standard.  but also agree that pinning this does not feel like a good medium or long term solution.

jmartin-sul (approved):
Approved by automated merge script

PR #3216, opened 2023-07-10
[SPIKE] Use axe-core-rspec and nokogiri gems to test accessibility

mjgiarlo:
# Why was this change made? 🤔

This commit uses the axe-core-spec gem along with the newly added selenium web driver to audit the accessibility of selected pages. It also uses nokogiri to validate HTML of selected pages. This is all done to help improve accessibility and discourage regressions over time.

Note that this is a spike to show some patterns. We would want to add more page coverage to these new feature specs once we're settled on a pattern we like.

CI

Nope.


Gemfile
```diff
@@ -42,6 +43,7 @@ group :test do
   gem "capybara", "~> 3.34"
   gem "capybara-screenshot"
   gem "cuprite"
+  gem "selenium-webdriver"
```
mjgiarlo, line 46: The `axe-core-rspec` gem does not have any plans to support cuprite. It defaults to the selenium driver, so it is added here.

spec/features/accessibility_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Accessibility tests", driver: :selenium_chrome_headless do
```
mjgiarlo, line 5: Note these tests require the selenium web driver, and we have a new pattern (introduced below) for injecting a custom driver for a set of tests.

spec/features/accessibility_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+  it "audits the home page" do
+    visit root_path
+    expect(page).to be_accessible
+  end
```
mjgiarlo, line 9: We would obviously want to add a whole lot more URLs here! This is a start.

spec/features/html_validity_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "HTML validity" do
+  it "validates the home page" do
+    visit root_path
+    expect(page.body).to be_valid_html
+  end
```
mjgiarlo, line 9: We would obviously want to add a whole lot more URLs here! This is a start.

spec/features/html_validity_spec.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "HTML validity" do
+  it "validates the home page" do
+    visit root_path
+    expect(page.body).to be_valid_html
```
mjgiarlo, line 8: This is a new custom matcher introduced below.

aaron-collier (approved):

Gemfile
ndushay: make this a comment on line 46 please (e.g.,  # needed for axe-core-rspec)

I don't know what cuprite is for, and would welcome a comment there, fwiw.

ndushay (commented):
I like this approach, but I hear JLitt's concern about selenium as well.  

Can we do something this simple to add Cypress accessibility and validation tests?  (I think I saw a something about w3c html validation as a node package??)

spec/support/html_validator.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+# Use Nokogiri to validate HTML
+RSpec::Matchers.define :be_valid_html do
+  match do |actual_html|
+    Nokogiri::HTML5(actual_html).errors.empty?
```
justinlittman, line 6: Have you verified that this works as intended? At least according to https://nokogiri.org/rdoc/Nokogiri/HTML5.html "by default, no parse errors are reported".

Also, are you sure that this catches the sort of HTML5 errors that the W3C validator does? If this doesn't do much more than catch XML errors, not sure it is worth the overhead.

Gemfile
mjgiarlo: Done.

spec/features/accessibility_spec.rb
mjgiarlo: Yeah, we'd need to add the first-time welcome page. We presumably also want to add lots more pages (deposit form, collection pages, etc.).

spec/support/html_validator.rb
mjgiarlo: @justinlittman Good catch. I removed the `max_errors` param in a refactoring without knowing what it did. 😅 It's back now.

This would catch any errors encountered when using this parser: https://github.com/sparklemotion/nokogiri/tree/main/gumbo-parser/src

Not sure how that stacks up to the W3C validator. There's not a lot of code here we'd need to throw out, at least, if we come up a superior way to automate validation of app HTML.

spec/support/capybara_actions.rb
```diff
@@ -9,6 +9,11 @@ def blur_from(locator)
   def within_section(title, &)
     within(:xpath, "//section[contains(header/text(),'#{title}')]", &)
   end
+
+  # An alias so our tests are less coupled to the aXe implementation
+  def be_accessible(...)
+    be_axe_clean(...)
```
lwrubel, line 15: What's the default standard used for `be_axe_clean`? I see in their docs we could specify `according_to :wcaag2a, :wcag2aa` for example.

mjgiarlo: @lwrubel I'm just about to push up a change that makes this clearer. Thanks for flagging it!

spec/support/capybara_actions.rb
```diff
@@ -9,6 +9,16 @@ def blur_from(locator)
...
     within(:xpath, "//section[contains(header/text(),'#{title}')]", &)
   end
+
+  # An alias so our tests are less coupled to the aXe implementation
+  def be_accessible(...)
+    be_axe_clean(...).according_to(
+      :wcag2a,
+      :wcag2aa,
+      :"best-practice",
+      :experimental
```
mjgiarlo, line 19: @lwrubel Here you go.

mjgiarlo: See also https://github.com/dequelabs/axe-core/blob/master/doc/rule-descriptions.md for a list of the rules and which tags (e.g., `:wcag2aa`) include them.

spec/support/capybara_actions.rb
lwrubel: Looking again at the [tags in the docs](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#axe-core-tags), I think we'd want the following to be using WCAG 2.1:

```
:wcag21a
:wcag21aa
```

lwrubel: I'm not sure it makes sense to test WCAG 2.0 and WCAG 2.1 standards. I think WCAG 2.1 is what our target is.

lwrubel (approved):

spec/support/capybara_actions.rb
mjgiarlo: @lwrubel Even better.

.circleci/config.yml
```diff
@@ -33,11 +33,27 @@ jobs:
       - run:
           name: Lint ERBs
           command: 'bundle exec erblint --lint-all --format compact'
+  accessibility-features:
```
justinlittman, line 36: Ticket adding this to orb?

spec/features/html_validity_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "HTML validity" do
```
justinlittman, line 5: Can this be combined with the accessibility spec? Otherwise, we're unlikely to remember to add additional to both tests.

ndushay: Good point. Maybe there's a way to have a list of pages to loop through in one place?

.circleci/config.yml
mjgiarlo: I would like to defer until we have this in 2-3 codebases, which gives us time to work out the best patterns.

spec/features/html_validity_spec.rb
mjgiarlo: @ndushay I looked into automating this and it requires going way outside RSpec best practices, and it was ugly.

@justinlittman I can add these tests to the a11y spec, sure.

app/components/works/detail_component.html.erb
```diff
@@ -15,7 +15,7 @@
   <% end %>
 
   <header class="title clearfix"><%# clearfix here prevents the EditButtonComponent from spilling outside the container if the title is long %>
-    <span class="header-text"><%= title %></span>
+    <h1 class="header-text d-inline-block"><%= title %></h1>
```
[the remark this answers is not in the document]
mjgiarlo: We already made this change in a bunch of places and Amy was fine with the change. It's not noticeable.

spec/features/html_validity_spec.rb
mjgiarlo: Combined these into a single file.

PR #3219, opened 2023-07-10
_collection_summary_row.html.erb: make link text more meaningful

### LINKED ISSUE #3190: Dashboard: "Copy persistent link" elements need unique, meaningful text
similar to #3188 

from SODA report:

[Issue]
There are link elements that do not have meaningful text.  Within the table, there is link text that is not unique, such as the "Copy persistent link" links.

[User Impact]
Without meaningful text, screen reader users who navigate the page or navigate using a list of links (provided by their Assistive Technology, e.g., JAWS, VoiceOver, etc...) will not be able to discern the purpose of these links.

[Recommendation]
Developers must ensure that links have meaningful text within context. This can be achieved by changing the on-screen text or by using an aria-label, aria-labelledby to provide link purpose in context.

- The aria-label can be used to include the deposit name, which is what is being done for the edit links in the Actions column.  
- Because the link is in a table, this requirement can also be met by making the deposit name a row header.  However, this would not help screen reader users when reviewing a list of the page links.

ndushay:
# Why was this change made? 🤔

Fixes #3190 

Per suggestion from SODA report:  "The aria-label can be used to include the deposit name, which is what is being done for the edit links in the Actions column."

## After this fix

image

image

image

## Before this fix

image

image

image

wrote a spec, plus ran SiteImprove, WAVE and axe tool

No.

### DIFF
--- a/spec/components/dashboard/collection_component_spec.rb
+++ b/spec/components/dashboard/collection_component_spec.rb
@@ -122,5 +122,9 @@ RSpec.describe Dashboard::CollectionComponent, type: :component do
     it "renders a link to purl" do
       expect(rendered.css("a").map { |node| node["href"] }).to include "https://purl.stanford.edu/yq268qt4607"
     end
+
+    it "renders a copy link button with a unique aria-label" do
+      expect(rendered.css("a").map { |node| node["aria-label"] }).to include "Copy persistent link to #{WorkTitlePresenter.show(work.head)}"
+    end
   end
 end
--- a/app/views/dashboards/_collection_summary_row.html.erb
+++ b/app/views/dashboards/_collection_summary_row.html.erb
@@ -10,7 +10,7 @@
   <td>
     <% if work.purl %>
       <%= link_to work.purl, work.purl %>
-      <a class="copy-button" data-controller="copy" data-copy-clip-value="<%= work.purl %>" data-action="copy#copy" aria-label="Copy persistent link" href=""><span class="fa-solid fa-copy"></span></a>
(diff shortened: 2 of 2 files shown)


spec/components/dashboard/collection_component_spec.rb
```diff
@@ -122,5 +122,9 @@
     it "renders a link to purl" do
       expect(rendered.css("a").map { |node| node["href"] }).to include "https://purl.stanford.edu/yq268qt4607"
     end
+
+    it "Copy icon for purl link has aria-label with work title" do
```
mjgiarlo, line 126: Would you consider the more English-y

```suggestion
    it "includes an aria-label with work title" do
```

❓

ndushay: well,
- the block context is `with a work that has a druid` 
- the aria label is for the cute little copy icon, not the purl link

which is how I arrived at the verbose description.

mjgiarlo: OK, then how about:

```suggestion
    it "renders a copy icon link with an informative aria-label" do
```

Trying to lean into RSpec description best practice to avoid gobbledygook output.

ndushay: it's not the purl link that has the aria-label.  There is a copy icon image that has the link.

ndushay: I reworded slightly.

mjgiarlo: Do you want to come up with a concise English phrase for this test that is of the form "verbs a noun"?

mjgiarlo (approved):

PR #3227, opened 2023-07-11
Improve performance of collection deposits page

### LINKED ISSUE #3218: Spike on performance of collection display page
While investigating #3195 it was discovered that ActiveStorage is queried for all files in each object of a collection (presumably to provide the file size sum) which causes a significant performance impact with there are a lot of files involved.

This will identify where and how that query (or queries) are being executed and more performant ways to address this concern.

aaron-collier:
# Why was this change made? 🤔

Closes #3218 

These 2 changes improve the performance of how we are fetching the data displayed on the collection deposits page. The main change shows a direct SQL query to avoid preforming multiple queries per file to get the size (this becomes incredibly inefficient with a large number of files). The direct query improves the result time that was several minutes into a second or less.

### DIFF
--- a/app/components/collections/work_row_component.html.erb
+++ b/app/components/collections/work_row_component.html.erb
@@ -7,7 +7,7 @@
   <td><%= work.owner.sunetid %></td>
   <td><%= render Works::StateDisplayComponent.new(work_version: work_version) %></td>
   <td><%= render LocalTimeComponent.new(datetime: work.updated_at, show_time: false) %></td>
-  <td><%= attached_files.count %></td>
+  <td><%= attached_files.size %></td>
   <td><%= size %></td>
   <td><%= link_to work.purl, work.purl if work.purl %></td>
   <td><%= render CitationComponent.new(work_version: work_version) %></td>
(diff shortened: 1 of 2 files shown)


app/components/collections/work_row_component.html.erb
```diff
@@ -7,7 +7,7 @@
   <td><%= work.owner.sunetid %></td>
   <td><%= render Works::StateDisplayComponent.new(work_version: work_version) %></td>
   <td><%= render LocalTimeComponent.new(datetime: work.updated_at, show_time: false) %></td>
-  <td><%= attached_files.count %></td>
+  <td><%= attached_files.size %></td>
```
aaron-collier, line 10: This is based on an article I found about the performance of `count` vs. `size`: https://www.speedshop.co/2019/01/10/three-activerecord-mistakes.html. This can be confirmed locally by creating a work with a lot of files (I tested with 25K) and manually performing the `count` query, you will see a slow query of all objects in order to get the count of records vs. `size` not performing a query against all records.

justinlittman (approved):

PR #3229, opened 2023-07-11
Do not render globus instructions until globus endpoint is set.

### LINKED ISSUE #3212: Handle globus endpoint folder link before full path is available
Because the globus endpoint folder path (e.g. `/lwrubel/work123/version1`) is created async via a job, the full endpoint is not available for creating the "shared folder at Globus" link in the globus instructions on the item view page after a work is first saved. This means that the user may first get a link that goes to a truncated path at the globus endpoint (e.g. `https://app.globus.org/file-manager?&destination_id=e32f7087-d32d-4588-8517-b2d0d32d53b8&destination_path=/uploads/`) until it appears via the `after_save` callback reloading the page. 

Screenshot 2023-07-10 at 8 23 10 AM

Need a design for what should display instead of a truncated link until the page reloads with the full endpoint path available. Currently the link to the folder is text, so it's hard to insert a Loading... spinner in the middle of the text.

justinlittman:
closes #3212

# Why was this change made? 🤔
User was getting an incomplete link.

Unit

### DIFF
--- a/app/components/works/globus_setup_component.rb
+++ b/app/components/works/globus_setup_component.rb
@@ -10,7 +10,11 @@ module Works
     attr_reader :work_version
 
     def render?
-      work_version.globus? && work_version.globus_endpoint
+      work_version.globus?
+    end
+
+    def placeholder?
+      work_version.globus_endpoint.blank?
     end
 
     def globus_user_valid?
--- a/app/components/works/globus_setup_component.rb
+++ b/app/components/works/globus_setup_component.rb
@@ -10,7 +10,7 @@ module Works
     attr_reader :work_version
 
     def render?
-      work_version.globus?
+      work_version.globus? && work_version.globus_endpoint
     end
 
     def globus_user_valid?
--- a/app/components/works/globus_setup_component.html.erb
+++ b/app/components/works/globus_setup_component.html.erb
@@ -2,16 +2,26 @@
   <div class="row my-5 mx-2 alert alert-danger">
     <div class="col-12">
       <p><strong>How to complete your deposit using Globus:</strong></p>
-      <ol>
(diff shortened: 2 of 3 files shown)


lwrubel:
There can be a few seconds while waiting for the instructions to be ready to display. This seems a little confusing? Would want @amyehodge or @astridu's to weigh in. 

https://github.com/sul-dlss/happy-heron/assets/1619369/10c00087-c478-48ff-b909-316087173a79

@justinlittman this approach you took is very simple, which I like. Did you also consider a placeholder for the instructions box and decide that was too complicated?

astridu:
Sorry, I wasn't clear. I meant that you could have the pink box w/title show, and then the entire block of text below it would be in that 'skeleton' (the gray boxes pulsing) until the link was available.

justinlittman:
I'd suggest that it is equally confusing to have a late appearing instruction box as it is to have a placeholder. (Why is there a placeholder for instruction text?)

A third option is to have an alert that states something like "Setting up globus  ... [spinner]" that is then replaced with the instructions. If I'm not mistaken, this is similar to what we do with other async activities like unzipping.

astridu:
I don't think it's confusing. It's just loading. It's just a few seconds. It was much more jarring to see the pink box appear all of a sudden when I wasn't expecting it. Having the box there with the skeleton gives the user some idea for what to expect.

app/components/works/globus_setup_component.rb
```diff
@@ -10,7 +10,7 @@ def initialize(work_version:)
     attr_reader :work_version
 
     def render?
-      work_version.globus?
+      work_version.globus? && work_version.globus_endpoint
```
mjgiarlo, line 13: Wondering about adding a `.present?` here. This field defaults to `nil` in the DB, I assume, and thus any non-nil value is fine here? We can't imagine this value being an empty string, right? If so, this is probably fine as is.

mjgiarlo (approved):

justinlittman:
How is this @astridu ?

image

astridu:
Thank you. Looks great!

mjgiarlo (approved):

PR #3245, opened 2023-07-13
Strip whitespace

### LINKED ISSUE #3207: Trim leading and trailing white space from free text entry fields
@justinlittman and I noticed that a number of organization entries had leading and/or trailing white space. These are likely present for all free text entry fields and should be trimmed.

Fields that should have leading and trailing white space trimming:
For item form:
- Title of deposit
- contact email
- first name
- last name
- abstract
- keywords
- citation
- related works
- link text
- URL (not sure if this needs it?)

For collection form:
- collection name
- description
- contact email
- link text
- URL (not sure if this needs it?)

-

edsu:
# Why was this change made? 🤔

Adds the [strip-attributes](https://github.com/rmm5t/strip_attributes) gem to easily remove leading/trailing whitespace from designated model attributes.

I chose to configure strip-attributes to only strip specific identified attributes, and also turned off the auto-conversion of `""` to `nil` which broke some tests because the columns were `NOT NULL`.

Fixes #3207

Local development instance and unit tests.

No

### DIFF
--- a/spec/models/contact_email_spec.rb
+++ b/spec/models/contact_email_spec.rb
@@ -14,4 +14,14 @@ RSpec.describe ContactEmail do
   it "belongs to a work" do
     expect(contact_email.emailable).to eq work
   end
+
+  context "when select attributes contain leading/trailing whitespace" do
+    let(:contact_email) do
+      create(:contact_email, email: " timbl@w3.org ", emailable: work)
+    end
+
+    it("strips email") do
+      expect(contact_email.email).to eq("timbl@w3.org")
+    end
+  end
 end
--- a/spec/models/keyword_spec.rb
+++ b/spec/models/keyword_spec.rb
@@ -12,4 +12,24 @@ RSpec.describe Keyword do
   it "has a url" do
     expect(keyword.uri).to be_present
   end
+
+  context "when select attributes contain leading/trailing whitespace" do
+    let(:work_version) { build(:work_version) }
+    let(:keyword) do
+      create(
+        :keyword,
+        label: " World Wide Web ",
+        uri: " http://www.wikidata.org/entity/Q466 ",
+        work_version:
+      )
+    end
+
+    it "strips label" do
+      expect(keyword.label).to eq("World Wide Web")
+    end
+
+    it "strips uri" do
+      expect(keyword.uri).to eq("http://www.wikidata.org/entity/Q466")
+    end
+  end
 end
--- a/spec/models/related_work_spec.rb
+++ b/spec/models/related_work_spec.rb
@@ -8,4 +8,14 @@ RSpec.describe RelatedWork do
   it "has a citation" do
     expect(related_work.citation).to be_present
   end
+
+  context "when select attributes contain leading/trailing whitespace" do
+    let(:related_work) do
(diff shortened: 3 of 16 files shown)


mjgiarlo (approved):
LGTM!

spec/models/contact_email_spec.rb
```diff
@@ -14,4 +14,18 @@
...
+  describe "whitespace" do
+    let(:contact_email) do
+      create(
+        :contact_email,
+        email: " timbl@w3.org ",
+        emailable: work
+      )
+    end
+
+    it("strips email") do
```
mjgiarlo, line 27: I think you can drop the parens here but NBD

spec/models/keyword_spec.rb
```diff
@@ -5,11 +5,32 @@
 RSpec.describe Keyword do
   subject(:keyword) { build(:keyword) }
 
+  let(:work_version) { build(:work_version) }
```
mjgiarlo, line 8: If `work_version` is only used in the whitespace tests, you could consider moving the `let` statement down into the relevant `describe` block.

spec/models/related_work_spec.rb
```diff
@@ -8,4 +8,14 @@
   it "has a citation" do
     expect(related_work.citation).to be_present
   end
+
+  describe "whitespace" do
+    let(:related_work) do
+      create(:related_work, citation: " The RTF file generated from the above, with scalable drawings ")
+    end
+
+    it "stripped citation" do
```
mjgiarlo, line 17: `s/stripped/strips/` ?

spec/models/work_version_spec.rb
```diff
@@ -330,4 +330,27 @@
...
+      expect(work_version.title).to eq("Information Management: A Proposal")
+    end
+
+    it "stripped the abstract" do
+      expect(work_version.abstract).to eq("This proposal concerns the management of general information about accelerators and experiments at CERN. It discusses the problems of loss of information about complex evolving systems and derives a solution based on a distributed hypertext system.")
+    end
+
+    it "stripped the citation" do
+      expect(work_version.citation).to eq("Berners-Lee, Tim (1998). Information Management: A Proposal, CERN.")
+    end
```
mjgiarlo, line 354: `s/stripped/strips/g` ?

spec/models/work_version_spec.rb
```diff
@@ -330,4 +330,27 @@
       end
     end
   end
+
+  describe "whitespace" do
```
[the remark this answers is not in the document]
edsu: I felt like I was doing it wrong here, and was hoping a reviewer would suggest the right way to do it -- thanks!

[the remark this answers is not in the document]
mjgiarlo: OK, good!

spec/models/contact_email_spec.rb
edsu: I switched it to use no parens on one line and now I get a rubocop warning:

```
spec/models/contact_email_spec.rb:20:7: C: [Correctable] FactoryBot/ConsistentParenthesesStyle: Prefer method call with parentheses
```

mjgiarlo (approved):

PR #3248, opened 2023-07-13
Avoid race condition by moving job enqueuing out of state machine

### LINKED ISSUE #3071: object not in 'first_draft' state after fetch_globus_job completed
~~During the week of May 15, we saw two unexplained issues:~~ (moved first issue to it's own ticket #3134)

HB error: https://app.honeybadger.io/projects/77112/faults/96031287

```
    "backtrace": [
      {
        "number": "224",
        "file": "[GEM_ROOT]/gems/state_machines-0.5.0/lib/state_machines/event.rb",
        "method": "block in add_actions",
        "source": {
          "222": "      # Fires the event, raising an exception if it fails\n",
          "223": "      machine.define_helper(:instance, \"#{qualified_name}!\") do |machine, object, *args|\n",
          "224": "        object.send(qualified_name, *args) || raise(StateMachines::InvalidTransition.new(object, machine, name))\n",
          "225": "      end\n",
          "226": "    end\n"
        }
      },
 snip...
    "component": "FetchGlobusJob",
    "action": "perform",
snip ...
    "message": "StateMachines::InvalidTransition: Cannot transition state via :fetch_globus_complete from :first_draft (Reason(s): State cannot transition via \"fetch globus complete\")",
snip ...
    "created_at": "2023-06-08T19:21:02.267999Z",
    "causes": [
      {
        "class": "Sidekiq::JobRetry::Skip",
        "message": "Sidekiq::JobRetry::Skip",
        "backtrace": [
...
        ]
      }
    ]

```

Work ID 6598 (first draft) was a single ZIP file added via globus.  It triggered this HB alert from `app/jobs/fetch_globus_job.rb`: https://app.honeybadger.io/projects/77112/faults/96031287  

The HB alert suggests the job completed, but the object was not in the correct state (`fetch_globus_first_draft`) to allow the transition back to `first_draft` to occur.    

The code suggests that the work_version should go into the `fetch_globus_first_draft` state before the job is run, and then the job is completed, it should go back to the `first_draft` state via the `fetch_globus_complete` transition in `app/jobs/fetch_globus_job.rb`.  

Running the job again by initiating the event `fetch_globus` allowed the job to run successfully with no HB alerts, so it's not clear why one of the transitions didn't work the first time.  May be related to #3005

lwrubel:
# Why was this change made? 🤔
Resolves #3071. Both FetchGlobusJob and UnzipJob, which send an event at the end, had been called from an `after_transition` callback. Moving that step out of the state machine, to the controller. 

Unit tests and QA testing. 

No front-end changes

### DIFF
--- a/app/controllers/works_controller.rb
+++ b/app/controllers/works_controller.rb
@@ -183,7 +183,11 @@ class WorksController < ObjectsController
 
     # Instead of crazy conditional logic, constructing the event name and sending
     state_event = state_event_for(work_version)
-    work_version.send(state_event) if state_event
+    if state_event
+      work_version.send(state_event)
+      FetchGlobusJob.perform_later(work_version) if state_event.starts_with?("fetch_globus")
+      UnzipJob.perform_later(work_version) if state_event.starts_with?("unzip")
+    end
 
     return redirect_to next_step_review_work_path(work) if deposit_button_pushed? && work.collection.review_enabled?
     return redirect_to next_step_work_path(work) if deposit_button_pushed?
--- a/app/models/concerns/work_version_state_machine.rb
+++ b/app/models/concerns/work_version_state_machine.rb
@@ -26,12 +26,6 @@ module WorkVersionStateMachine
       after_transition on: :deposit_complete, do: WorkObserver.method(:after_deposit_complete)
       after_transition on: :deposit_complete, do: CollectionObserver.method(:item_deposited)
       after_transition on: :decommission, do: WorkObserver.method(:after_decommission)
-      after_transition on: :unzip, do: WorkObserver.method(:after_unzip)
-      after_transition on: :unzip_and_submit_for_review, do: WorkObserver.method(:after_unzip)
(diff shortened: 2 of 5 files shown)


app/controllers/works_controller.rb
```diff
@@ -183,7 +183,11 @@ def after_save(form:, event_context:)
 
     # Instead of crazy conditional logic, constructing the event name and sending
     state_event = state_event_for(work_version)
-    work_version.send(state_event) if state_event
+    if state_event
+      work_version.send(state_event)
+      FetchGlobusJob.perform_later(work_version) if state_event.starts_with?("fetch_globus")
+      UnzipJob.perform_later(work_version) if state_event.starts_with?("unzip")
+    end
```
lwrubel, line 190: With the job enqueued here, I haven't been able to trigger any invalid transition errors. I think it *might* still be possible for the job to finish and send a job completed event before the state machine has saved the new state in the database, but I haven't seen that happen yet. Open to other ideas about how to do this.

mjgiarlo (approved):
LGTM, thanks for untangling this!

PR #3257, opened 2023-07-17
Add role term type radio buttons

### LINKED ISSUE #3232: Require user to choose individual or organizational role category and then display only terms for that category
This change will help make sure users are choosing a role term that is appropriate to the type of entity that they need to enter as an author or contributor. We've noted that some users are choosing roles like "author" and "contributing author" from the organizational list when they should have chosen them from the individual list. 

Require users to choose either individual or organization as the role category from a drop-down list, then in the next drop-down list, display only the role terms available for the category chosen.

The list defaults should be "individual" and then "author".

peetucket:
# Why was this change made? 🤔

Fixes #3232 - add radio button toggles for role term type for contributors to avoid users choosing the wrong contributor type.  Selecting a radio button changes the options shown in the drop down menu.  Also, adjust layout to match designs shown here: https://github.com/sul-dlss/happy-heron/issues/3232#issuecomment-1631553241.  Finally, refactor to remove the "individual|" and "organization|" prefix in the code for the drop down menus, since we can set that directly now in the form via the radio buttons.

https://github.com/sul-dlss/happy-heron/assets/47137/5130cdbe-8403-47f8-826a-64c46b16e1c0

Localhost and CI


app/components/works/contributor_role_component.rb
```diff
@@ -3,18 +3,20 @@
 module Works
   # Renders a widget for selecting a contributor role
   class ContributorRoleComponent < ApplicationComponent
-    def initialize(form:, data_options:)
+    def initialize(form:, data_options:, contributor_type:, visible:)
       @form = form
+      @contributor_type = contributor_type # person or organization
+      @visible = visible # if drop down starts out as hidden/disabled
       @data_options = data_options
```
peetucket, line 10: We now render two drop down menus for the contribute role type, one for "person" and one for "organization".  Only one starts out as visible depending on the current state of the entered data.

app/components/works/contributor_role_component.rb
```diff
@@ -52,8 +54,8 @@ def key
 
     # list for work_form pulldown
 
-    def grouped_options
-      %w[person organization].map do |key|
+    def grouped_options(contributor_type)
+      [contributor_type].map do |key|
         ContributorType.new(key:, citable: false)
       end
```
peetucket, line 60: Allows us to render just type of drop down at a time.  Yes, it is still a grouped select, with just a single group.  I like the way this displays though (it reinforces the "Individual" vs "Organization" header).

app/components/works/contributor_row_component.rb
```diff
@@ -62,6 +70,14 @@ def orcid?
...
     end
 
+    def person?
+      model.contributor_type == "person" || model.contributor_type.blank?
+    end
+
+    def organization?
+      model.contributor_type == "organization"
+    end
+
```
peetucket, line 80: let's us figure out if the model object represents a person or organization (default to person for a new row)

app/components/works/contributor_row_component.rb
```diff
@@ -113,17 +129,14 @@ def with_required(label)
...
+        contributors_target: "select#{contributor_type.titlecase}Role"
       }.tap do |opts|
-        actions = ["change->contributors#roleChanged"]
         if author?
+          opts[:action] = "change->auto-citation#updateDisplay"
           opts[:auto_citation_target] = "contributorRole"
-          actions << ["change->auto-citation#updateDisplay"]
         end
-        opts[:action] = actions.join(" ")
       end
```
peetucket, line 140: this method gets a bit simpler now actually

app/javascript/controllers/auto_citation_controller.js
```diff
@@ -60,14 +60,18 @@ export default class extends Controller {
...
   }
 
   get contributorRoles() {
     return this.contributorRoleTargets.filter(elem => elem.disabled == false)
   }
 
+  get contributorOrgNames() {
+    return this.contributorOrgTargets.filter(elem => elem.disabled == false)
+  }
+
```
peetucket, line 74: this ensures we only find organization names that are not disabled when building the citation...this mimics what we do for names 

i think we should have had this change before this work too, maybe it still worked with JS errors :shrug: but this normalizes it

app/forms/contributor_form.rb
```diff
@@ -34,7 +34,8 @@ def form_properties(validate:)
         property :last_name
         property :full_name
         property :orcid
-        property :role_term
+        property :role
+        property :contributor_type
         property :_destroy, virtual: true, type: Dry::Types["params.nil"] | Dry::Types["params.bool"]
```
peetucket, line 39: We now have a radio button to tell us the type of contributor, so we do not need the `role_term` construct, which had prepended the type to the role previously.

app/javascript/controllers/auto_citation_controller.js
```diff
@@ -55,19 +55,23 @@ export default class extends Controller {
   // Authors (person and organization) as an array of strings.
   get authors() {
     return this.contributorRoles.map((roleField, index) => {
-      if (roleField.value.startsWith('person')) {
+      if (roleField.attributes['data-contributors-target'].value == 'selectPersonRole') {
         const firstInitial = `${this.contributorFirsts[index].value.charAt(0)}.`
```
peetucket, line 59: This is how we know we are dealing with a person and not an organization (since we don't have that prepended "person|" stuff anymore

app/models/abstract_contributor.rb
```diff
@@ -78,17 +76,4 @@ def self.grouped_roles(citable:)
...
-    [contributor_type, role].join(SEPARATOR)
-  end
-
-  # used by DraftWorkForm for setting values from the form.
-
-  def role_term=(val)
-    contributor_type, role = val.split(SEPARATOR)
-    self.attributes = {contributor_type:, role:}
-  end
 end
```
peetucket, line 79: don't need this abstraction anymore, we can set the `role` and `contributor_type` directly from the drop down menu and radio buttons

aaron-collier (approved):

PR #3258, opened 2023-07-17
fix tou icons not showing sometimes

### LINKED ISSUE #3255: Terms of deposit icons not showing correctly (sometimes)
An accessibility violation was noted as animated icons, these animations are due to a broken fa icon, and they should not be.

We should figure out why they are sometimes showing as broken.  I have not found a pattern, sometimes first page load shows them as broken, refresh fixes them.  Doesn't seem to matter what page you are on.  Could have something to do with the terms being loaded in a turbo frame on demand.

**Broken:**

Screen Shot 2023-07-17 at 1 29 04 PM

**What it should look like:**

Screen Shot 2023-07-17 at 2 07 22 PM

FAQ = should be question mark icon
Save = should be disk icon
Print = should be printer icon

peetucket:
# Why was this change made? 🤔

Fixes #3255 (at least I can't reproduce anymore on localhost after making this change)

Localhost

### DIFF
--- a/app/views/shared/_terms_of_deposit.html.erb
+++ b/app/views/shared/_terms_of_deposit.html.erb
@@ -7,17 +7,17 @@
           </div>
           <div class="col-2">
             <%= link_to Settings.faq_url, aria: {label: "FAQs about Terms of Deposit"}, id: "faq-link", class: "text-decoration-none", target: "_blank" do %>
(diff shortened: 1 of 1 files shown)


lwrubel (approved):
I am curious how this tweak has the effect of making the icons always appear, but I understand this has been hard to test. It's not going to make anything worse, so approving.

PR #3269, opened 2023-07-18
Adds details to affiliation autosuggest.

### LINKED ISSUE #3236: Add display of additional details for affiliation term matches
For the affiliation type-ahead search, display additional information along with the name matches to help users choose the right entity.

ROR's recommendations: https://ror.readme.io/docs/create-ror-powered-typeaheads#show-information-beyond-the-primary-organization-name

These include displaying in addition to the primary name:

- addresses.geonames_city.city (City name)
- coutnry.country_name (country)
- organization type (I don't think we need this)
- acronyms
- aliases
- labels (these are names in other languages)

I'm concerned that this might get kind of a lot to display, but I don't know what the data look like. 

The ROR example shows it laid out like this:

justinlittman:
closes #3236

# Why was this change made? 🤔
More context for users.

Unit, local

image

### DIFF
--- a/app/controllers/ror_controller.rb
+++ b/app/controllers/ror_controller.rb
@@ -33,8 +33,33 @@ class RorController < ApplicationController
 
   def parse(body)
     JSON.parse(body)["items"].map do |item|
-      [item["id"], item["name"]]
+      [item["id"], item["name"], details(item)]
     end
+  end
+
+  def details(item)
+    [
+      item["name"],
+      location_detail(item),
+      other_name_details(item)
+    ].compact
+  end
+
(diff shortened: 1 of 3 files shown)


peetucket:
Even though we have now hidden the affiliations UI component, I'd vote for moving this PR forward anyway, so that when we return to adding the view back in bug free, it will be available.

Can I suggest you rebase from main and let the tests run again to be sure it works with the UI component not there (no reason to suspect it won't but just double check).

justinlittman:
@peetucket Rebased and green.

peetucket (approved):

Issue #3094, opened 2023-06-13
Allow depositors to enter a custom rights statement


peetucket:
IMPORTANT NOTE: This work should be based on the branch in #3271 and end as a PR into #3271

Background:
- Currently all H2 objects receive the same Terms of Use statement, which is hard-coded in the `settings.yml file` in the key `access:use_and_reproduction_statement`.  
- This statement is then added to cocina so that it can be shown on PURL pages.  
- Some publishers require authors to add specific text to their deposit in order to comply with licensing requirements. Read more here: https://docs.google.com/document/d/18RGicHs5zvFU7KLDt29DKFvFvvjZnfAQagi8q57vExk/edit#heading=h.4irp318p0ni3)
- This requirement thus allows depositors to add a custom statement which is prepended (does NOT replace) to the current hard-coded statement.  
- This is optional (i.e. many depositors will continue to just have the hard-coded statement).
- This ability is only available on works in collections for which the collection manager has enabled it: see #3097

Requirements:
- Free text entry field to support several sentences/a paragraph. 
- Field label: "Additional Terms of Use"
- Default instructions: "For open access articles, use this section for required publisher's statement. For other deposits, use this section to describe other terms of use not covered by the license you select below. This section does not constitute a formal Data Use Agreement."
- Include this new field in the "Select a license" section of the form, below the license selection.
- Re-title this section of the form "Terms of use."
- Field should be editable on later versions like most of the other fields on the form.
- Existing Use and Reproduction statement (see above) will continue to be included on all items and does not need to be alterable when a custom statement is entered.  Any statement entered by a user here is pre-pended when adding to cocina.
- Entry of content in this field will in no way alter requirements for end users to access the files, i.e. this will not change file or object rights and will not require end users to agree to anything. This is not a formal data use agreement.
- Items with these entries will not trigger a requirement for review.

Design: https://projects.invisionapp.com/share/NP138M0QX7T3#/screens/473365681_custom_Rights_Statement
Requirements: https://docs.google.com/document/d/1eWkkCIbdxO4UGT-YQoMCx-O3m_Bo3UaEaDj8OVkcTc0/edit#

jmartin-sul:
part of https://github.com/sul-dlss/happy-heron/issues/3093

Issue #3220, opened 2023-07-11
Add text to instructions for zip file upload


amyehodge:
Amend the subtext for the zip file upload to add:

Unzipping may take several minutes if your zip file contains a large number of individual files.

So, the full line will read:

Viewers will see the file hierarchy and be able to download individual files. Unzipping may take several minutes if your zip file contains a large number of individual files.

amyehodge:
Please let @astridu know if you need a design mockup for this.

PR #3271, opened 2023-07-18
collection managers can allow custom rights statements, depositors can enter custom rights statements if a collection allows it

### LINKED ISSUE #3097: Allow collection managers to indicate if a custom rights statement can be set
See #3094 for the ability for depositors to set custom rights statements in the works.

To prevent abuse of this option, we will make this an option that can be turned on for a collection by the collection manager. 

Collection managers should be able to:

- Turn on this option when they want it available for items in their collection.
- Require that specific terms always be used for items in the collection or allow users to provide their own terms.
- Present default instructions for the user on what to enter in this field or allow the collection manager to provide their own instructions that would appear for items in the collection.

Design: https://projects.invisionapp.com/share/NP138M0QX7T3#/screens/473602644_custom_Statement_Y
Requirements: https://docs.google.com/document/d/1eWkkCIbdxO4UGT-YQoMCx-O3m_Bo3UaEaDj8OVkcTc0/edit#

jmartin-sul:
# Why was this change made? 🤔

close #3097

~~[HOLD] until #3094, #3293 #3291 are completed.  Then rebase and merge.~~ those are done, branch has been rebased

~~[HOLD] for Amy's review~~ Reviewed

_**NOTE**_ -- most of the below testing info and review context was written, and demo video was recorded, before this turned into a feature branch where follow-on work would be held till things are complete enough to merge to `main`.  will refresh/update the below as needed once the above mentioned tickets are completed and their PRs are merged into this branch, and will ask for one last once over before merge to `main`.

manually in local dev environment, unit tests, reviewed in stage by PO

## video running through the basic interactions (_of only the collection settings work this PR was originally for_).  it demonstrates:

_other follow-on PRs have screenshots of additional work for this feature_

* the dynamic `fieldset` hiding/showing
* blanking out no longer applicable fields when saving (e.g. blanking out custom instructions when switching back to default instructions)
* saving chosen options and reloading them on the edit page
* not saving on cancel

https://github.com/sul-dlss/happy-heron/assets/7741604/cf1c9240-2d40-4b22-a82f-164312ab7f61

This video only demos using a draft collection, but in my local testing things also worked with deposited updates.  That's a bit more circuitous and slow to test on localhost (because you have to switch back to terminal to run `rake complete_deposits`), which would've led to an even longer demo video.

none AFAICT
* no regressions according to existing a11y tests in test suite
* fixed issues found by firefox accessibility inspector -- right click on new form section -> `Inspect Accessibility Properties`
  * no issues with new section found after that by SiteImprove browser plugin

_note_: checked for accessibility regressions where it seemed appropriate when creating/reviewing the follow-on PRs that got merged into this feature branch.  once this is merged, i'll do another accessibility once-over for the feature in general (create/show custom rights settings for collection, edit/show custom rights for work creation editing).  i'll file tickets for any issues that are discovered in that testing.  the above blurb about testing only applies to what this PR was first for, editing collection settings wrt custom righs, not for the feature branch as a whole (which is what this PR turned into).

# other notes
* Though not in the mockups/requirements, I added the `Choose the source of the custom use statement` to deal with an accessibility complaint about that `fieldset` being unlabelled.
* ~~For the default instructions, I used the text from the mockups, which I later realized differs from the text in the requirements doc.  Happy to change that if desired, it's a setting, so that's easy to edit.~~ Discussed in planning meeting and got the appropriate text for the default instructions as well as the popover that can display in the custom terms section of the work create/edit page.
* Mostly leaving this bullet point for historical context since we're probably sticking with the current data model, but here's an in the weeds data model question we'd discussed on Thurs 2023-07-20... A change suggested by @justinlittman when I was first running into unit test trouble was to model this in a slightly different way.  Instead of using a boolean `allow_custom_rights_statement` field, and determining where the custom rights statement comes from based on a combo of that and the presence of `provided_custom_rights_statement` (see `Collection#custom_rights_statement_source_option`), he'd suggested something like a `custom_rights_statement_option` enum, with values like `no_custom`, `provided_by_collection`, and `entered_by_depositor`.  That should let us trade the `allow_custom_rights_statement?` and `custom_rights_statement_source_option` methods on the model for one `custom_rights_statement_option` method that maps directly to the database field.  That might then make for a slightly more understandable `Collection` class, and I think would also eliminate the use of Reform `virtual` fields from this PR.  Which might make the model and reform code slightly more maintainable.  The one downside I can see is that it'd make the form a bit weird, in that i think it'd lead to three radio buttons (`No`, `Provide specific terms`, and `Allow depositor to enter their own terms`) across two different fieldsets being mapped to one option field, and it'd make the `Yes` radio button solely a UI thing with no real form submission or collection update effect.  When the intuition might be that the `No` and `Yes` radio buttons go together for one field (as it is right now).  Not sure which approach would be easier to maintain overall, but the real reason I didn't end up trying out that data model change was that I spent more time than planned yesterday figuring out accessibility stuff (mostly association of a label with both a radio button and a text area), and we'd already agreed in standup that I should timebox the data model change, since this ticket is blocking other work needed to complete this feature.  If people decide that data model change is worth it, we should probably do it in this PR, so that subsequent work that builds on this doesn't need changing too.  I'd be slightly in favor of just going with this data model at this point, but happy to do otherwise if that's the consensus.
* requirements doc says: _"Field should support hyperlinks, i.e. a fully written out hyperlink in this box should work as a hyperlink on the purl page."_  That seemed like more of a display-time issue, so nothing in this PR addresses that req, but something to consider when reviewing.


spec/components/collections/edit_license_component_spec.rb
```diff
@@ -33,6 +33,71 @@
...
+      let(:collection) { create(:collection, :with_custom_rights_from_collection) }
+      let(:collection_version) { create(:collection_version, collection:) }
+
+      before { collection_version.collection.head = collection_version }
+
+      # TODO: this hack didn't work.  should get this through model methods acting on factorybot defined fields anyway.
+      # before { allow(collection).to receive(:custom_rights_statement_source_option).and_return("provided_by_collection") }
+
+      it "selects the 'Provide specific terms' option" do
+        # byebug
```
jmartin-sul, line 66: Here's what I find most perplexing at the moment about this test: if I uncomment this `byebug` call, and drop into the debugger, I see something like:
```
(byebug) collection.head == collection_version
true
(byebug) collection_form.collection == collection
true
(byebug) collection.custom_rights_statement_source_option
"provided_by_collection"
(byebug) collection_form.collection.custom_rights_statement_source_option
"provided_by_collection"
(byebug) collection&.custom_rights_statement_source_option
"provided_by_collection"
(byebug) 
```

The thing i find perplexing there is that `collection.custom_rights_statement_source_option` is `"provided_by_collection"` (what I'd expect given the fixture setup and the logic of `Collection#custom_rights_statement_source_option`).  And all the other associations seem correct.  But then it would seem that `EditLicenseComponent` isn't getting that value from the `Collection` object in the test or something, because the radio button that's examined by this test isn't being selected according to the value in the collection.  But it does work in the actual form when I test manually by creating or editing a collection in my local H2 instance.

When I was asking for debugging advice on this before last week's FR shift, @justinlittman suspected that the thing that was breaking the existing unit tests was that the `custom_rights_statement_source_option` property in `CollectionSettingsForm` was `writeable: true`.  Changing it to be `virtual: true` as it is now in this branch (i.e. not writeable or readable from the model, [see here](https://trailblazer.to/2.1/docs/reform.html#virtual)) unbroke existing tests, but it does seem fishy to me that the misbehaving property here is `virtual`, and that the tests for values in `textarea` elements (which are not virtual and come directly from the model) are passing.  One of Justin's suggestions was a re-thinking of the data model so that there's an actual `custom_rights_statement_option` field with values like `no_custom`, `provided_by_collection`, and `entered_by_depositor` (instead of the current `allow_custom_rights_statement` boolean and some logic acting on that and the presence of e.g. custom instructions text).  I initially went for the half-measure seen in this branch, where I kept the boolean but got rid of `writeable`, because I figured it'd be a more expedient way to get what I had working, and then I could revisit the data model change once I had the tests working with something closer to the approach I'd already started on.  But now I think I'll just try taking that suggestion in full.

If I follow that path, I think it'll mean that the "Yes" radio button is still a virtual property in the Reform code, and so maybe it'll be hard to test, but all of the other radio buttons should then become more testable, if that approach pays off.

app/components/collections/edit_license_component.html.erb
```diff
@@ -38,4 +38,71 @@
...
+          <%= form.radio_button :allow_custom_rights_statement, true, class: "form-check-input", data: {action: "custom-rights#updateChooseCustomRightsFieldsetEnabled", custom_rights_target: "yesIncludeCustomRightsRadioButton"} %>
+          <%= form.label :allow_custom_rights_statement_true, "Yes", class: "form-check-label" %>
+        </div>
+
+        <div class="ms-4">
+          <fieldset data-custom-rights-target="customRightsConfigFieldset">
+            <legend class="fs-6"><strong>Choose the source of the custom use statement</strong></legend>
+            <div class="form-check<%= " is-invalid" if error? %>">
+              <%= form.radio_button :custom_rights_statement_source_option, "provided_by_collection", checked: (custom_rights_statement_source_option == "provided_by_collection"), class: "form-check-input", data: {action: "custom-rights#updateCustomRightsInstructionsFieldsetEnabled", custom_rights_target: "customRightsSourceIsCollectionButton"} %>
+              <%= form.label :custom_rights_statement_source_option_provided_by_collection, "Provide specific terms", class: "form-check-label pe-2",
```
peetucket, line 75:
```suggestion
              <%= form.label :custom_rights_statement_source_option_provided_by_collection, "Provide specific terms", class: "form-check-label pe-2",  style: "vertical-align: top",
```

peetucket: This ensures the label is aligned next to the radio button (else it will float down because the text box takes more than one line).

app/components/collections/edit_license_component.html.erb
```diff
@@ -38,4 +38,71 @@
...
+                <div class="form-check<%= " is-invalid" if error? %>">
+                  <%= form.radio_button :custom_rights_instructions_source_option, "default_instructions", checked: (custom_rights_instructions_source_option == "default_instructions"), class: "form-check-input" %>
+                  <%= form.label :custom_rights_instructions_source_option_default_instructions, "Yes, display the following instructions", class: "form-check-label" %>
+                  <p class="terms-of-use">
+                    <%= Settings.access.default_instructions_for_custom_use_statement %>
+                  </p>
+                </div>
+                <div class="form-check<%= " is-invalid" if error? %>">
+                  <%= form.radio_button :custom_rights_instructions_source_option, "provided_by_collection", checked: (custom_rights_instructions_source_option == "provided_by_collection"), class: "form-check-input" %>
+                  <%= form.label :custom_rights_instructions_source_option_provided_by_collection, "No, display these instructions", class: "form-check-label pe-2",
```
peetucket, line 97:
```suggestion
                  <%= form.label :custom_rights_instructions_source_option_provided_by_collection, "No, display these instructions", class: "form-check-label pe-2", style: "vertical-align: top",
```

peetucket: Same comment as above - ensure label aligns to radio button

peetucket (changes requested):
1. Couple suggested CSS edits.
2. I didn't realize there was so much code duplication on the `DraftCollectionForm` and the `CollectionSettingsForm`  ... oh well, nothing to do here

Otherwise it looks good.  Would be nice to have Amy review before merging.  If she spots text chnages (due to the mismatch between mockups and requirements), we can change them.  I created a separate ticket for the display on the collection view page, and agreed, I am not sure there is anything we can do about PURL making hyperlinks here.

peetucket (approved):
LGTM. We may want to have Amy check this out on QA before merging.

jmartin-sul:
@peetucket 💬 
> 2\. I didn't realize there was so much code duplication on the `DraftCollectionForm` and the `CollectionSettingsForm`  ... oh well, nothing to do here

yeah, that's not the best, and while this PR didn't introduce that problem, it did dig us into it slightly more.  but agreed this PR isn't really the occasion to dig us out.  i guess that might've been one more point in favor of the other data modeling approach, is that there'd have been less code to add to the `deserialize!` methods in each of those classes.  but i don't know that it'd have gotten us away from adding any code to those two methods, so probably a very marginal difference between the two approaches in terms of DRYness.  might be worth a tech debt ticket to explore refactoring?  but also, there's not an enormous amount of code in either class, and our form handling approach may drastically change if we ever move away from reform.  so maybe not worth worrying about, or at least not in the near term? 🤷

config/settings.yml
```diff
@@ -95,6 +95,7 @@ notifications:
 
 access:
   use_and_reproduction_statement: User agrees that, where applicable, content will not be used to identify or to otherwise infringe the privacy or confidentiality rights of individuals. Content distributed via the Stanford Digital Repository may be subject to additional license and use restrictions applied by the depositor.
+  default_instructions_for_custom_use_statement: Enter additional terms of use not covered by your chosen license or the default terms shown above, which also display on the PURL page.
```
jmartin-sul, line 98: note to self: probably more appropriate to define this in `config/locales/en.yml`.  see https://github.com/sul-dlss/happy-heron/pull/3309#discussion_r1275576563

jmartin-sul: done in #3220

jmartin-sul:
see also #3295 (ticket for tracking the merge of this feature branch)

PR #3281, opened 2023-07-19
add unique labels to remove buttons for contributors and keywords

### LINKED ISSUE #3181: Deposit form remove button accessibility
There are multiple "Remove" buttons that perform different actions. Form field labels should be unique.

peetucket:
# Why was this change made? 🤔

Fixes #3181 - make two generic delete button aira-labels (keywords and contributors) unique so users can tell what they are deleting.  The other trash can aria-labels all seem to be labelled appropriately as far as I can tell by searching through code.

For authors/contributors, aria-label for trash can will be:
- "Remove blank author" or "Remove blank contributor" for new rows (even if text added)
- "Remove [actual name]" if an existing row that was previously saved

For keywords, aria-label for trash can will be:
- "Remove blank keyword" for new rows (even if text added)
- "Remove [actual keyword]" if an existing row that was previously saved

Localhost and CI

### DIFF
--- a/app/components/works/contributor_row_component.rb
+++ b/app/components/works/contributor_row_component.rb
@@ -66,11 +66,19 @@ module Works
       form.object.class.method_defined?(:model) ? form.object.model : form.object
     end
 
+    def contributor_remove_label
+      "Remove #{contributor_name.empty? ? "blank contributor" : contributor_name}"
+    end
+
+    def contributor_name
+      "#{form.object.first_name} #{form.object.last_name}".strip
+    end
+
     def html_options_for_delete
       {
         type: "button",
         class: "btn btn-sm",
-        aria: {label: "Remove"},
+        aria: {label: contributor_remove_label},
         data: {}.tap do |data|
           actions = ["contributors#remove #{form_controller}#removeAssociation"]
           actions << "auto-citation#updateDisplay" if author?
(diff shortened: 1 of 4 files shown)


app/components/works/contributor_row_component.rb
```diff
@@ -66,11 +66,19 @@ def model
...
     end
 
+    def contributor_remove_label
+      "Remove #{contributor_name.empty? ? "blank contributor" : contributor_name}"
+    end
+
+    def contributor_name
+      "#{form.object.first_name} #{form.object.last_name}".strip
+    end
+
```
peetucket, line 76: this may be different for organization vs author contributors, and so this logic may need to be a bit more invovled

app/components/works/contributor_row_component.rb
```diff
@@ -66,11 +66,19 @@ def model
...
     end
 
+    def contributor_remove_label
+      "Remove #{contributor_name.blank? ? "blank #{model.class.name.downcase}" : contributor_name}"
+    end
+
+    def contributor_name
+      model.full_name.blank? ? "#{model.first_name} #{model.last_name}".strip : model.full_name
+    end
+
```
peetucket, line 76: This logic is needed since the "name" (i.e. display label) of an author contributor could be stored in either the `full_name` or `first_name` and `last_name` model attributes for either a contributor or author (depending on the role I selected). We search for either/or to decide if this are blank entries.

aaron-collier (approved):

PR #3282, opened 2023-07-19
Update logic so that auto generated citation is properly enabled when editing

### LINKED ISSUE #3277: Auto-generated citation option should be on unless user turns it off
The auto-generated citation is on when initially creating an item, but if you save the item as draft or deposit it, the next time you open it the auto-generated citation option is off. It should stay on unless the user turns it off.

Steps to recreate:
1. Create new draft item. Fill in form fields. Note that auto citation is on and is being properly generated.
2. Save as draft.
3. Click pencil icon to update.
4. Note that the auto generated citation is off and that changes to the form fields that are part of the citation are not reflected in the citation.

Same for immediately depositing:
1. Create new draft item. Fill in form fields. Note that auto citation is on and is being properly generated.
2. Deposit item
3. Once deposit is done, click pencil icon to create new version.
4. Note that the auto generated citation is off and that changes to the form fields that are part of the citation are not reflected in the citation.
5. Make change to the title
5. Turn on auto citation
6. Deposit item again.
7. Title change is seen in citation for deposited item, i.e. auto citation being turned on updated this like it should, but 
8. click pencil icon to edit. Note that auto citation is back off again.

aaron-collier:
# Why was this change made? 🤔

Fixes #3277 

This enables and disables the manual citation switch when editing items based on the citation value.

In the first 2 screenshots, you can see that I'm using the manual citation as the title in the citation is different than the title in the work. This disables the "auto" citation so that the manual edit is not lost.

Screenshot 2023-07-19 at 4 34 29 PM
Screenshot 2023-07-19 at 4 36 02 PM

When the citation matches the auto generated citation the switch is enabled properly:

Screenshot 2023-07-19 at 4 37 46 PM

### DIFF
--- a/app/javascript/controllers/auto_citation_controller.js
+++ b/app/javascript/controllers/auto_citation_controller.js
@@ -1,18 +1,21 @@
 import { Controller } from "@hotwired/stimulus";
 
+const PURL_PLACE_HOLDER = ":link will be inserted here automatically when available:"
+const DOI_PLACE_HOLDER = ":DOI will be inserted here automatically when available:"
+
 export default class extends Controller {
   static targets = ["titleField", "manual", "auto", "switch",
     "contributorFirst", "contributorLast", "contributorRole", "contributorOrg",
     "embargoYear", "embargo"];
 
   connect() {
-    this.purl = this.data.get("purl") || ":link will be inserted here automatically when available:" // Use a real purl on a persisted item or a placeholder
+    this.purl = this.data.get("purl") || PURL_PLACE_HOLDER // Use a real purl on a persisted item or a placeholder
     this.doi = this.data.get("doi") || ""
 
     this.updateDisplay()
 
-    // If the manualTarget matches the autoTarget or is blank, then display the auto.
-    const showAuto = this.manualTarget.value === '' || this.manualTarget.value === this.autoTarget.value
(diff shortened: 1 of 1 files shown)


app/javascript/controllers/auto_citation_controller.js
```diff
@@ -1,18 +1,21 @@
...
   connect() {
-    this.purl = this.data.get("purl") || ":link will be inserted here automatically when available:" // Use a real purl on a persisted item or a placeholder
+    this.purl = this.data.get("purl") || PURL_PLACE_HOLDER // Use a real purl on a persisted item or a placeholder
     this.doi = this.data.get("doi") || ""
 
     this.updateDisplay()
 
-    // If the manualTarget matches the autoTarget or is blank, then display the auto.
-    const showAuto = this.manualTarget.value === '' || this.manualTarget.value === this.autoTarget.value
+    // If the manualTarget is blank or the autoTarget matches the citation, then display the auto.
```
aaron-collier, line 17: The bug here is that the manual will never match the auto because the auto has the PURL and DOI inserted while the manual has the place holders. I did not like using a long line here with 2 replaces in a single line so moved this into a method to hopefully make the logic a bit more clear.

peetucket (approved):

PR #3290, opened 2023-07-21
Forms using yaaf

justinlittman:
# Why was this change made? 🤔

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,6 +85,7 @@ gem "sneakers", "~> 2.11" # rabbitMQ background processing
 gem "state_machines-activerecord"
 gem "strip_attributes"
 gem "turbo-rails", "~> 1.0"
-gem "view_component", "~> 2.56.2" # https://github.com/github/view_component/issues/1390
+gem "view_component", "~> 3.5"
 gem "whenever", require: false # Work around https://github.com/javan/whenever/issues/831
+gem "yaaf" # Form objects
 gem "zipline", "~> 1.4"
--- a/app/components/nested_form_component.html.erb
+++ b/app/components/nested_form_component.html.erb
@@ -1,0 +1,29 @@
+<div data-controller="<%= controller %>" data-<%= controller %>-wrapper-selector-value=".<%= controller %>-wrapper">
+  <template data-<%= controller %>-target="template">
+    <%= form.fields_for field, clazz.new, child_index: "NEW_RECORD" do |nested_form| %>
+      <div class="<%= controller %>-wrapper" data-new-record="<%= nested_form.object.main_model.new_record? %>">
+        <%= nested_form.hidden_field :_destroy %>
+        <%= render row.new(form: nested_form, controller: controller, **other_row_params) %>
+      </div>
+    <% end %>
+  </template>
+
+  <%= form.fields_for field, nested_forms do |nested_form| %>
+    <div class="<%= controller %>-wrapper" data-new-record="<%= nested_form.object.main_model.new_record? %>">
+      <%= nested_form.hidden_field :id %>
+      <%# :_destroy must come before nested form given queryselector that controller uses for find it. %>
+      <%= nested_form.hidden_field :_destroy %>
+      <%# nested_form may have a remove button which has remove action for the controller %>
+      <%= render row.new(form: nested_form, controller: controller, **other_row_params) %>
+    </div>
+  <% end %>
+
+  <%# Inserted elements will be injected before that target. %>
+  <div data-<%= controller %>-target="target"></div>
+
+  <% if add_another_button? %>
+    <%= add_another_button %>
+  <% else %>
+    <%= form.add_another_button controller, "+ Add another", class: "btn btn-outline-primary" %>
+  <% end %>
+</div>
--- a/app/components/nested_form_component.rb
+++ b/app/components/nested_form_component.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+# Support for nested forms.
+class NestedFormComponent < ApplicationComponent
+  # Optional slot. If provided, the button must have add as its action for the controller.
+  # See H2FormBuilder#add_another_button
+  renders_one :add_another_button
+
+  def initialize(controller:, form:, field:, clazz:, row:, ordered: false)
+    @controller = controller
+    @form = form
+    @field = field
+    @clazz = clazz
+    @row = row
+    @ordered = ordered
+  end
+
+  attr_reader :controller, :form, :field, :clazz, :row, :ordered
+
+  def nested_forms
+    # Form object must have a method that returns nested forms for this field.
+    form.object.send("#{field}_forms".to_sym)
+  end
+
+  def other_row_params
+    ordered ? {ordered:} : {}
+  end
+end
--- a/app/components/wokes/affiliation_row_component.html.erb
+++ b/app/components/wokes/affiliation_row_component.html.erb
@@ -1,0 +1,21 @@
+<div class="mb-3 row">
+  <div class="col-sm-5">
+      <div data-controller="autocomplete" data-autocomplete-url-value="/ror" data-autocomplete-min-length-value="3" class="dropdown" role="combobox" aria-expanded="false" data-autocomplete-ready-value="true" aria-label="Autocomplete for affiliation">
+        <!-- This input is populated by autocomplete controller.-->
+        <!-- That triggers the autocomplete-edit controller, which populates the uri and cocina type inputs. -->
+        <%= form.label :label, "Affiliation *", class: "col-form-label" %>
+        <%= form.text_field :label, class: "form-control affiliation-input", data: {autocomplete_target: "input"} %>
+        <%= form.hidden_field :uri, {"data-autocomplete-target": "hidden", "aria-label": "URI for affiliation", class: "affiliation-input"} %>
+        <ul class="list-group" data-autocomplete-target="results"></ul>
+      </div>
+  </div>
+  <div class="col-sm-5">
+      <%= form.label :department, "Department/Institute/Center", class: "col-form-label" %>
+      <%= form.text_field :department, class: "form-control affiliation-input" %>
+  </div>
+  <div class="offset-sm-1 col-sm-1">
+    <%= form.remove_button controller, "Remove", class: "btn btn-sm", aria: {label: "Remove affiliation"} do %>
+      <span class="fa-regular fa-trash-alt"></span>
+    <% end %>
+  </div>
+</div>
--- a/app/components/wokes/authors_component.html.erb
+++ b/app/components/wokes/authors_component.html.erb
@@ -1,0 +1,14 @@
+<section>
+  <header aria-describedby="popover-work.author">
+    Authors to include in citation
+    <%= render PopoverComponent.new key: "work.author" %>
+  </header>
+  <p>When there are multiple authors, list them in the order they should appear in the citation. If you need to change the order of the authors, click the arrows to move individual authors up or down in the list.</p>
+
+  <%= render NestedFormComponent.new(form: form, controller: "author-form", row: Wokes::ContributorRowComponent,
+        field: "authors", clazz: Forms::Author, ordered: true) do |component| %>
+    <% component.with_add_another_button do %>
+        <%= form.add_another_button "author-form", "+ Add another author", class: "btn btn-outline-primary" %>
+    <% end %>
+  <% end %>
+</section>
--- a/app/components/wokes/buttons_component.html.erb
+++ b/app/components/wokes/buttons_component.html.erb
@@ -1,0 +1,41 @@
+<!--
+<div data-deposit-button-target="globusMessage" class="row justify-content-end">
+  <div class="col-8">
+    <p class="text-end my-4">
+      <strong>
+        <span class="fa-solid fa-exclamation-circle me-2" style="color: var(--stanford-warning);"></span>
+        Deposit is disabled until file transfer is complete.
+      </strong>
+    </p>
+  </div>
+</div>
+-->
+<div class="row justify-content-between">
+  <div class="col-auto">
+    <% if show_first_draft_cancel? %>
+      <%= link_to model, method: :delete, aria: {label: "Delete #{title}"},
+            data: {
+              confirm: "Are you sure you want to delete this draft work? It cannot be undone.",
+              action: "unsaved-changes#allowFormSubmission"
+            } do %>
+        <span class="btn btn-primary"><span class="fa-regular fa-trash-alt"></span> Discard draft</span>
+      <% end %>
+    <% elsif show_version_draft_cancel? %>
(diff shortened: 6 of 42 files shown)


Gemfile
```diff
@@ -85,6 +85,7 @@ gem "sneakers", "~> 2.11" # rabbitMQ background processing
 gem "state_machines-activerecord"
 gem "strip_attributes"
 gem "turbo-rails", "~> 1.0"
-gem "view_component", "~> 2.56.2" # https://github.com/github/view_component/issues/1390
+gem "view_component", "~> 3.5"
```
justinlittman, line 88: Surprise! We're using an old version of View Components, which causes all sorts of problems for form rendering.

Gemfile
```diff
@@ -85,6 +85,7 @@ gem "sneakers", "~> 2.11" # rabbitMQ background processing
...
 gem "whenever", require: false # Work around https://github.com/javan/whenever/issues/831
+gem "yaaf" # Form objects
```
justinlittman, line 90: A tiny gem that helps with form objects.

app/components/nested_form_component.html.erb
```diff
@@ -0,0 +1,29 @@
+<div data-controller="<%= controller %>" data-<%= controller %>-wrapper-selector-value=".<%= controller %>-wrapper">
```
justinlittman, line 1: Part of the strategy is to abstract nested forms. This provides a nested form component that handles creating the template, rendering the rows, providing the elements needed by the nested forms stimulus controller, and rendering the add another button.

app/components/nested_form_component.html.erb
```diff
@@ -0,0 +1,29 @@
...
+      <%= nested_form.hidden_field :_destroy %>
+      <%# nested_form may have a remove button which has remove action for the controller %>
+      <%= render row.new(form: nested_form, controller: controller, **other_row_params) %>
+    </div>
+  <% end %>
+
+  <%# Inserted elements will be injected before that target. %>
+  <div data-<%= controller %>-target="target"></div>
+
+  <% if add_another_button? %>
```
justinlittman, line 24: Allow a customized add another button to be provided, but provide default.

app/components/nested_form_component.rb
```diff
@@ -0,0 +1,28 @@
...
+  # See H2FormBuilder#add_another_button
+  renders_one :add_another_button
+
+  def initialize(controller:, form:, field:, clazz:, row:, ordered: false)
+    @controller = controller
+    @form = form
+    @field = field
+    @clazz = clazz
+    @row = row
+    @ordered = ordered
```
justinlittman, line 15: This allows specifying that the nested forms are ordered (e.g, authors).

app/components/wokes/affiliation_row_component.html.erb
```diff
@@ -0,0 +1,21 @@
+<div class="mb-3 row">
```
justinlittman, line 1: A nested form must have a row component that renders an individual form. This is for affiliations. It is a bit more complicated because it uses autocomplete.

app/components/wokes/authors_component.html.erb
```diff
@@ -0,0 +1,14 @@
+<section>
+  <header aria-describedby="popover-work.author">
+    Authors to include in citation
+    <%= render PopoverComponent.new key: "work.author" %>
+  </header>
+  <p>When there are multiple authors, list them in the order they should appear in the citation. If you need to change the order of the authors, click the arrows to move individual authors up or down in the list.</p>
+
+  <%= render NestedFormComponent.new(form: form, controller: "author-form", row: Wokes::ContributorRowComponent,
```
justinlittman, line 8: Authors are just like contributors except they use a different controller name, a different field ("authors" instead of "contributors"), a different form class, and are ordered. However, the row component is the same (which is where the bulk of the rendering is).

app/components/wokes/authors_component.html.erb
```diff
@@ -0,0 +1,14 @@
...
+        field: "authors", clazz: Forms::Author, ordered: true) do |component| %>
+    <% component.with_add_another_button do %>
```
justinlittman, line 10: This is an example of providing a custom add another button.

app/components/wokes/buttons_component.html.erb
```diff
@@ -0,0 +1,41 @@
+<!--
```
justinlittman, line 1: Just copied, with some code that relies on files commented out.

app/components/wokes/buttons_component.rb
```diff
@@ -0,0 +1,62 @@
...
+    # def work_version
+    #   object.model.fetch(:work_version)
+    # end
+
+    # def model
+    #   object.model.fetch(:work)
+    # end
+
+    def collection
+      form.object.collection
```
justinlittman, line 31: Getting the object from a form is much more straight-forward.

app/components/wokes/contact_email_row_component.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="mb-3 row plain-container">
```
justinlittman, line 1: Contact email is probably the most basic nested form.

app/components/wokes/contributor_row_component.html.erb
```diff
@@ -0,0 +1,109 @@
...
+        <%= form.text_field :full_name, html_options("contributorOrg").merge("aria-describedby": "popover-work.organization") %>
+        <div class="invalid-feedback">You must provide a name</div>
+      </div>
+    </div>
+  </div>
+  <div class="col-md-1 text-end">
+    <%= form.remove_button controller, html_options_for_delete do %>
+      <span class="fa-regular fa-trash-alt"></span>
+    <% end %>
+    <% if ordered? %>
```
justinlittman, line 88: Adds the buttons/form fields for ordering.

app/components/wokes/form_component.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    <ul>
+      <% form.errors.each do |error| %>
+        <li><%= error.full_message %></li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
+<%= form_with model: form, url: url, scope: :work, builder: H2FormBuilder,
+      data: {
+        # controller: "auto-citation unsaved-changes deposit-button",
```
justinlittman, line 14: Commented out these controllers since they're unhappy with the partial form implementation.

app/controllers/wokes_controller.rb
```diff
@@ -0,0 +1,63 @@
+class WokesController < ObjectsController
```
justinlittman, line 1: The pattern here is similar to any other Active Record controller (even though using form objects).

app/forms/h2_form_builder.rb
```diff
@@ -0,0 +1,42 @@
+class H2FormBuilder < ActionView::Helpers::FormBuilder
```
justinlittman, line 1: As part of abstracting nested forms, helpers for nested form buttons are provided. These provide the necessary data attributes for the nested form stimulus controller, so as to avoid have them copied throughout the ERBs.

app/javascript/controllers/index.js
```diff
@@ -1,8 +1,15 @@
...
 import { Autocomplete } from 'stimulus-autocomplete'
+import NestedForm from 'stimulus-rails-nested-form'
+import OrderedForm from './ordered_form_controller'
 import { definitions } from 'stimulus:./'
 
 const application = Application.start()
 application.load(definitions)
 
 application.register('autocomplete', Autocomplete)
+application.register('author-form', OrderedForm)
```
justinlittman, line 11: Because of double nesting, each nested form controller must be separately registered with a unique name.

app/javascript/controllers/ordered_form_controller.js
```diff
@@ -0,0 +1,72 @@
+import NestedForm from 'stimulus-rails-nested-form'
+
+export default class extends NestedForm {
```
justinlittman, line 3: This is based on our existing stimulus controller for supporting ordering, but extends NestedForm instead of our own custom nested form controller implementation.

app/models/forms/affiliation.rb
```diff
@@ -0,0 +1,41 @@
+module Forms
```
justinlittman, line 1: Yes, it probably makes sense to put this in `app/forms`. However, since we already have our reform forms in this location, I put them in `app/models/forms`.

app/models/forms/affiliation.rb
```diff
@@ -0,0 +1,41 @@
...
+  class Affiliation < Base
+    # Indicates that this is a deposit, and therefore should be fully validated.
+    attr_accessor :_deposit
+    attr_accessor :id
```
justinlittman, line 5: Each attribute get an accessor.

app/models/forms/affiliation.rb
```diff
@@ -0,0 +1,41 @@
...
+    attr_accessor :label
+    attr_accessor :uri
+    attr_accessor :department
+    attr_accessor :abstract_contributor
+    attr_accessor :_destroy
+
+    validates :label, presence: true, if: :_deposit
```
justinlittman, line 12: Notice that validation is triggered by the _deposit field. This allows for conditional validation (no validation when saving draft; only fully validate when depositing).

app/models/forms/affiliation.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def main_model
+      affiliation
+    end
+
+    def affiliation
+      @affiliation ||= begin
+        affiliation = if id.present?
```
justinlittman, line 20: Update an existing affiliation if there is one; otherwise populate a new affiliation.

app/models/forms/affiliation.rb
```diff
@@ -0,0 +1,41 @@
...
+        end
+        affiliation.label = label
+        affiliation.uri = uri
+        affiliation.department = department
+        affiliation.abstract_contributor = abstract_contributor
+        affiliation
+      end
+    end
+
+    def self.new_from_model(affiliation)
```
justinlittman, line 33: This populates the form object from an existing affiliation instance.

app/models/forms/author.rb
```diff
@@ -0,0 +1,7 @@
+module Forms
+  class Author < Contributor
```
justinlittman, line 2: The only difference between the form object for an author and a contributor, is that an author form object creates author active records and a contributor creates contributor active records.

app/models/forms/base.rb
```diff
@@ -0,0 +1,42 @@
...
+  class Base < YAAF::Form
+    after_save :destroy_associated_models
+    attr_reader :main_models, :related_modes, :associated_forms
+
+    delegate :to_param, :persisted?, to: :main_model
+
+    def initialize(attributes = {})
+      super(attributes)
+
+      @models = parent_models + [main_model] + associated_forms
```
justinlittman, line 11: This is the crux of configuration for YAAF: providing the models to be persisted. Only the main model is required.

justinlittman: P.S. The order matters here for active record.

app/models/forms/base.rb
```diff
@@ -0,0 +1,42 @@
...
+    def parent_models
+      []
+    end
+
+    # May be implemented by subclasses
+    def associated_forms
+      []
+    end
+
+    def self.reject_all_blank?(params)
```
justinlittman, line 30: Helper to determine if a submitted form should be ignored because it is blank.

app/models/forms/base.rb
```diff
@@ -0,0 +1,42 @@
+module Forms
+  class Base < YAAF::Form
+    after_save :destroy_associated_models
```
justinlittman, line 3: This provides automagic destruction of the associated models.

app/models/forms/contributor.rb
```diff
@@ -0,0 +1,99 @@
+module Forms
+  class Contributor < Base
+    # Indicates that this is a deposit, and therefore should be fully validated.
+    attr_accessor :_deposit
+    attr_accessor :id
+    attr_accessor :first_name
+    attr_accessor :last_name
+    attr_accessor :full_name
+    attr_accessor :with_orcid # radio button
```
justinlittman, line 9: Here is an example of where the form object reflects the form that is rendered. In reform, this would have been a virtual attribute.

app/models/forms/contributor.rb
```diff
@@ -0,0 +1,99 @@
...
+      elsif contributor.affiliations.present?
+        contributor.affiliations.map do |affiliation|
+          Forms::Affiliation.new_from_model(affiliation)
+        end
+      else
+        []
+      end
+    end
+
+    def affiliations_forms
```
justinlittman, line 79: `<model>_forms` is used to provide a blank form when there are none. This is expected by the nested form component.

app/models/forms/contributor.rb
```diff
@@ -0,0 +1,99 @@
...
+        last_name: contributor.last_name,
+        work_version: contributor.work_version,
+        role_term: contributor.role_term,
+        full_name: contributor.full_name,
+        orcid: contributor.orcid,
+        with_orcid: contributor.orcid.present?,
+        weight: contributor.weight)
+    end
+
+    def self.reject_all_blank?(params)
```
justinlittman, line 95: For some form objects, some form attributes should be ignored when determining if a form is blank. (These attributes always have values.)

app/models/forms/work.rb
```diff
@@ -0,0 +1,150 @@
+module Forms
+  class Work < Base
```
justinlittman, line 2: Work is kinda complicated because it needs to handle both the work and work version models. It also has multiple nested forms (authors, contributors, email). This is an example of where a parent model is used.

config/application.rb
```diff
@@ -49,5 +49,7 @@ class Application < Rails::Application
         config.report_data = false
       end
     end
+
+    config.view_component.capture_compatibility_patch_enabled = true
```
justinlittman, line 53: View Components struggles with forms. This patches some of those problems.

package.json
```diff
@@ -16,8 +17,8 @@
     "lodash.debounce": "^4.0.8",
     "sass": "^1.43.2",
     "simple-datatables": "^3.0.0",
-    "@hotwired/stimulus": "^3.2.1",
-    "stimulus-autocomplete": "^3.1.0"
+    "stimulus-autocomplete": "^3.1.0",
+    "stimulus-rails-nested-form": "^4.1.0"
```
justinlittman, line 21: Using stimulus-rails-nested-form instead of rolling our own.

justinlittman:
For the purposes of this proof-of-concept, I implemented a wokes controller (that sits alongside the existing works controller).

To create a new work (with the wokes controller): `/collections/1/wokes/new`
To edit an existing work (with the wokes controller): `/wokes/17/edit`

justinlittman:
I only implemented part of the work form:
image

However, it includes authors/contributors which requires double nesting (for affiliations), polymorphism (for authors vs contributors), and ordering (for authors), so, in theory, everything else is easy.

PR #3308, opened 2023-07-26
Change globus endpoint permissions on deposit

### LINKED ISSUE #3302: Change globus permissions to read-only when depositing
When a user starts the deposit of a work version with a globus_endpoint, change the user's permissions on the globus_endpoint directory from `"rw"` to `"r"`. This will prevent the user from changing the files there while accessioning is happening. 

While this work was [originally ticketed in sdr-api](https://github.com/sul-dlss/sdr-api/issues/526), it seems like it would be simpler to implement in H2, which is already kicking off globus operations related to managing endpoints and is using theGlobusClient.  

globus_client has a `disallow_writes` method already implemented to support this functionality.

lwrubel:
# Why was this change made? 🤔

Resolves #3302 to change globus permissions to "r" when depositing starts. 

Unit and stage deploy. Also ran globus integration test. 

If a user tries to rename or edit a file on the globus endpoint once the deposit job starts, they'll get a permission denied message. 

Screenshot 2023-07-26 at 4 08 46 PM

Screenshot 2023-07-26 at 4 07 50 PM

Does not touch front-end.

### DIFF
--- a/app/jobs/deposit_job.rb
+++ b/app/jobs/deposit_job.rb
@@ -13,6 +13,8 @@ class DepositJob < BaseDepositJob
     # NOTE: this login ensures `Repository.find` and various SdrClient::* calls below all have a valid token
     perform_login
 
+    disallow_writes(work_version) if work_version.globus_endpoint.present?
+
     cocina_obj = Repository.find(druid) if druid
     request_dro = CocinaGenerator::DROGenerator.generate_model(work_version:, cocina_obj:)
     new_request_dro = update_dro_with_file_identifiers(request_dro, work_version)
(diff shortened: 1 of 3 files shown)


peetucket (approved):
Couple questions, else looks good.

app/jobs/deposit_job.rb
```diff
@@ -32,6 +34,14 @@ def perform_login
     raise login_result.failure unless login_result.success?
   end
 
+  def disallow_writes(work_version)
+    return true if test_mode?
+    return true if integration_test_mode? && work_version.globus_endpoint == integration_endpoint
+
+    # user_id nil because unneeded for permission update operations
+    GlobusClient.disallow_writes(path: work_version.globus_endpoint, user_id: nil)
+  end
```
peetucket, line 43: presume this method is already in the client gem used here, right?

lwrubel: Yes, method was already in the gem, just unused.

lwrubel:
> Just to verify, once deposit begins, this folder is now read-only permanently, which is fine because if they make a new version, they get a new folder?

It's read-only while depositing / accessioning is happening. We don't want people moving files around. The next step is to implement https://github.com/sul-dlss/globus_client/issues/95 to remove access altogether once the deposit is complete. Once a deposit happens, we want them to create a new version and they'll get a new endpoint for that. At least, that's what we're doing now.

PR #3332, opened 2023-08-02
Minor update to accessibility best practice update for modal headers

aaron-collier:
# Why was this change made? 🤔

SiteImprove suggest the best practice for pages (and modals, etc) is that the top/header visual element be an H1 to apply proper hierarchy, starting with h2/h3 violates this best practice. From SiteImprove:

```
Heading tags help assistive technology users to understand how content is structured. They're also used for in-page screen reader navigation.

When determining heading levels, you should ensure that they follow a nested structure. h1 is used for the primary heading, followed by h2 for subheadings, and so on to h6.
```

Fixes 2 violations.

### DIFF
--- a/app/views/shared/_help_modal.html.erb
+++ b/app/views/shared/_help_modal.html.erb
@@ -2,7 +2,7 @@
   <div class="modal-dialog modal-dialog-centered ">
     <div class="modal-content">
       <div class="modal-header">
-        <h3 class="modal-title" id="help-prompt">Help</h3>
+        <h1 class="modal-title h3" id="help-prompt">Help</h1>
         <button type="button" class="btn-close" data-bs-dismiss="modal" aria-label="Close"></button>
       </div>
       <div class="modal-body">
--- a/app/views/shared/_terms_of_deposit.html.erb
+++ b/app/views/shared/_terms_of_deposit.html.erb
@@ -3,7 +3,7 @@
     <div class="modal-content">
       <div class="row justify-content-between modal-header">
           <div class="col-5">
-            <h3 class="modal-title" id="terms-of-deposit-prompt">Terms of Deposit</h3>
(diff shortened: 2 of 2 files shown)


app/views/shared/_help_modal.html.erb
```diff
@@ -2,7 +2,7 @@
   <div class="modal-dialog modal-dialog-centered ">
     <div class="modal-content">
       <div class="modal-header">
-        <h3 class="modal-title" id="help-prompt">Help</h3>
+        <h1 class="modal-title" id="help-prompt">Help</h1>
```
mjgiarlo, line 5: I wonder if it's OK for the size of this text to bump up significantly, or if we want to use the `<h1>` element for semantics and use CSS to size the text closer to that of an `<h3>`. I believe adding the `h3` CSS class might do this (via bootstrap)?

app/views/shared/_terms_of_deposit.html.erb
```diff
@@ -3,7 +3,7 @@
     <div class="modal-content">
       <div class="row justify-content-between modal-header">
           <div class="col-5">
-            <h3 class="modal-title" id="terms-of-deposit-prompt">Terms of Deposit</h3>
+            <h1 class="modal-title" id="terms-of-deposit-prompt">Terms of Deposit</h1>
```
mjgiarlo, line 6: I wonder if it's OK for the size of this text to bump up significantly, or if we want to use the `<h1>` element for semantics and use CSS to size the text closer to that of an `<h3>`. I believe adding the `h3` CSS class might do this (via bootstrap)?

app/views/shared/_help_modal.html.erb
aaron-collier: Good suggestion @mjgiarlo fixed.

Issue #3170, opened 2023-06-29
Deposit form:  "major" accessibility issues that may not need design


ndushay:
From https://docs.google.com/spreadsheets/d/1Fs5sY1zgpovWVdHMqpJWRu2Zvyr53jw7, here are "major" accessibility issues for the **Deposit Form** that we can fix without design changes.  Note that I addressed accessibility level A and AA problems noted by the SiteImprove, WAVE and Axe browser plugins in PR #3169.

Note that solutions may be suggested in https://docs.google.com/spreadsheets/d/1Fs5sY1zgpovWVdHMqpJWRu2Zvyr53jw7.

These can be separate PRs, and/or split into separate tickets.

- [ ] ~~There are multiple "Remove" buttons that perform different actions.  Form field labels should be unique.~~ See #3181
- [x] When the "Upload one or more files (more than 10GB) which will be displayed as uploaded, including any file hierarchy" option  is selected, the six steps listed are not structured as a list. See #3333
- [ ] When the "See more options" control is activated, focus moves past the new content to the "Continue" button (in both the modal dialog and in the main page).
- [x] When the "See more options"/"See fewer options" disclosure widget is used it is missing an aria-expanded (in the modal dialog and on the main page).
- [x] There are headings that are only styled to look like headings but are not marked as such in HTML.  Each section has a visible heading, but they are not marked up as headings.  Examples include:
    - Add your files*
    - Title of deposit and contact information*
    - List authors and contributors *
and many more.
- [x] There are groups of form controls that do not provide <fieldset> and <legend> elements.  The radio buttons in "Add your files*" and the date fields in "Enter dates related to your deposit" are not grouped appropriately. 
- [ ] Form field constraints and/or errors are either not clearly displayed or are not programmatically associated with the field they apply to.  Errors are not explicitly associated with form fields, which means that the error text is not conveyed when the fields have focus and are in error.   
        There are `<select>` elements with asterisk that are not programmatically conveyed as required: 
          - In the List authors and contributors section, the Role term `<select>`  
          - In the Select a license section, the License `<select>`
- [x] The toggle for "Use auto-generated citation" does not receive keyboard focus: PR #3335
- [x] The "Remove" buttons (trash can images) receive keyboard focus with no visual indication when using TAB or SHIFT+TAB to navigate through them. PR #3335
[...]

PR #3333, opened 2023-08-02
Switch upload instructions to a list

aaron-collier:
# Why was this change made? 🤔

Refs #3170 

This updates some of the divs to a `ul > li` list so that it is presented as a list to screen readers but does not change the visual representation.
 
Screenshot 2023-08-02 at 2 26 17 PM

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -78,36 +78,36 @@
     <div data-file-uploads-target="globusUpload">
       <p>Please notify us at <a href="mailto:sdr-contact@lists.stanford.edu">sdr-contact@lists.stanford.edu</a> if you are uploading 1TB or more of content.</p>
       <div class="container py-4">
-        <div class="row g-3">
-          <div class="col">
+        <ol class="row g-3 list-unstyled">
+          <li class="col">
             <div class="border border-dark h-100 p-2 bg-light">
               <p>STEP 1</p>
               <p>If your files are located on Stanford Google Drive, Oak, or Sherlock, select where they are located below.</p>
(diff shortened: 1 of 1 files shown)


app/components/works/add_files_component.html.erb
```diff
@@ -78,36 +78,36 @@
     <div data-file-uploads-target="globusUpload">
       <p>Please notify us at <a href="mailto:sdr-contact@lists.stanford.edu">sdr-contact@lists.stanford.edu</a> if you are uploading 1TB or more of content.</p>
       <div class="container py-4">
-        <div class="row g-3">
-          <div class="col">
+        <ul class="row g-3 list-unstyled"">
```
mjgiarlo, line 81: Looks like there may be an extra `"` character here

mjgiarlo: I'm wondering about using `ul` here, as that implies a lack of order in the list. Is `ol` more semantically appropriate? Or is there a different HTML tag that might be more appropriate for this content? (If so, I'm not sure what that might be.)

aaron-collier: Oops, good catch.

app/components/works/add_files_component.html.erb
```diff
@@ -78,36 +78,36 @@
     <div data-file-uploads-target="globusUpload">
       <p>Please notify us at <a href="mailto:sdr-contact@lists.stanford.edu">sdr-contact@lists.stanford.edu</a> if you are uploading 1TB or more of content.</p>
       <div class="container py-4">
-        <div class="row g-3">
-          <div class="col">
+        <ol class="row g-3 list-unstyled">
```
mjgiarlo, line 81: Have you considered a `ul` instead? 😆

mjgiarlo (approved):

PR #3334, opened 2023-08-02
Make form headings fit h1/h2/h3 hierarchy for accessibility

aaron-collier:
# Why was this change made? 🤔

Refs #3170 

This moves the `<header>` components on the deposit form to actual `h2` and `h3` values to support the proper hierarchy for accessibility. 

Note from SODA recommendation:
```
Including the text in a <header> element does not make the text a heading.  
```

### DIFF
--- a/spec/support/capybara_actions.rb
+++ b/spec/support/capybara_actions.rb
@@ -7,7 +7,9 @@ module CapybaraActions
   end
 
   def within_section(title, &)
-    within(:xpath, "//section[contains(header/text(),'#{title}')]", &)
+    within(:xpath, "//section[contains(h2/text(),'#{title}')]", &)
+  rescue Capybara::ElementNotFound
+    within(:xpath, "//section[contains(h3/text(),'#{title}')]", &)
   end
 
   # NOTE: this is here to ensure all turbo frames are loaded before auditing a11y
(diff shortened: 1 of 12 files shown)


spec/support/capybara_actions.rb
```diff
@@ -7,7 +7,9 @@ def blur_from(locator)
   end
 
   def within_section(title, &)
-    within(:xpath, "//section[contains(header/text(),'#{title}')]", &)
+    within(:xpath, "//section[contains(h2/text(),'#{title}')]", &)
+  rescue Capybara::ElementNotFound
```
aaron-collier, line 11: This may not be the best approach, but I could not get a value xpath OR query to work, and `within` raises an error when not found, so we can't do something like `within(..) || within(..)`... it also complained if I passed a string in (h2 or h3) and used string interpolation.. 

I'm open to suggestion if this seems problematic.

app/assets/stylesheets/editor.scss
```diff
@@ -36,6 +36,10 @@
     margin-bottom: 1rem;
   }
 
+  .deposit_header {
+    font-weight: bold;
+  }
```
mjgiarlo, line 41: Since this new class only contains a single rule, I wonder if we need it or if we could use the Bootstrap built-in class (`fw-bold`) below instead of the new custom class?

mjgiarlo (commented):
Big fan of this change. Wondering about using custom CSS classes vs. Bootstrap built-ins.

mjgiarlo (approved):

PR #3335, opened 2023-08-03
Visually indicate when the form buttons get focus

aaron-collier:
# Why was this change made? 🤔

Refs #3170 - this adds a visual indication when a button has gained focus using tab navigation.

WGAG Recommendation:
```
Ensure that when actionable elements receive focus, that it appears on screen and that a visible focus
indicator is present. Focus can be indicated in different ways, including change of background, lightness,
border, outline, underline, and other visual methods. Avoid the use of the "outline: none" CSS property.
As of WCAG 2.1, a custom focus indicator should have a minimum contrast ratio of 3:1 against the background.
```

### DIFF
--- a/app/components/works/description_component.html.erb
+++ b/app/components/works/description_component.html.erb
@@ -49,7 +49,7 @@
                   }
                 },
                 "true", "false" %>
-          <div class="slider round">
+          <div class="slider round" tabindex="0">
             <span class="on">Yes</span>
             <span class="off">No</span>
           </div>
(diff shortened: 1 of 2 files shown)


app/components/works/description_component.html.erb
```diff
@@ -49,7 +49,7 @@
                   }
                 },
                 "true", "false" %>
-          <div class="slider round">
+          <div class="slider round" tabindex="0">
```
aaron-collier, line 52: Note: this allows the auto-citation slider to gain tab focus.

PR #3337, opened 2023-08-03
Add a basic example to complete the help modal accessibility

### LINKED ISSUE #3193: Help modal: address "critical" and "major" accessibility issues
# Critical

- [x] Ensure form fields provide a programmatically associated label that describes the purpose of each field. 

From SODA: 
```Each of the form fields is duplicated in the Help modal dialog and the relationships between the for and id values are negated because they are duplicated.

Users may not be able to fill out form fields correctly, if there is no programmatically associated and visible label for each field.
```

# Major

- [x] Ensure each element has a unique ID

From SODA:

```Some elements are not well formed.  There are multiple id attributes with the same value that are referenced by other elements.  This occurs for each of the form fields that are duplicated in the Help dialog.

Though more of a usability issue, it is not clear why a user who might need help with form fields would gain assistance from a dialog with the same fields repeated.

When elements are not well formed, assistive technologies may not behave as intended.
```

- [x] Ensure each field has a meaningful example that communicates purpose.

From SODA:
```

There are common input fields that lack autocomplete attributes to communicate their purpose programmatically: the "what is your name" and "what is your email address" fields in the main screen and in the Help dialog.

When autocomplete is used to communicate the purpose of a field, users of assistive technology or users with cognitive disabilities can have access to icons or familiar terms indicating the field's purpose. In addition, when a form is filled out automatically using information the user has previously entered, the user does not need to decipher input fields that may appear differently in different forms and may not need to remember and re-enter data.
```

aaron-collier:
# Why was this change made? 🤔

Fixes #3193 by adding meaningful examples to the fields
 
Screenshot 2023-08-03 at 2 47 44 PM

CI

SiteImprove

### DIFF
--- a/app/views/helps/new.html.erb
+++ b/app/views/helps/new.html.erb
@@ -6,7 +6,7 @@
          <div class="col-md-9">
            <div class="form-group">
              <%= form.label(:name, "What is your name?") %>
-             <%= form.text_field(:name, autocomplete: "name", class: "form-control") %>
+             <%= form.text_field(:name, autocomplete: "name", class: "form-control", value: @name) %>
            </div>
          </div>
       </div>
@@ -22,7 +22,7 @@
          <div class="col-md-9">
            <div class="form-group">
              <%= form.label(:affiliation, "What is your Stanford affiliation/department?") %>
-             <%= form.text_field(:affiliation, class: "form-control") %>
+             <%= form.text_field(:affiliation, class: "form-control", placeholder: "(ex: Student)") %>
            </div>
          </div>
       </div>
--- a/app/controllers/helps_controller.rb
+++ b/app/controllers/helps_controller.rb
@@ -4,6 +4,7 @@
 class HelpsController < ApplicationController
   def new
     @email = current_user&.email
+    @name = current_user&.name
     @help_how_value = if params[:show_collections] == "true"
       "Request access to another collection"
     elsif !current_user


app/views/helps/new.html.erb
```diff
@@ -22,7 +22,7 @@
          <div class="col-md-9">
            <div class="form-group">
              <%= form.label(:affiliation, "What is your Stanford affiliation/department?") %>
-             <%= form.text_field(:affiliation, class: "form-control") %>
+             <%= form.text_field(:affiliation, class: "form-control", value: "(ex: Student)") %>
```
mjgiarlo, line 25: If it's of equal accessibility value, I'm on the fence about using an input field value here vs. a placeholder (which renders until the user starts typing). Thoughts?

mjgiarlo (approved):
Approve pending consideration of question

app/views/helps/new.html.erb
aaron-collier: Placeholder is probably better. I missed that option, fixing.

mjgiarlo: OK, cool!

aaron-collier: @mjgiarlo In this case, `placeholder` is perfect - and it works better for the screen reader as well! So double win, thanks!

PR #3341, opened 2023-08-07
Improve aria-labels for uniqueness and accessibility

aaron-collier:
# Why was this change made? 🤔

Improves the uniqueness of a few labels so that screen readers use the correct label.

Si

SI

### DIFF
--- a/app/components/works/contact_email_row_component.html.erb
+++ b/app/components/works/contact_email_row_component.html.erb
@@ -8,7 +8,7 @@
     <div class="invalid-feedback">You must provide a valid email address</div>
   </div>
   <div class="col-sm-1">
-    <%= button_tag type: "button", class: "btn btn-sm float-end", aria: {label: "Remove"},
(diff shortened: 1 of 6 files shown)


app/components/works/contact_email_row_component.html.erb
```diff
@@ -8,7 +8,7 @@
     <div class="invalid-feedback">You must provide a valid email address</div>
   </div>
   <div class="col-sm-1">
-    <%= button_tag type: "button", class: "btn btn-sm float-end", aria: {label: "Remove"},
+    <%= button_tag type: "button", class: "btn btn-sm float-end", aria: {label: "Remove Association"},
```
mjgiarlo, line 11: "Remove Association" might strike the user as vague.  What's being removed here?

aaron-collier: Funny that "Remove" was dinged as not meaningful...

mjgiarlo: Would something like `Remove contact email` be appropriate?

mjgiarlo (approved):

PR #3344, opened 2023-08-07
Use list types when presenting content as a list

aaron-collier:
# Why was this change made? 🤔

This reformats the terms of use modal link at the top to use a list for assisting screen readers. The visual representation did not change:

before:
Screenshot 2023-08-07 at 3 34 56 PM

after:
Screenshot 2023-08-07 at 3 43 20 PM

### DIFF
--- a/app/views/shared/_terms_of_deposit.html.erb
+++ b/app/views/shared/_terms_of_deposit.html.erb
@@ -5,20 +5,24 @@
           <div class="col-5">
             <h1 class="modal-title h3" id="terms-of-deposit-prompt">Terms of Deposit</h1>
           </div>
-          <div class="col-2">
-            <%= link_to Settings.faq_url, aria: {label: "FAQs about Terms of Deposit"}, id: "faq-link", class: "text-decoration-none", target: "_blank" do %>
-              <span class="fa-solid fa-question-circle"></span> FAQs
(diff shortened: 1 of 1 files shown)


app/views/shared/_terms_of_deposit.html.erb
```diff
@@ -5,20 +5,24 @@
...
+                  <span class="fa-solid fa-save"></span> Save
+                <% end %>
+              </li>
+              <li class="col-4">
+                <%= link_to print_terms_of_deposit_path, aria: {label: "Print Terms of Deposit"}, id: "save-link", class: "text-decoration-none", target: "_blank" do %>
+                  <span class="fa-solid fa-print"></span> Print
+                <% end %>
+              </li>
+            </ul>
           </div>
```
mjgiarlo, line 26: Couple quick questions:

1. I naïvely assumed each `li` would wind up with a `col-2` class given the containing `div` is a `col-6`
2. Did you find that the new containing `div` is needed to retain the current look & feel?

mjgiarlo: @aaron-collier Ah, OK, makes sense. Thanks. 🙂

mjgiarlo (approved):

PR #3361, opened 2023-08-10
Do not provide file upload option when too many files.

### LINKED ISSUE #3352: Restrict file add/update/remove options when user has uploaded >250 files
When a depositor adds more than 250 files via the single zip file upload or Globus and then attempts to edit the item, two things happen that are not good user experiences:

- the files may load in the file upload section, but the user is unable to add more files because we have limited this to 250 files
- the app is unable to display all the files and eventually crashes.

To prevent these problems, the proposal is that we will

- not display more than 250 files in the file upload section
- if there are more than 250 files, only permit the user to modify their files using the single zip file or the Globus option. The first option where individual files can be uploaded in a flat list will not be displayed to these users. The user will instead see a message that indicates they have more than 250 files and they need to use one of the other options if they want to modify their files.
- This will apply for any deposited item with more than 250 files or any item in a draft state with more than 250 files.

This does mean that users with more than 250 files will not be able to hide files or add descriptions, but it is unlikely users with this much content are using these features, and we can handle any such requests manually via Argo if necessary.

justinlittman:
closes #3352

# Why was this change made? 🤔
The browser upload section does not function well with too many files.

Local, unit

image

Tested with SiteImprove plugin.

### DIFF
--- a/app/components/works/add_files_component.html.erb
+++ b/app/components/works/add_files_component.html.erb
@@ -10,36 +10,44 @@
   <fieldset>
     <legend class="visually-hidden">Add or modify files</legend>
     <div class="form-check" data-complex-radio-target="selection">
-      <%= form.radio_button :upload_type, "browser", class: "form-check-input", "data-file-uploads-target": "browserRadioButton",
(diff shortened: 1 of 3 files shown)


mjgiarlo (approved):
LGTM. Do you think this would benefit from PO review or should we ship it?

PR #3363, opened 2023-08-10
Refactor custom term form.

### LINKED ISSUE #3357: Revise custom terms options for collection managers and add validation
After some discussion with @justinlittman about adding validations to the form as is, we came up with a plan that would simplify the form, make the structure of these options more similar to other options on the form, and make it easier to apply the necessary validations.

The top part of the custom terms section of the form should stay as is:

Below this header there should be three options with radio buttons. The first option ("No...") should be selected by default.
- No, do not include a custom use statement
- Yes, include the following custom use statement for every deposit in this collection  [**with entry field for custom terms**]
- Yes, allow the depositor to enter their own custom use statement. The depositor will see either the default instructions or your custom instructions if provided below.

Default instructions: Enter additional terms of use not covered by your chosen license or the default terms shown above, which also display on the PURL page. 
Custom instructions: [** with entry field for custom instructions**]

The instruction lines could be a bulleted list indented below the third option.

- When the second option is chosen, require that there be text in the entry field.
- For the third option, when there is no text entered for instructions, use the default instructions.

justinlittman:
closes #3357

# Why was this change made? 🤔
Simpler form, easier to validate.

Unit, Amy

Tested with SiteImprove.


app/javascript/controllers/complex_radio_controller.js
```diff
@@ -7,11 +7,12 @@ export default class extends Controller {
     this.disableUnselectedInputs();
   }
 
-  disableUnselectedInputs(_event) {
+  disableUnselectedInputs() {
+    console.log('disableUnselectedInputs')
```
mjgiarlo, line 11: Cruft?

mjgiarlo (approved):

app/javascript/controllers/complex_radio_controller.js
justinlittman: Yup. Removed.

Issue #3377, opened 2023-09-11
Unable to add new collection creators


amyehodge:
Adding new users to the dlss:hydrus-app-collection-creators workgroup should give them permission to create a new collection. A button like the one shown below should show up on their dashboard:

<img width="958" alt="Screenshot 2023-09-11 at 2 55 44 PM" src="https://github.com/sul-dlss/happy-heron/assets/5558511/9daa42a6-367f-406a-b6e5-2411c047a5fe">

However, I added a new user, zimmerd, to this workgroup on 09-01-23 and he is still unable to see this button. Just today I tried adding Cory Lown to this workgroup and he was unable to see the button as well. He tried two browsers (Chrome and Firefox, I believe).

I gave zimmerd depositor permissions on a test collection of mine and he was able to log into the app, but still can't see the button.

amyehodge:
Existing collection creator permissions do not seem to be affected. It looks like it's just that new additions to the workgroup are not seeing the button in H2.

peetucket:
I have looked at the code and don't see any recent changes to functionality, so am not sure why it is occurring yet.  Things outside our code that could have changed that would cause issues:

- Workgroup name (`dlss:hydrus-app-collection-creators`)
- Header (`HTTP_X_GROUPS`) coming from shibboleth that determines which groups a user is in
 
But if either of these changed, all current users should be affected too.   

I made a small PR to add some additional logging to see if this gives us any clues for debugging.  

One possible idea is that I noted we currently cache groups in a session variable, so if you add a user to a new group, it won't kick in until their session expires and they log in again.  I assumed the users who are now having problems tried again after clearing their session (I would think closing the browser full and restarting would do it, I used web developer extension to do it on demand).   But again, this caching behavior hasn't changed recently either.

amyehodge:
I just checked and the workgroup name definitely has not changed.
Also, Cory was one of my testers and he now sees the button. He had to authenticate, so it looks like his session was cleared between the last time he checked and now when he can see it.
I'm checking with my other user. Perhaps we just ran into a batch of caching problems that I hadn't seen in a long time and this is okay? I'll check back in when i hear from my other user.

amyehodge:
My other has reported that they can now see the button as well, so I think we can chalk this up to the caching problem creeping back in for the first time in a while.

@peetucket I think we can close this, unless there is something else you think we should look into?

peetucket:
[...]

PR #3381, opened 2023-09-15
add some info the logs on groups passed in via headers for debugging

peetucket:
# Why was this change made? 🤔

To investigate #3377

While we currently log roles coming from the header, it's at the "debug" level, which means it will never log in production/stage/QA.  This may help figure out what's going on.

QA

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -44,7 +44,7 @@ class ApplicationController < ActionController::Base
       raw_header = roles if Rails.env.development?
       # This test setting is for cypress.
       raw_header = roles if Rails.env.test? && roles
-      logger.debug("Roles are #{raw_header}")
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -44,7 +44,7 @@ def groups_from_request_env
       raw_header = roles if Rails.env.development?
       # This test setting is for cypress.
       raw_header = roles if Rails.env.test? && roles
-      logger.debug("Roles are #{raw_header}")
+      logger.info("Rails Env: #{Rails.env}; User: #{current_user}; Roles: #{raw_header}")
       raw_header&.split(";") || []
```
peetucket, line 48: Note: since we do something different in non-production Rails env, let's just log that too for sanity.  Also, log the username so we can be sure we are dealing with the user we expect.  

Since this is only run once per session for a given user, it will only happen when they first log in and get a new session (i.e. it will not be noisy in the logs).

jmartin-sul (approved):

PR #3394, opened 2023-10-06
Preserve files when creating / updating a work version with > 250 files.

### LINKED ISSUE #3392: Depositors can't make metadata-only updates to items where the UI doesn't let them edit files
When a deposit has more than 250 files, users are shown a message that they cannot update their files in the UI, they can only use the zip or Globus option. 

Screenshot 2023-10-05 at 11 19 50 AM

Unfortunately, it appears that the UI is requiring users to modify their files in this situation. A metadata-only update, such as a change in the contact email, is rejected with a message saying that at least one file must be uploaded.

ScreenshotErrorSDR 62

**Expected behavior**

Users should not be required to modify files in order to make metadata updates.

justinlittman:
closes #3392

# Why was this change made? 🤔
Per ticket.

Local, stage

### DIFF
--- a/app/forms/base_work_form.rb
+++ b/app/forms/base_work_form.rb
@@ -197,8 +197,25 @@ class BaseWorkForm < Reform::Form
   def save_model
     super
     # For a zipfile, both the existing files and the zipfile will be attached. These will be removed during unzip.
+
+    # If the previous version has > 250 files and the upload_type is browser then copy the files from previous version.
+    # Having > 250 files and upload_type is browser indicates that the files were originally uploaded via zip or globus.
+    # However, because of the number of files the user is not able to edit them via the UI and they are not provided
+    # in the form, so they must be copied from the previous version.
+    # Note that the octets of the files are not being copied, only the ActiveStorage references.
+    if work.head && work_version.upload_type == "browser" && work.head.attached_files.length > Settings.max_upload_files && work_version.attached_files.empty?
+      dupe_attached_files
+    end
     dedupe_keywords
     work.update(head: work_version)
+  end
+
+  def dupe_attached_files
+    work.head.attached_files.each do |existing_attached_file|
+      new_attached_file = existing_attached_file.dup
+      new_attached_file.file.attach(existing_attached_file.file.blob.signed_id)
+      work_version.attached_files << new_attached_file
+    end
   end
 
   # Override reform so that this looks just like a Work
--- a/spec/features/update_existing_work_spec.rb
+++ b/spec/features/update_existing_work_spec.rb
@@ -49,4 +49,38 @@ RSpec.describe "Update an existing work in a deposited collection", js: true do
       expect(page).to have_content "title of deposit modified, abstract modified", count: 1
     end
   end
+
+  context "when existing work has more than max_upload_files" do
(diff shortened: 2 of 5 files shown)


spec/features/update_existing_work_spec.rb
```diff
@@ -49,4 +49,32 @@
...
+
+      click_link "Edit #{original_title}"
+
+      expect(page).to have_content("You have more than 1 files in your deposit.")
+
+      # update the title and abstract
+      fill_in "Title of deposit", with: new_title
+      fill_in "Abstract", with: "I did really cool stuff"
+
+      click_button "Save as draft"
```
jmartin-sul, line 71: since andrew's ticket mentioned that the draft saves, but without any files, i'm wondering if this could be changed to `click_button "Deposit"` and/or file count could be checked in the page expectations after form submission?  feels to me like one or both of those changes would more precisely check for the buggy behavior being fixed in this PR.

app/forms/base_work_form.rb
```diff
@@ -197,10 +197,23 @@ def save!(*)
   def save_model
     super
     # For a zipfile, both the existing files and the zipfile will be attached. These will be removed during unzip.
+
+    # If the previous version has > 250 files and the upload_type is browser then copy the files from previous version.
```
jmartin-sul, line 201: i've never had the deepest understanding of how file attachments are modeled/handled in H2, and since i don't really understand the root cause of the issue here or why the fix works, could i ask for a little explanation in this comment about why this is only needed when there are more than `max_upload_files` attachments?

also, assuming my understanding is correct that only the database rows are being copied, could the comment be tweaked to indicate that?  (e.g. "copy the attached_files db rows from previous version")

looking at `BaseWorkForm`, i want to ask something like "would it be more idiomatic/maintainable to do this duplication in `AttachedFilesPopulator`?", but given my (lack of) facility with ReForm, i'm not even sure that is a sensible question.

jmartin-sul (commented):
i totally believe it works, but i don't really understand why, so i left a couple questions

spec/features/update_existing_work_spec.rb
justinlittman: In this test, the number of max upload files was set to 1 (instead of 250) and then a work version with 2 files was edited so that this test condition was triggered (editing a deposit with more than max files.)

In the expects below, the presence of both files is verified. I will add a comment to make that clearer.

app/forms/base_work_form.rb
justinlittman: See additional comments which should address most of your questions.

Perhaps this could be done in AttachedFilesPopulator. But I look at all of this as an eyesore until we get rid of reform, so unclear which is best.

jmartin-sul (approved):
thanks for the additional comments, definitely helpful for reviewing, hopefully also helpful for our future selves!

andrewjbtw:
It doesn't look like the QA object has been tested since it's still on version 1. I can test with this deployed to QA.

andrewjbtw:
I tested on QA and this partially works.

- if you edit and press deposit without saving a draft, it does not work
    - this results in the same prompt to attach a file 
- if you edit and save a draft, then go back and reopen that draft, then deposit, it works

PR #3402, opened 2023-10-17
Filter additional OS X files from zip.

### LINKED ISSUE #3401: Ignore additional MacOS-specific file paths in zip and Globus file uploads
See https://github.com/sul-dlss/happy-heron/issues/2935, which points to the code for eliminating some files paths for Globus uploads committed in https://github.com/sul-dlss/happy-heron/pull/2939 and similar code for doing the same for zip uploads which lives here: https://github.com/sul-dlss/happy-heron/blob/main/app/jobs/unzip_job.rb#L32-L34.

Recently a user noticed some [MacOS-specific files that start with "._"](https://apple.stackexchange.com/questions/14980/why-are-dot-underscore-files-created-and-how-can-i-avoid-them) creeping into a deposit. This deposit was originally done with the zip file upload, but then also later via Globus.

These files should be filtered out of uploads in a similar fashion to what was done above.

https://purl.stanford.edu/cp231wr9194
A few of these files are shown in the screenshot below.

Screenshot 2023-10-17 at 4 09 47 PM

justinlittman:
closes #3401

# Why was this change made? 🤔
Per @amyehodge 

Unit


mjgiarlo:
@justinlittman Where does this work stand? Anything I can do to help as current FR?

justinlittman:
@mjgiarlo I just undrafted, so ready for review.

mjgiarlo (approved):

PR #3406, opened 2023-10-19
Skip deposit completion for objects lacking druids

mjgiarlo:
On hold 'til J-Litt has time to do some analysis to determine if this is a legit situation or a weird data corner case. I'm throwing up a quick HOLD PR at the end of my day so J-Litt can take a gander in the morning. Yay timezones.

# Why was this change made?

They are waiting on a druid assignment and we cannot ask the Workflow service if they are accessioned or not until they have a druid. This bug was revealed in a cronjob today.

# How was this change tested?

CI

### DIFF
--- a/spec/services/deposit_complete_auditor_spec.rb
+++ b/spec/services/deposit_complete_auditor_spec.rb
@@ -3,34 +3,91 @@
 require "rails_helper"
 
 RSpec.describe DepositCompleteAuditor do
-  let(:work_still_accessioning) { create(:work_version_with_work_and_collection, state: :depositing, druid: "druid:bc123df4567").work }
-  let(:work_done_accessioning) { create(:work_version_with_work_and_collection, state: :depositing, druid: "druid:bc123df4568").work }
-  let(:collection_still_accessioning) { create(:collection_version_with_collection, state: :depositing, collection_druid: "druid:bc123df4569").collection }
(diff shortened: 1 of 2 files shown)


spec/services/deposit_complete_auditor_spec.rb
```diff
@@ -3,34 +3,91 @@
...
     allow(Repository).to receive(:valid_version?).and_return(true)
-    # These works / collections should be ignored.
-    build(:valid_deposited_work_version)
-    build(:work_version)
-    build(:collection_version_with_collection) # Deposited
-    build(:collection_version)
     allow(Dor::Workflow::Client).to receive(:new).and_return(workflow_client)
-    allow(workflow_client).to receive(:active_lifecycle).with(druid: work_still_accessioning.druid, milestone_name: "submitted", version: work_still_accessioning.head.version.to_s).and_return(Time.now)
-    allow(workflow_client).to receive(:active_lifecycle).with(druid: work_done_accessioning.druid, milestone_name: "submitted", version: work_done_accessioning.head.version.to_s).and_return(nil)
-    allow(workflow_client).to receive(:active_lifecycle).with(druid: collection_still_accessioning.druid, milestone_name: "submitted", version: collection_still_accessioning.head.version.to_s).and_return(Time.now)
```
mjgiarlo, line 23: This commit also includes a bit of a test refactoring, replacing one mega-test with a bunch of smaller, more targeted tests. I can roll that part back if folks think what we had before was a better way to test this behavior.

PR #3436, opened 2023-12-14
Adds okcomputer check for rabbit queues

justinlittman:
# Why was this change made? 🤔
So that we are notified when queues disappear.

QA

### DIFF
--- a/config/initializers/okcomputer.rb
+++ b/config/initializers/okcomputer.rb
@@ -6,12 +6,37 @@ OkComputer.check_in_parallel = true
 # Required
 OkComputer::Registry.register "ruby_version", OkComputer::RubyVersionCheck.new
 OkComputer::Registry.register "redis", OkComputer::RedisCheck.new(url: ENV.fetch("REDIS_URL", Settings.redis_url))
-if Settings.rabbitmq.enabled
-  OkComputer::Registry.register "rabbit",
-    OkComputer::RabbitmqCheck.new(hostname: Settings.rabbitmq.hostname,
+
+class RabbitQueueExistsCheck < OkComputer::Check
+  attr_reader :queue_names, :conn
+
+  def initialize(queue_names)
+    @queue_names = Array(queue_names)
+    @conn = Bunny.new(hostname: Settings.rabbitmq.hostname,
       vhost: Settings.rabbitmq.vhost,
       username: Settings.rabbitmq.username,
       password: Settings.rabbitmq.password)
+  end
+
+  def check
+    conn.start
+    status = conn.status
+    missing_queue_names = queue_names.reject { |queue_name| conn.queue_exists?(queue_name) }
+    if missing_queue_names.empty?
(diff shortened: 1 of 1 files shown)


config/initializers/okcomputer.rb
```diff
@@ -6,12 +6,37 @@
 # Required
 OkComputer::Registry.register "ruby_version", OkComputer::RubyVersionCheck.new
 OkComputer::Registry.register "redis", OkComputer::RedisCheck.new(url: ENV.fetch("REDIS_URL", Settings.redis_url))
-if Settings.rabbitmq.enabled
-  OkComputer::Registry.register "rabbit",
-    OkComputer::RabbitmqCheck.new(hostname: Settings.rabbitmq.hostname,
+
+class RabbitQueueExistsCheck < OkComputer::Check
+  attr_accessor :queue_names, :conn
```
mjgiarlo, line 11: Not a big deal: would an `attr_reader` suffice?

config/initializers/okcomputer.rb
```diff
@@ -6,12 +6,37 @@
...
+
+  def initialize(queue_names)
+    self.queue_names = Array(queue_names)
+    self.conn = Bunny.new(hostname: Settings.rabbitmq.hostname,
```
mjgiarlo, line 15: I was a bit surprised to see `self.` here instead of `@`. I don't think there's a functional difference in this case.

config/initializers/okcomputer.rb
```diff
@@ -6,12 +6,37 @@
...
+    conn.start
+    status = conn.status
+    missing_queue_names = queue_names.reject { |queue_name| conn.queue_exists?(queue_name) }
+    if missing_queue_names.empty?
+      mark_message "'#{queue_names.join(", ")}' exists, connection status: #{status}"
+    else
+      mark_message "'#{missing_queue_names.join(", ")}' does not exist"
+      mark_failure
+    end
+    conn.close
```
mjgiarlo, line 31: Should `conn.close` be within a method-level `ensure` block to make sure the connection is always closed?

config/initializers/okcomputer.rb
```diff
@@ -6,12 +6,37 @@
...
+  rescue
+    mark_message "Error: '#{e}'"
```
mjgiarlo, line 33: Where is `e` set?

config/initializers/okcomputer.rb
justinlittman: I was just cribbing some code, but maybe I cribbed too much.

config/initializers/okcomputer.rb
justinlittman: Not sure. It isn't in https://github.com/sportngin/okcomputer/blob/master/lib/ok_computer/built_in_checks/rabbitmq_check.rb

mjgiarlo (approved):

PR #3442, opened 2024-01-04
Revert to plain-old rubo.

### LINKED ISSUE #3373: remove standard gem (revert to plain rubocop)
remove the standard gem in favor of plain rubocop

per discussion at dev planning https://docs.google.com/document/d/1gzKgbqnlCc_GctBBpoYTvZl22mNZflsbqzyz4xXJlho/edit#heading=h.hmax9277viw

Consider what, besides the bare minimum, to put in `.rubocop.yml` as:
1. we may want to explicitly have some non-default settings for rules (e.g. line length)
2. if there is a lot of stuff in logs or tmp or vendor or whatever, running rubocop can be much slower since it looks at every directory.

With this in mind, I recommend doing at least one of the following:
1. see https://github.com/sul-dlss/DeveloperPlaybook/blob/main/style/infra_template.rubocop.yml
2. spelunk for the pull request that switched to standard and consider the old rubocop.yml file from that, or, at a minimum, 
3. look at another app with a UI, such as ETD or PreAssembly and take appropriate settings from that. (Note that Argo used standard for a while and just got rid of it with sul-dlss/argo/pull/4200 ... and required follow on sul-dlss/argo/issues/4214 to address these very concerns)

justinlittman:
closes #3373

# Why was this change made? 🤔
!standard


.rubocop.yml
```diff
@@ -1,26 +1,673 @@
...
-  - rubocop-capybara
-  - rubocop-factory_bot
   - rubocop-performance
   - rubocop-rails
   - rubocop-rspec
+  - rubocop-capybara
+  - rubocop-factory_bot
+
+# Don't even think about it.
+# inherit_from: .rubocop_todo.yml
```
ndushay, line 9: can you say something about why?   This isn't terribly helpful.  Also ... does it need to be first in the file?

ndushay (commented):
anything beyond double quotes to single quotes to call out?   Otherwise, this is way too big to review easily.

.rubocop.yml
justinlittman: Removed.

justinlittman:
@ndushay Nothing too interesting. I reformatted some long lines and added various `rubocop:disable` instead of indiscriminate `.rubocop_todo`.

ndushay:
> @ndushay Nothing too interesting. I reformatted some long lines and added various `rubocop:disable` instead of indiscriminate `.rubocop_todo`.

Did you know there's an option you can use so rubocop_todo shows the offending files?  `rubocop --auto-gen-config --auto-gen-only-exclude`

https://docs.rubocop.org/rubocop/configuration.html#automatically-generated-configuration

justinlittman:
Yes, but once we have a `.rubocop_todo` available, we tend to get lazy. For new codebases, I'd suggest that `.rubocop_todo` is an anti-pattern.

ndushay:
> Yes, but once we have a `.rubocop_todo` available, we tend to get lazy. For new codebases, I'd suggest that `.rubocop_todo` is an anti-pattern.

only when we agree with the rules.  :-P

ndushay (approved):

PR #3460, opened 2024-02-02
[WIP] validate the user has attached a zip file when selecting zip option

### LINKED ISSUE #3459: Validate that ZIP file is added when saving with ZIP upload option
See #3454 for related work which fixes some situations where the ZIP upload job can get stuck.  While this should not occur anymore, there is still an edge case which can lead to unexpected behavior.

Basically the way we have it setup right now, a user can:

1. Select browser upload and upload some files, including a ZIP file.
2. Save the draft.
3. Edit the draft, select the "ZIP upload" option, and then save the draft again without actually adding a file.

If the last file attached in step 1 happened to be a ZIP file, the ZIP Upload Job will unzip this file and replace all of the other files in the object.  This is the expected behavior for ZIP Uploads, but in this case may be unexpected, since the user never actually supplied a new ZIP after selecting that option.

So, this new validation would prevent the user from submitting the form (even a draft save) if they have selected ZIP Upload but then not provided a ZIP file.

Pending text of validation error message.

peetucket:
# Why was this change made? 🤔

Fixes #3459 - the user needs to add a ZIP file when selecting that option each time they save the page or else unexpected things might happen.

Todo: figure out how to get the spec to work (to mimic a user uploading a new ZIP file)

Localhost, need to get specs working


app/jobs/unzip_job.rb
```diff
@@ -36,11 +36,7 @@ def perform(work_version)
 
   # determines if we should actually unzip this file, based on it not being in globus and being a zip file
   def unzip_file?(zip_attached_file)
-    !zip_attached_file.in_globus? && zip_mime_types.include?(zip_attached_file.content_type)
-  end
-
-  def zip_mime_types
-    ['application/zip', 'application/x-zip-compressed', 'application/x-zip', 'multipart/x-zip']
+    !zip_attached_file.in_globus? && zip_attached_file.zip?
   end
```
peetucket, line 40: move this logic from the job to the attached file class so we can have access to it in our new validation

app/forms/base_work_form.rb
```diff
@@ -47,6 +47,7 @@ class BaseWorkForm < Reform::Form
   validates :work_type, presence: true, work_type: true
   validate :unique_filenames
   validate :globus_files_provided, if: proc { |form| form.fetch_globus_files }
+  validate :zip_file_provided_on_save, if: proc { |form| form.upload_type == 'zipfile' }
```
mjgiarlo, line 50:
```suggestion
  validate :zip_file_provided_on_save, if: proc { |form| form.needs_zipfile_validation? }
```

app/forms/base_work_form.rb
```diff
@@ -110,6 +111,27 @@ def globus_files_provided
...
+  # see https://github.com/sul-dlss/happy-heron/issues/3459
+  def zip_file_provided_on_save
+    # no need to validate this if there aren't already uploaded files in the object
+    return if attached_files.empty? || work.head&.attached_files&.empty?
+
+    # validation passes if the last file uploaded is a zip file
+    return if last_file_uploaded_is_zip?
+
+    errors.add(:base, I18n.t('work.zip_upload_option_validation'))
+  end
```
mjgiarlo, line 129:
```suggestion
  def needs_zipfile_validation?
    upload_type == 'zipfile' && (attached_files.any? || work.head&.attached_files&.any?)
  end
  
  def zip_file_provided_on_save
    # validation passes if the last file uploaded is a zip file
    return if last_file_uploaded_is_zip?

    errors.add(:base, I18n.t('work.zip_upload_option_validation'))
  end
```

mjgiarlo (commented):
Looks good! I have dropped a couple suggestions on the `BaseWorkForm` that I think may make clearer which parts of the zipfile validation machinery are about checking pre-conditions and which are the validation itself. LMK if I misunderstood!

mjgiarlo (approved):

PR #3463, opened 2024-02-05
Fix display of collection move event

### LINKED ISSUE #3462: Add and correct history table entries when items are moved to another collection
1. When an item is moved to another collection, an entry is logged in the history table for that item at the bottom of the view page, but it lacks a label in the left column.

See https://sdr.stanford.edu/works/5242 for an example, also screenshotted below.

Screen Shot 2024-02-02 at 2 14 34 PM

2. In the case shown above, where the "description of changes" should include the title of the collection, this one is showing the description of the collection instead of the title. 

However, this is not happening in all cases. See https://sdr.stanford.edu/works/2759 (or works 3180, 6626, or 6584), which are missing the label as above but have the correct collection title in the "description of changes."

Screenshot 2024-02-05 at 10 26 48 AM

peetucket:
# Why was this change made? 🤔

Fixes #3462:

- The locale file was missing a description of the collection moved event, I set it to "Changed collection".  This will fix all history event displays once deployed.
- The name of the collection moved to in the event description was set to use the collection "description" field instead of the "name" (which is the title).  I changed it to use "name" going forward.  This will **not** fix previous entries unless we change them in the database manually for each event.

The reason it appeared to work OK for others is simply because the name and the description are sometimes very similar, e.g. 

"Undergraduate Theses, Materials Science and Engineering" == name/title
"Honors theses for undergraduates majoring in Materials Science and Engineering." == description

When done, go onto https://github.com/sul-dlss/happy-heron/issues/3464

On Stage

### DIFF
--- a/app/controllers/work_move_controller.rb
+++ b/app/controllers/work_move_controller.rb
@@ -34,7 +34,7 @@ class WorkMoveController < ApplicationController
         collection.save!
         work.update!(collection:)
         work.events.create(user: current_user, event_type: 'collection_moved',
-                           description: "Moved to \"#{collection.head.description}\" collection")
+                           description: "Moved to \"#{collection.head.name}\" collection")
       end
       flash[:success] = "Moved #{work.head.title} to #{collection.head.name}"
     end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -26,7 +26,7 @@ en:
       unlock_work: Work unlocked
       decommission: Decommissioned
       no_review_workflow: Review no longer required
-      collection_moved: ''
+      collection_moved: Changed collection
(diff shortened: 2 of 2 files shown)


amyehodge:
I tested moving an item to another collection on stage and this worked fine. I didn't have any previously moved items, so I couldn't test whether the label shows up for all the older items that had been moved or not. 

It would be great to fix the descriptions for all the existing moved items.

peetucket:
> It would be great to fix the descriptions for all the existing moved items.

Created a new ticket for the remediation: #3464 ... shouldn't be hard (see details in the ticket)

edsu (approved):

PR #3483, opened 2024-03-20
Ignore mac Icon files.

### LINKED ISSUE #3477: Exclude MacOS "Icon\r" files from deposits
We should exclude MacOS `Icon\r` files from deposits, just like we exclude `.DS_Store` files. These files cause a number of problems:

- they have a control character `\r` (carriage return) in the name, which causes issues with parsing checksum manifests (see: https://github.com/sul-dlss/preservation_robots/issues/540 )
- they use MacOS filesystem resource forks and so do not copy properly to non-Mac systems
- probably because of the resource fork issue, they appear to end up as empty files when copied to a Linux filesystem
- they clutter the file list on the Purl/sul-embed

(For more background, I found some info [here](https://superuser.com/questions/298785/icon-file-on-os-x-desktop).)

Excluding these files won't affect the data people and will make our lives easier, as we spent a considerable amount of time recently trying to get a deposit accessioned that came with `Icon\r` files.

justinlittman:
closes #3477

# Why was this change made? 🤔
Files are not necessary and clutter things up.

Unit, Andrew.

### DIFF
--- a/app/services/ignore_file_service.rb
+++ b/app/services/ignore_file_service.rb
@@ -5,6 +5,6 @@ class IgnoreFileService
   def self.ignore?(filename)
     filename.start_with?('__MACOSX', '._') \
     || File.basename(filename).start_with?('__MACOSX', '._') \
-    || filename.end_with?('.DS_Store')
+    || filename.end_with?('.DS_Store', "Icon\r")
   end
 end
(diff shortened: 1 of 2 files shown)


jmartin-sul (approved):

mjgiarlo:
@justinlittman @jmartin-sul Still on hold or is this shippable?

jmartin-sul:
> @justinlittman @jmartin-sul Still on hold or is this shippable?

Guessing (vaguely recalling) that the hold was for testing by Andrew.  No other reason I can think of to hold.

PR #3502, opened 2024-05-01
Replace byebug with ruby/debug

jcoyne:
# Why was this change made? 🤔
Rails 7 replaced byebug with ruby/debug


README.md
```diff
@@ -57,20 +58,12 @@ To get the config values from vault, see the [Globus client README](https://gith
...
-
-And then start up the debugger in another window (only byebug is supported at this time):
-
-```
-bundle exec byebug -R localhost:8989
+bin/dev
 ```
 
-Note that, by default, the debugger will run on `localhost` on port `8989`. To change these values, add the `DEBUGGER_HOST` and `DEBUGGER_PORT` environment variables when invoking `bin/dev` above and make sure to poing `byebug -R` at those same values when starting up the debugger.
-
```
mjgiarlo, line 70: Worth including a note about how to use rdbg when running `bin/dev`? Maybe we can point to documentation elsewhere.

jcoyne: If we wanted to do that we'd have to start rdbg in the Procfile like in argo.

mjgiarlo: Cool. Would you consider adding that to this PR?

jcoyne: added

mjgiarlo: Much obliged. How about adding a README line around here like we do in Argo:

> The webserver process is configured to run a remote debug session. You may attach (`-A`) the debugger in another terminal window:
> 
> ```
> rdbg -A
> ```

❓

jcoyne: Does it make sense to explain this in every project, or should we add something about how to use a debugger in https://github.com/sul-dlss/DeveloperPlaybook?

mjgiarlo: @jcoyne Covering this in the playbook is better (IMO) than copypasta in lots of READMEs, yeah.

mjgiarlo (approved):

mjgiarlo (approved):

PR #3507, opened 2024-05-01
Add migration to specify WorkVersion#user_version

### LINKED ISSUE #3499: Add user_version attribute to WorkVersion
* Add user_version as an optional integer attribute of WorkVersion.
* Create a unique index on user_version and work_id.
* For all new works, default to 1 for a Work’s first WorkVersion.
* For all existing Works, set to 1 for head version.

jcoyne:
# Why was this change made? 🤔

Fixes #3499

### DIFF
--- a/db/migrate/20240501135224_add_user_version_to_work_versions.rb
+++ b/db/migrate/20240501135224_add_user_version_to_work_versions.rb
@@ -1,0 +1,7 @@
+class AddUserVersionToWorkVersions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :work_versions, :user_version, :integer
+    add_index :work_versions, [:work_id, :user_version], unique: true
+    Work.all.each { |work| work.head.update(user_version: 1) }
+  end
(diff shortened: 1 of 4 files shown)


db/migrate/20240501135224_add_user_version_to_work_versions.rb
```diff
@@ -0,0 +1,7 @@
+class AddUserVersionToWorkVersions < ActiveRecord::Migration[7.0]
+  def change
+    add_column :work_versions, :user_version, :integer
+    add_index :work_versions, [:work_id, :user_version], unique: true
+    Work.all.each { |work| work.head.update(user_version: 1) }
```
mjgiarlo, line 5: Does this make the migration irreversible? Maybe it's not worth caring about?

jcoyne: Yes, we'd have to first comment this line out and then do `db:rollback`

PR #3523, opened 2024-05-08
Set the user version on the prior version on destroy if nil

### LINKED ISSUE #3515: Update user version when discarding a draft.
When discarding a draft work version and the draft work version has a user_version:
* If the previous work version has a user_version do nothing.
* If the previous work version does not have a user_version, move the user_version from the draft work_version to the previous work_version.

aaron-collier:
# Why was this change made? 🤔

Fixes #3515 


justinlittman (changes requested):
I think this is going to be obviated with the new approach for user versions, so let's hold.

PR #3527, opened 2024-05-13
Change approach to persisting user_versions.

### LINKED ISSUE #3506: Update WorkVersion.user_version when submitting work form
This ticket involves changing the approach for persisting user versions. Currently, user_version of an attribute of a WorkVersion with a unique index on id, user_version. So, for example:

| Work Version | user_version | 
| --- | --- |
1 | |
2 | 1 |
3 | |
4 | 2 |

In this case, user version 1 is work version 2 and user version 2 is work version 4.

In the new approach, every WorkVersion will have a user_version (except maybe the current draft).
| Work Version | user_version | 
| --- | --- |
1 | 1 |
2 | 1 |
3 | 2 |
4 | 2 |

To find the Work Version for a user version, you find the highest WorkVersion with the user_version value. In this case, user version 1 is work version 2 and user version 2 is work version 4. (Though there aren't currently any features that require this.)

This means that discarding a draft doesn't require making any changes to user_version values. If WorkVersion 4 would be discarded then work version 3 would become user version 2:

| Work Version | user_version | 
| --- | --- |
1 | 1 |
2 | 1 |
3 | 2 |

This also allows it to be determined which radio button to select when displaying the new version section:
image

When a draft WorkVersion is created, no radio button should be selected. While the user will have to make a choice before depositing, they can still save a draft without making a choice. No choice is represented by a null. Thus, if there is no user_version, select no radio buttons:
| Work Version | user_version | 
| --- | --- |
1 | 1 |
2 | 1 |
3 | 2 |
4 |  |

If the user_version is the same as the last WorkVersion, then select the "No" radio:
| Work Version | user_version | 
| --- | --- |
1 | 1 |
2 | 1 |
3 | 2 |
4 | 2  |

If the user_version is different than the last WorkVersion, then select the "Yes" radio:
| Work Version | user_version | 
| --- | --- |
1 | 1 |
2 | 1 |
3 | 2 |
4 | 3  |

Thus, the work for this ticket includes (but may not be limited to):
1. Remove the unique index.
2. Populate user_version for existing WorkVersions. (The value is probably just 1.)
3. When creating a new WorkVersion, set user_version to nil, except the first WorkVersion which should be 1.
4. If feature flag is set, populate the version section radio button and version description field as described above.
5. If feature flag is set, when saving a WorkVersion, update the user_version based on the radio button. (If "yes", increment the user_version of the previous WorkVersion. If "no", use the user_version of the previous WorkVersion. If neither, set to nil.)
6. If the feature flag is not set, when saving a WorkVersion, set the user_version to the user_version of the previous WorkVersion.
7. Make sure that there are enough code comments and/or README docs to document this approach for our future selves.

justinlittman:
closes #3506

# Why was this change made? 🤔
* Make it easier to manage user versions.
* Allow user versions to be persisted and synced with form.

Unit, Amy

### DIFF
--- a/app/components/works/buttons_component.html.erb
+++ b/app/components/works/buttons_component.html.erb
@@ -31,17 +31,9 @@
 
   <div class="col-auto">
     <%= link_to 'Cancel', cancel_link_location, class: 'btn btn-link' %>
-    <% if validate_user_version? %>
-        <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button',
-                                         data: { action: 'unsaved-changes#allowFormSubmission new-user-version#validateUserVersionSelection' } %>
-        <%= form.submit submit_button_label, class: 'btn btn-primary',
-                                             disabled: false,
-                                             data: { action: 'unsaved-changes#allowFormSubmission new-user-version#validateUserVersionSelection', deposit_button_target: 'depositButton' } %>
-    <% else %>
-      <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button',
-                                       data: { action: 'unsaved-changes#allowFormSubmission' } %>
-      <%= form.submit submit_button_label, class: 'btn btn-primary', disabled: false,
-                                           data: { action: 'unsaved-changes#allowFormSubmission', deposit_button_target: 'depositButton' } %>
-   <% end %>
+    <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button',
+                                     data: { action: 'unsaved-changes#allowFormSubmission' } %>
+    <%= form.submit submit_button_label, class: 'btn btn-primary', disabled: false,
+                                         data: { action: 'unsaved-changes#allowFormSubmission', deposit_button_target: 'depositButton' } %>
   </div>
 </div>
--- a/app/components/works/version_description_component.html.erb
+++ b/app/components/works/version_description_component.html.erb
@@ -1,47 +1,52 @@
-<% if Settings.user_versions_ui_enabled %>
-  <section class='version'>
-    <header>Do you want to create a new version of this deposit? *</header>
-    <div>
-      <div class='invalid-feedback' data-new-user-version-target='versionDescriptionError'></div>
-      <%= form.hidden_field :version_description, value: work_version.version_description, data: { new_user_version_target: 'versionDescription' } %>
-      <div class='mb-3 row'>
-        <div class='col-sm-10 form-check'>
-          <%= form.radio_button :new_user_version, 'yes', data: { action: 'click->new-user-version#displayVersionDescription', new_user_version_target: 'userVersionYes' }, class: 'form-check-input' %>
-          <%= form.label :new_user_version, 'Yes', value: 'yes', class: 'form-check-label fw-semibold' %>
-          <ul>
-            <li>Edit files and any or all form fields.</li>
-            <li>A new version of this deposit will be created. For example, if you are editing version 1, choosing this option will create a version 2 once you click “Deposit.”</li>
-            <li>The new version and any older versions will be accessible from the PURL page.</li>
-            <li>If this deposit has a DOI, the DOI will be the same for all versions of the item.</li>
-          </ul>
-          <%= form.label :new_user_version_description, "What's changing?", class: 'col-form-label' %>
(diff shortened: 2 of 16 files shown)


app/components/works/buttons_component.html.erb
```diff
@@ -31,17 +31,9 @@
...
-        <%= form.submit submit_button_label, class: 'btn btn-primary',
-                                             disabled: false,
-                                             data: { action: 'unsaved-changes#allowFormSubmission new-user-version#validateUserVersionSelection', deposit_button_target: 'depositButton' } %>
-    <% else %>
-      <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button',
-                                       data: { action: 'unsaved-changes#allowFormSubmission' } %>
-      <%= form.submit submit_button_label, class: 'btn btn-primary', disabled: false,
-                                           data: { action: 'unsaved-changes#allowFormSubmission', deposit_button_target: 'depositButton' } %>
-   <% end %>
+    <%= form.submit 'Save as draft', class: 'btn btn-primary', id: 'save-draft-button',
```
justinlittman, line 34: Altered approach so that don't need to do special validation in js controller.

app/components/works/version_description_component.html.erb
```diff
@@ -1,47 +1,52 @@
...
-        <%= form.text_field :version_description, class: 'form-control', required: true %>
-        <div class='invalid-feedback'>You must describe your changes.</div>
-      </div>
-    </div>
-  </section>
+    </section>
+  <% end %>
+<% end %>
+<% if show_hidden? %>
+  <%= form.hidden_field :new_user_version, value: 'no' %>
```
justinlittman, line 51: There will always be a `new_user_version` param (either selected by the user or hidden).

app/components/works/version_description_component.rb
```diff
@@ -13,8 +13,12 @@ def work_version
       form.object.model.fetch(:work_version)
     end
 
-    def render?
```
justinlittman, line 16: Always rendering.

app/controllers/works_controller.rb
```diff
@@ -72,9 +72,7 @@ def update
     clean_params = orig_clean_params.deep_dup
 
     if work_version.deposited?
-      # Here we are moving the user_version from the previous WorkVersion to the head WorkVersion.
       work_version = create_new_version(work_version)
-      orig_work_version.update(user_version: nil)
```
justinlittman, line 77: No longer setting the previous work version to nil.

app/forms/base_work_form.rb
```diff
@@ -13,9 +13,16 @@ class BaseWorkForm < Reform::Form
 
   property :work_type, on: :work_version
   property :version_description, on: :work_version
-  property :new_user_version, virtual: true # in the future, will need to be persisted to the work_version for drafts
-  property :new_user_version_description, virtual: true
```
justinlittman, line 17: No longer need separate properties for this version descriptions.

app/javascript/controllers/new_user_version_controller.js
```diff
@@ -4,20 +4,30 @@ export default class extends Controller {
   static targets = ['versionDescription', 'userVersionYes', 'userVersionNo', 'versionDescriptionYes', 'versionDescriptionNo', 'versionDescriptionError']
 
   connect () {
-    this.versionDescriptionYesTarget.disabled = true
-    this.versionDescriptionNoTarget.disabled = true
-    this.versionDescriptionYesTarget.value = this.versionDescriptionTarget.value
-    this.versionDescriptionNoTarget.value = this.versionDescriptionTarget.value
+    if (this.userVersionYesTarget.checked === true) {
```
justinlittman, line 7: Now setting `required` for client-side validation.

app/components/works/version_description_component.rb
```diff
@@ -13,8 +13,12 @@ def work_version
...
+    def show?
       %w[deposited rejected version_draft].include? work_version.state
     end
+
+    def show_hidden?
```
lwrubel, line 20: I'm finding this method name confusing. (Is the show section hidden? Or are we showing something hidden?) I can't figure out what a good name would be however, something like `new_version_not_applicable?` or `user_version_na?` or `use_fallback_user_version?`

lwrubel (approved):
Approved pending consideration of question about method name

PR #3532, opened 2024-05-14
Disable files panel when selecting no

### LINKED ISSUE #3505: Disable files section when user selects to not create a new version
If the user has selected nothing, the file section should be enabled.
If the user has selected "no", the file section should be disabled.
If the user has selected "yes", the file section should be enabled.

See #3504 for the addition of the new version section.

Please verify that when a disabled file section is changed, the files remain unchanged.

lwrubel:
# Why was this change made? 🤔
Resolves #3505. HOLD for testing in deployed env and by PO. Please squash when merging!

Screenshot 2024-05-21 at 4 16 06 PM

Locally, deployed to QA for PO testing

No new ones


lwrubel:
**Bug 1:**
@amyehodge has found a problem with attached files that I think is related to initially saving a draft with neither Yes/No select. 

1) Edit a deposited work
2) Don't select Yes or No. Save the draft
3) Edit it again and select Yes. Add a file. 
4) HB error: https://app.honeybadger.io/projects/55564/faults/100610329

I am not sure yet of the source of the error, but I think it has to do with the `use_previous_attached_files` method in the `base_work_form.rb` and it attaching (empty?) files that don't get checksummed later. 

I am not sure why we removed logic to require users to select Yes/No even when creating a draft. It seems like that would be good to do anyway, to avoid having a nil `new_user_version` param  saving `nil` `user_version` numbers into the work_version.

**Bug 2**
If you don't select Yes/No, save the draft, and then re-open the draft, the previous version's version description is getting displayed:

Screenshot 2024-05-22 at 4 29 15 PM

lwrubel:
I made a change in the last commit to require users to select Yes/No for a new version even in draft mode.  Forcing a selection seems to prevent the bugs above, pending further testing in QA. The styling could use a tweak (see below) to be bolder. 

If this does not match with the desired UX, we will need to find a different solution. 

Screenshot 2024-05-22 at 5 21 37 PM

justinlittman (approved):

PR #3533, opened 2024-05-14
Add version to auto citation.

### LINKED ISSUE #3506: Update WorkVersion.user_version when submitting work form

### LINKED ISSUE #3521: Add version information to automated preferred citation
When user_versions_ui_enabled feature flag is enabled:

When version 1 of an item is created, include "Version 1" in the preferred citation. This should be inserted after the title. Also, use the /v1 URL. Example:

> Weaver, K.J. (2020). Juxtaposing Stakeholder Perceptions of Diversity in Higher Education Law. Version 1. Stanford Digital Repository. Available at https://purl.stanford.edu/qv180wz3234/v1.

When subsequent versions are made, the automated citation should be updated to reflect the new version number, both after the title and in the URL. Example:

> > Weaver, K.J. (2020). Juxtaposing Stakeholder Perceptions of Diversity in Higher Education Law. Version 2. Stanford Digital Repository. Available at https://purl.stanford.edu/qv180wz3234/v2.

The intent of this change is to help prevent a situation where someone uses the base purl (purl.stanford.edu/druid) for a citation, then later creates more user versions. At this point, it will be unclear which version the base version was intended to point to. So, we want the /v1 version of the purl (purl.stanford.edu/druid/v1) to be available when the first user version is created and to be inserted in the preferred citation from the start.

justinlittman:
closes #3521

# Why was this change made? 🤔
Per ticket.

Locally, Amy

### DIFF
--- a/app/components/works/form_component.html.erb
+++ b/app/components/works/form_component.html.erb
@@ -4,7 +4,9 @@
                 controller: data_controllers,
                 action: 'change->unsaved-changes#changed beforeunload@window->unsaved-changes#leavingPage turbo:before-visit@window->unsaved-changes#leavingPage',
                 auto_citation_purl: purl,
-                auto_citation_doi: doi_field
+                auto_citation_doi: doi_field,
+                auto_citation_user_version_value: user_version,
+                auto_citation_user_versions_ui_enabled_value: user_versions_ui_enabled?
               },
               html: { class: 'needs-validation work-editor', novalidate: true, multipart: true } do |f| %>
   <% if work_form.errors.present? %>
--- a/app/components/works/form_component.rb
+++ b/app/components/works/form_component.rb
@@ -33,6 +33,12 @@ module Works
       work_form.model.fetch(:work_version)
     end
 
+    delegate :user_version, to: :work_version
+
+    def user_versions_ui_enabled?
+      Settings.user_versions_ui_enabled
+    end
+
     def doi_field
       return "https://doi.org/#{doi}." if doi
 
@@ -45,7 +51,7 @@ module Works
     end
 
     def data_controllers
-      return 'auto-citation unsaved-changes deposit-button new-user-version' if Settings.user_versions_ui_enabled
+      return 'auto-citation unsaved-changes deposit-button new-user-version' if user_versions_ui_enabled?
 
       'auto-citation unsaved-changes deposit-button'
     end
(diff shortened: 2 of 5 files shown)


justinlittman:
When "no":
image

image

justinlittman:
When "yes":
image

image

mjgiarlo (approved):

PR #3585, opened 2024-07-26
Show previous user version purl links

### LINKED ISSUE #3582: Add links to item view page for previous user versions of a work
This includes the following changes to the Details table at the top of an item view page:

- change existing "Version details" label to say "Current version". Note that the content displayed in the second column is not changing. 
- add a new row to the table below the "Current version" row with the label "Previous version(s)"
- display a list of numbers for the previous versions, starting with the most recent and counting back to 1, separated by commas. Each number should be linked to the PURL for that specific version

Screenshot 2024-07-24 at 2 14 41 PM

lwrubel:
# Why was this change made? 🤔
Resolves #3582 to show links to purls for previous versions. Does not change the current prod view where user versions are not enabled. HOLD for PO review.

When there is only one user version, no Previous versions row displays:

Screenshot 2024-07-26 at 10 31 14 AM

When previous user versions exist: 

Screenshot 2024-07-26 at 10 35 03 AM

Unit and testing on QA.

### DIFF
--- a/app/components/works/detail_component.rb
+++ b/app/components/works/detail_component.rb
@@ -115,7 +115,7 @@ module Works
     end
 
     def version_purl(version)
-      purl + "/v/#{version}"
+      purl + "/v#{version}"
     end
 
     private
--- a/app/components/works/detail_component.rb
+++ b/app/components/works/detail_component.rb
@@ -104,6 +104,22 @@ module Works
       work.last_rejection_description
     end
 
+    def previous_user_versions?
+      return false unless Settings.user_versions_ui_enabled
+
+      return true if work_version.user_version > 1
+
+      false
+    end
+
+    def previous_user_versions
+      (1..work_version.user_version - 1).to_a.reverse
+    end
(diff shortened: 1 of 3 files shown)


app/components/works/detail_component.rb
```diff
@@ -104,6 +104,22 @@ def rejection_reason
       work.last_rejection_description
     end
 
+    def previous_user_versions?
+      return false unless Settings.user_versions_ui_enabled
+
+      return true if work_version.user_version > 1
```
justinlittman, line 110: L110-112 could just be `work_version.user_version > 1`

justinlittman (approved):
Approve pending consideration of comment and approval by Amy.

app/components/works/detail_component.rb
lwrubel: Fixed, thanks.

lwrubel:
Merging after @amyehodge tested H2 and approved.

PR #3588, opened 2024-07-29
Upgrade datatablesi and relative time.

### LINKED ISSUE #3583: Clicking on the arrows to sort the deposits table by Modified data should work
Currently clicking these arrows reloads the table, but the sort order does not change. Other sorting seems to be working just fine for this table, it's just this one column. It doesn't appear to be sorting in either direction properly.

This is https://sdr.stanford.edu/collections/195/works as an example, with the two sort order options selected for modified date, but they look the same:

Screenshot 2024-07-24 at 3 01 29 PM

Screenshot 2024-07-24 at 3 01 37 PM

justinlittman:
closes #3583

# Why was this change made? 🤔
Datatables weren't sorting correctly.

Unit, stage


app/components/collections/work_row_component.html.erb
```diff
@@ -6,7 +6,7 @@
   </td>
   <td><%= work.owner.sunetid %></td>
   <td><%= render Works::StateDisplayComponent.new(work_version:) %></td>
-  <td><%= render LocalTimeComponent.new(datetime: work.updated_at, show_time: false) %></td>
+  <td><%= work.updated_at.strftime("%b %e, %Y") %></td>
```
justinlittman, line 9: local time / relative time simply won't work with data tables.

app/components/collections/works_component.html.erb
```diff
@@ -5,7 +5,7 @@
       <th><span class="visually-hidden"><%= I18n.t 'collection.actions' %></span></th>
       <th width="10%"><%= I18n.t 'collection.owner' %></th>
       <th><%= I18n.t 'collection.deposit_status' %></th>
-      <th data-type="date" data-format="MMM D, YYYY" aria-sort="descending"><%= I18n.t 'collection.last_modified' %></th>
+      <th data-type="date" aria-sort="descending"><%= I18n.t 'collection.last_modified' %></th>
```
justinlittman, line 8: data-format not necessary here.

app/components/local_time_component.rb
```diff
@@ -1,15 +1,15 @@
 # frozen_string_literal: true
 
-# Draws a custom element for https://github.com/github/time-elements
+# Draws a custom element for https://github.com/github/relative-time-element
```
justinlittman, line 3: Updated this to use relative-time-element, as time-elements is deprecated.

app/javascript/controllers/datatable_works_controller.js
```diff
@@ -20,13 +20,13 @@ export default class extends Controller {
 
     // This scrolls the top of the table into view when paging.
     dt.on('datatable.page', () => {
-      dt.table.scrollIntoView()
+      dt.dom.scrollIntoView()
```
justinlittman, line 23: Changes here for upgrade from datatables v3 to 9.

app/components/collections/works_component.html.erb
mjgiarlo: @justinlittman Removing this seems to have broken sorting by dates (see: #3592). When I put it back, date sorting works again. Can you recall why you removed this attr?

justinlittman: I believe that according to the documentation, the date format we were using would be recognized, so `data-format` wasn't necessary. I thought I had tested, but apparently not successfully.

mjgiarlo: You mean we're not the only software team that is challenged by documentation updates? Huh!

PR #3589, opened 2024-07-30
Fix to editing reservations.

### LINKED ISSUE #3572: Setting work type for a PURL reservation fails from the work show page
Once you reserve a PURL, you need to set the work type to edit the work. There is a modal that shows in this situation before you get to the work page.  In fixing this issue: https://github.com/sul-dlss/happy-heron/issues/3570 (users potentially going directly to the edit work URL), we discovered that setting the work type from the work show page doesn't work, even though it works from the dashboard or the collection list page.

To reproduce:

1. Reserve a PURL within a given collection.
2. Visit the work detail page for your reserved PURL.
3. Click the pencil icon next to the title - you should get the work type modal.
4. Select a work type and click submit.  It will set the work type but you will get a blank page with a "Content missing" error.
5. Refresh the page and you are back on the Work detail page, and now you can edit it as normal.

If you follow the same process, but click the pencil icon for the reserved PURL on the dashboard, you will see it correctly redirects to the edit page after setting the work type.

Suspect something not working correctly in turbo.

This PR adds some test expectations in https://github.com/sul-dlss/happy-heron/blob/main/spec/features/purl_reservation_spec.rb to verify the error: https://github.com/sul-dlss/happy-heron/pull/3571/files  Uncomment them out when this is fixed

justinlittman:
closes #3572

# Why was this change made? 🤔

### DIFF
--- a/app/components/works/detail_component.html.erb
+++ b/app/components/works/detail_component.html.erb
@@ -238,5 +238,5 @@
       <% end %>
     </tbody>
   </table>
-  <%= render Works::WorkTypeModalComponent.new %>
+  <%= render Works::WorkTypeModalComponent.new(turbo: false, method: :patch, form_authenticity_token:) %>
 </section>
--- a/app/javascript/controllers/work_type_controller.js
+++ b/app/javascript/controllers/work_type_controller.js
@@ -18,7 +18,13 @@ export default class extends Controller {
     // icon, target is the span and currentTarget is the anchor.
     this.formTarget.action = event.currentTarget.dataset.destination
     if (event.currentTarget.dataset.formMethod) {
-      this.formTarget.method = event.currentTarget.dataset.formMethod
+      if (this.formTarget.dataset.turbo === 'false') {
+        // If this is a non-Turbo form, we need to set the method using a hidden input
+        this.formTarget.method = 'post'
+        this.formTarget.elements._method.value = event.currentTarget.dataset.formMethod
+      } else {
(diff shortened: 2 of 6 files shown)


app/components/works/detail_component.html.erb
```diff
@@ -238,5 +238,5 @@
       <% end %>
     </tbody>
   </table>
-  <%= render Works::WorkTypeModalComponent.new %>
+  <%= render Works::WorkTypeModalComponent.new(turbo: false, method: :patch, form_authenticity_token:) %>
```
justinlittman, line 241: Setting turbo to false since in a frame. This then requires setting the method and providing a form authenticity token so that the form works correctly.

app/javascript/controllers/work_type_controller.js
```diff
@@ -18,7 +18,13 @@ export default class extends Controller {
     // icon, target is the span and currentTarget is the anchor.
     this.formTarget.action = event.currentTarget.dataset.destination
     if (event.currentTarget.dataset.formMethod) {
-      this.formTarget.method = event.currentTarget.dataset.formMethod
+      if (this.formTarget.dataset.turbo === 'false') {
+        // If this is a non-Turbo form, we need to set the method using a hidden input
```
edsu, line 22: I didn't realize that `_method` could be used this way on the Rails side. Thanks!

https://guides.rubyonrails.org/form_helpers.html#how-do-forms-with-patch-put-or-delete-methods-work-questionmark

edsu (approved):
Thanks for untangling this one!

Issue #3581, opened 2024-07-22
FAST vocabulary type ahead lookup in H2 should work


amyehodge:
Currently this box in H2 behaves as a free text entry box. No type ahead search results appear when you type in the box.

`App 176307 output: W, [2024-07-22T09:40:55.455746 #176307]  WARN -- : [f24fa9fe-2590-42b5-ae17-5047e08d6a50] Autocomplete results for river returned 404`

edsu:
It might be helpful to see if this service is now available at a supported REST endpoint:

https://www.oclc.org/developer/api/oclc-apis/fast-api/assign-fast.en.html

edsu:
It seems like [OCLC's FAST user interface](http://fast.oclc.org/searchfast/) is now sending Solr requests like this:

http://fast.oclc.org/cgi-bin/perlProxy.pl?http://authfastdb-m1.prod.oclc.org/fastapps-db/fastIndex/select?q=keywords:(village)&rows=10&start=0&version=2.2&indent=on&fl=id,fullphrase,type,usage,status&sort=usage%20desc

amyehodge:
From OCLC FAST Team:

I believe our service is working okay but perhaps you are using an  ‘experimental.worldcat.org’ url? We recently discontinued that avenue in favor of [https://fast.oclc.org/searchfast/fastsuggest?&query=stuff](https://urldefense.com/v3/__https://fast.oclc.org/searchfast/fastsuggest?&query=stuff__;!!G92We9drHetJ8EofZw!dfbq6lpQY0mgwvf9UyvKiFYXXYiHe0BMN1gvFAyAFA7obTwr6Kun1S5--J2wmweJLIrFu5TLa0n4gCIC_2WR$). It should function the same, but if you run into any difficulties, please let us know.
 
Hope this helps,
Russell
 
 
-- 
Russell Schelby
OCLC, Technical Manager, Authorities
[schelbyr@oclc.org](mailto:schelbyr@oclc.org)

justinlittman:
@amyehodge We are using http://fast.oclc.org/fastIndex/select

justinlittman:
https://fast.oclc.org/searchfast/fastsuggest?&query=stuff doesn't help because it doesn't include any identifiers.

amyehodge:
@justinlittman I've sent another message to the FAST team about this.

edsu:
[...]

PR #3593, opened 2024-08-07
Alert Honeybadger when OCLC FAST API service returns errors

mjgiarlo:
Connects to #3581

# Why was this change made?

So we know when the keywords typeahead API goes down or is decommissioned.

# How was this change tested?

CI

### DIFF
--- a/app/controllers/fast_controller.rb
+++ b/app/controllers/fast_controller.rb
@@ -48,9 +48,10 @@ class FastController < ApplicationController
       req.params['sort'] = 'usage desc'
     end
 
-    return Failure("Autocomplete results for #{query} returned #{resp.status}") unless resp.success?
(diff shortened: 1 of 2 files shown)


justinlittman:
@mjgiarlo Has our FAST integration been fixed or is this going to immediately start reporting HB alerts?

mjgiarlo:
> @mjgiarlo Has our FAST integration been fixed or is this going to immediately start reporting HB alerts?

That's a separate PR that I'm hoping to get merged today or tomorrow. We're already over our limit of alerts for the month. I expect some new ones will come in, but probably not enough to be concerned about? And it'll be good to verify the alerts look right and have useful context.

edsu (approved):

PR #3599, opened 2024-08-08
Don't show dash when no version description

lwrubel:
# Why was this change made? 🤔

A version 1 user version may not have a version description. Instead of showing a dash next to empty space, hide the dash. (I know this wasn't ticketed, @amyehodge, but it's been bugging me while we've been working on versions and is easy to fix.) 

Before:

Screenshot 2024-08-08 at 12 38 57 PM

After:

Screenshot 2024-08-08 at 12 45 11 PM

If a work version and description created later:

Screenshot 2024-08-08 at 12 58 14 PM

### DIFF
--- a/app/components/works/detail_component.rb
+++ b/app/components/works/detail_component.rb
@@ -21,7 +21,9 @@ module Works
     def version
       return '1 - initial version' if first_draft?
 
-      "#{version_number} - #{version_description}"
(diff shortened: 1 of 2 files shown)


lwrubel:
I'll add a test to this to improve coverage.

mjgiarlo (approved):

PR #3603, opened 2024-08-15
work edit, file upload via browser: allow a user to re-add a file of the same name as one that was removed from the list

### LINKED ISSUE #3602: Users should be able to delete and then reupload a file with the same name without saving as draft
We seemed to have addressed this issue previously, but it's now not working again. See #1579. This may have also been impacted by later work on #2798.

What I'm seeing currently is the following: (testing on https://sdr.stanford.edu/works/5999)
1. Work has file called LongRoom.jpg.
2. Click pencil icon to edit work.
3. Click trash can icon next to file name to remove file.
4. Drag over the same file and it looks like it's uploading, but the bar is red and an error message displays that says "Unable to upload file due to: Duplicate file. Delete the file and try again." I am unable to save as draft in this state.
5. Click trash can to remove second file.
6. Save as draft.
7. Click pencil icon to edit.
8. Drag over the same file, which now works fine.

What the user should be able to do is drag over the replacement file with the same name in step 4 and that should work without error messages and they should then be able to save as draft.

ReplaceFile

jmartin-sul:
# Why was this change made? 🤔

fixes #3602

- augmented feature spec (commented out fix temporarily on localhost and confirmed the test failed because it couldn't find the expected success indicator when trying to re-add the same file after removal)
- manual testing in firefox and chrome, on localhost and stage

n/a

it shouldn't

### DIFF
--- a/app/javascript/controllers/dropzone_controller.js
+++ b/app/javascript/controllers/dropzone_controller.js
@@ -31,6 +31,12 @@ export default class extends Controller {
     }
   }
 
+  isVisible (elem) {
+    // cobbled together from a combo of Jeremy's solution in https://github.com/sul-dlss/happy-heron/pull/1582 (commit 27c81bc2a) and
+    // this stack overflow discussion, particularly this thread: https://stackoverflow.com/a/33456469
+    return !!(elem.offsetParent || elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length) && window.getComputedStyle(elem).visibility !== 'hidden'
+  }
+
   checkForDuplicates (fileName) {
     // Extract all filenames that are visible
     // Unfortunately, the filename targets are not contained in the scope of the controller.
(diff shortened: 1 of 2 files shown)


justinlittman (approved):

app/javascript/controllers/dropzone_controller.js
```diff
@@ -31,6 +31,12 @@ export default class extends Controller {
     }
   }
 
+  isVisible (elem) {
+    // cobbled together from a combo of Jeremy's solution in https://github.com/sul-dlss/happy-heron/pull/1582 (commit 27c81bc2a) and
+    // this stack overflow discussion, particularly this thread: https://stackoverflow.com/a/33456469
+    return !!(elem.offsetParent || elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length) && window.getComputedStyle(elem).visibility !== 'hidden'
```
mjgiarlo, line 37: @jmartin-sul I wonder if there's enough [browser support](https://caniuse.com/mdn-api_element_checkvisibility) now to consider the solution outlined in https://stackoverflow.com/a/72717388/24039837?

jmartin-sul: oh, possibly, thanks! i'll give this a shot in a bit, and post a follow up PR if it works out.

mjgiarlo: @jmartin-sul awesome, thx!

jmartin-sul: fixed in https://github.com/sul-dlss/happy-heron/pull/3604

PR #3613, opened 2024-09-12
File uploads should be disabled until version selection made

### LINKED ISSUE #3612: Modify versioning workflow to prevent issue where users think they've changed files but they haven't
Currently, if a user modifies the files on a deposit and after that selects "no" on the versioning question, they may not realize that their file changes are not being saved.

To prevent this, we have decided on the following modification to the versioning workflow:

1. The file upload/modification section of the form should be grayed out by default. 
2. If the user chooses the "yes" option for the versioning question, the file upload/modification section should be activated.

lwrubel:
# Why was this change made? 🤔
Resolves #3612 to keep files upload section disabled until user makes a version selection. 

Unit and PO. 

### DIFF
--- a/app/javascript/controllers/new_user_version_controller.js
+++ b/app/javascript/controllers/new_user_version_controller.js
@@ -6,6 +6,11 @@ export default class extends Controller {
   connect () {
     if (!this.hasUserVersionYesTarget || !this.hasUserVersionNoTarget) return
 
+    // when first loading page, neither version radio is selected
+    if (this.userVersionYesTarget.checked === false && this.userVersionNoTarget.checked === false) {
+      this.fileUploadsFieldsetTarget.disabled = true
+      this.fileSectionTarget.style.opacity = 0.5
+    }
     if (this.userVersionYesTarget.checked === true) {
       this.versionDescriptionNoTarget.disabled = true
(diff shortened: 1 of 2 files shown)


app/javascript/controllers/new_user_version_controller.js
```diff
@@ -6,6 +6,11 @@ export default class extends Controller {
   connect () {
     if (!this.hasUserVersionYesTarget || !this.hasUserVersionNoTarget) return
 
+    // when first loading page, neither version radio is selected
+    if (this.userVersionYesTarget.checked === false && this.userVersionNoTarget.checked === false) {
+      this.fileUploadsFieldsetTarget.disabled = true
+      this.fileSectionTarget.style.opacity = 0.5
```
justinlittman, line 12: You should just be able to add the `opacity-50` class. See https://getbootstrap.com/docs/5.3/utilities/opacity

justinlittman (approved):
Approve pending consideration of comments.

app/javascript/controllers/new_user_version_controller.js
lwrubel: Cool, thanks for pointing out that's available.

PR #3652, opened 2024-12-16
Enable new cops and a address a new cop complaint

aaron-collier:
# Why was this change made? 🤔

Another small bit of housekeeping.

### DIFF
--- a/app/services/cocina_generator/structural/generator.rb
+++ b/app/services/cocina_generator/structural/generator.rb
@@ -54,7 +54,8 @@ module CocinaGenerator
       # rubocop:disable Metrics/AbcSize
       def build_fileset(attached_file:, fileset: nil)
         # h2 only has one file per fileset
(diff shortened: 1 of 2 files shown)


app/services/cocina_generator/structural/generator.rb
```diff
@@ -54,7 +54,8 @@ def get_preserved_file(filename)
       # rubocop:disable Metrics/AbcSize
       def build_fileset(attached_file:, fileset: nil)
         # h2 only has one file per fileset
-        cocina_file = fileset&.structural&.contains&.first
+        structural = fileset&.structural
+        cocina_file = structural&.contains&.first
```
aaron-collier, line 58: This is to address the new `Style/SafeNavigationChainLength` cop without just doing a disable on the line. If we'd prefer the disable, I'm happy to take that approach as well.

justinlittman (approved):

PR #3674, opened 2025-02-10
Update dependencies

sul-devops-team:


.circleci/config.yml
```diff
@@ -1,37 +1,25 @@
 ---
 version: 2.1
 orbs:
-  ruby-rails: sul-dlss/ruby-rails@4.4.0
-  browser-tools: circleci/browser-tools@1.5.0
-  node: circleci/node@7.0.0
```
mjgiarlo, line 6: Now we use the cypress orb instead of the node orb to handle JS dependencies for the Cypress job.

.circleci/config.yml
```diff
@@ -1,37 +1,25 @@
...
-      - run:
-          name: Install javascript dependencies and build assets
-          command: bin/rake test:prepare
-      - run:
-          command: dockerize -wait tcp://localhost:5432 -timeout 1m
-          name: Wait for DB
-      - run:
-          command: bundle exec bin/cypress run
-          name: Test
+          package-manager: yarn
```
mjgiarlo, line 15: We were like 20 versions behind in Firefox. Bumping to the latest Firefox and Cypress versions led me down a bit of an orb rabbithole, but what we have now is terser, uses the orbs to a greater extent, and is hopefully more readable.

peetucket (approved):

PR #3690, opened 2025-03-20
add rake task to produce csv report on work types/subtypes

### LINKED ISSUE #3689: Report on work types and subtypes in H2
I'd like to look at which work subtypes are used most often with the text, data, image, sound, video, and software/code work types in H2. I'm not interested in the mixed materials, music, or other work types, but it's fine if they are included by default.

The goal being to see if we are presenting the most likely to be used ones in the short lists (a tester commented on not seeing something they expected to see in the short list). I can do some analysis on my end if I can get my hands on the data in some format. From H2 I can get this info from item reports, but have to run one report per collection and then compile all the data. This is not urgent.

peetucket:
# Why was this change made? 🤔

Fixes #3689 - create a csv file with work types/subtypes

Rails console on the server

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -97,7 +97,6 @@ GEM
     base64 (0.2.0)
     bcrypt (3.1.20)
     bcrypt_pbkdf (1.1.1)
-    bcrypt_pbkdf (1.1.1-arm64-darwin)
     benchmark (0.4.0)
     better_html (2.1.1)
       actionview (>= 6.0)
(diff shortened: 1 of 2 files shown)


mjgiarlo (approved):
LGTM, though we might not want the CSV committed in GitHub. Might it suffice to generate and share with the PO instead?

peetucket:
> LGTM, though we might not want the CSV committed in GitHub. Might it suffice to generate and share with the PO instead?

oops. yes will delete and export to tmp folder instead so it will be .gitignored

peetucket:
Will wait to merge after Amy confirms that has what she needs.

PR #3707, opened 2025-04-23
Improve caching of SUNet ID lookups

### LINKED ISSUE #3706: Collection managers should be able to add users via SUNet ID
SUNet IDs are not being found in all cases, even though we are fairly confidant these are valid SUNet IDs.

All of these are currently students enrolled at Stanford. For pbaker03, I've checked StanfordWho and this is listed as the SUNet ID there. This is where I would normally look up a SUNet -- this field should always be the real ID and not an alias.

Image

This is the list of IDs that the user said aren't working:
pbaker03
cbranche
dgiang25
tenzing
jaeahkim
oledbury
jamachin
zdo
maxcshen
lrn25
mxu37
sidraxu
lyilmaz
ariarias
marslani
vcarde
mayahg64
jlhamad
irehkim
tayne
dloth
audreynh
troper01
yjx717
ahzhang

This is the list of SUNet IDs the user was able to add:
apadam
mberlaga
kdong73
iainjung
leeallie
sshah21
etok
cwenardy
dbhatt1
bbroido
jastonm
soyeon25

mjgiarlo:
# Why was this change made?

Fixes #3706

This commit modifies how `AccountService` responses are cached:

1. When hitting the accountws API, retry a small number (5) of times in rapid succession to increase the chances a user has of getting a valid entry back quickly. Why? The accountws API is flaky, returning 500 error responses frequently.
2. When we get a 500 error currently, H2 is mistakenly caching the empty response, so legit users are showing up as missing when we consult the cache. With retries in place, now we writing nothing to the cache when none succeed.

# How was this change tested?

- [x] CI
- [x] qa/stage

### DIFF
--- a/app/services/account_service.rb
+++ b/app/services/account_service.rb
@@ -2,6 +2,8 @@
 
 # Retrieve names from the account API run by UIT
 class AccountService
+  class AccountServiceHiccough < StandardError; end
+
   def initialize
     pem_file = File.read(Settings.accountws.pem_file)
     @key = OpenSSL::PKey.read pem_file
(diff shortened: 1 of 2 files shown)


app/services/account_service.rb
```diff
@@ -2,18 +2,42 @@
...
+      begin
+        tries ||= 1
+        response = connection.get(url).body
+        # Raise and retry if the response is an HTTP 500.
+        #
+        # If, on the other hand, a bogus sunetid is provided, the `status` of the response will be 404, and then we:
+        #
+        # 1. Do *not* want to retry; but
+        # 2. *Do* want to cache the empty document
+        raise AccountServiceHiccough if response['status'] == 500
```
peetucket, line 29: small niggle about how `response` is the body of the response (maybe `response_body` is clearer?)  - this also count on the putting the status in the body of the response as opposed to checking the actual http response code itself (presuming they are doing the correct thing and actually returning a 500 code when there is an error)

mjgiarlo: @peetucket Good call. I renamed the variable to `response_body`. They do include the HTTP status in the body for both 404 and 500 errors, at least.

peetucket (approved):

PR #3743, opened 2025-05-23
Supports setting document object type.

### LINKED ISSUE #3737: Map to document object type
H3 supports mapping to the document object type when the work contains certain contents so that the document viewer is used on the PURL page. This is not currently supported in H2; for some works the object type has been changed in Argo.

In order to re-deposit works from H2 so that they can be imported into H3, H2 should also support mapping to the document object type.

See H3 for the rules under which a work should have the document object type.

This should allow removing the following normalization code:
```
def normalize_document
    # H2 doesn't set the object type to document for PDFs; H3 does.
    return unless document?

    cocina_attrs[:type] = Cocina::Models::ObjectType.document
    Array(cocina_attrs.dig(:structural, :contains)).each do |file_set_attrs|
      # file_set_attrs[:version] = cocina_object.version

      # Assuming one file per fileset.
      file_attrs = file_set_attrs.dig(:structural, :contains, 0)
      file_set_attrs[:type] = Cocina::Models::FileSetType.document if document_file?(file_attrs)
    end
  end

  def document_file?(file_attrs)
    file_attrs[:hasMimeType] == 'application/pdf' && file_attrs.dig(:administrative, :publish) == true
  end
```

justinlittman:
closes #3737

# Why was this change made? 🤔
To make H3 migration easier.


justinlittman:
Note that this is behind a feature flag.

mjgiarlo (approved):

